PR #2, opened 2020-08-11
Added Jenkinsfile

oshercc:


Jenkinsfile
```diff
@@ -0,0 +1,41 @@
...
+              when {
+                  branch 'master'
+              }
+
+              steps {
+                  withCredentials([usernamePassword(credentialsId: 'ocpmetal_cred', passwordVariable: 'PASS', usernameVariable: 'USER')]) {
+                      sh '''docker login quay.io -u $USER -p $PASS'''
+                  }
+
+                  sh '''docker tag  ${INSTALLER} ${INSTALLER}:latest'''
```
romfreiman, line 34: what is installer?

Jenkinsfile
```diff
@@ -0,0 +1,41 @@
+pipeline {
+
+  environment {
+        AGENT_IMAGE = 'quay.io/introspector-build'
```
romfreiman, line 4: how can it be that agent image is build?

Jenkinsfile
filanov: We need to make sure that we use the same image that we use for tests. so maybe you should create and env for it.
Take a look how it's done in assisted-service https://github.com/openshift/assisted-service/blob/master/Jenkinsfile#L4

Jenkinsfile
romfreiman: @oshercc ?

PR #5, opened 2020-08-11
MGMT-1695 Support HTTPS transport

empovit:
This PR adds control over the default HTTPS transport by adding `--insecure` option (skip certificate validation) and `--cacert` option (provide a custom root CA certificate as a PEM file).

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,9 @@ Currently, the functionality of **connectivity_check** is also bundled in the ag
 * *--text*: Dump the inventory as json to the standard output and exit.
 * *--connectivity*: Perform connectivity check and dump the results as json to the standard output.  The details of the nodes to check the connectivity with are provided as a parameter having json format.
 * *--with-text-logging*: Enable writing the agent logs to /var/log/agent.log. Default is true.
-* *--with-journal-logging*: Enable writing logs to systemd journal. Default is true.   
+* *--with-journal-logging*: Enable writing logs to systemd journal. Default is true.
+* *--insecure*: Skip certificate validation in case of HTTPS transport. Should be used only for testing. Default is false.
+* *--cacert*: Path to a custom CA certificate file in PEM format.
 * *--help*: Provide help message.
 
 ### Packaging
--- a/src/config/agent_config.go
+++ b/src/config/agent_config.go
@@ -22,6 +22,8 @@ var GlobalAgentConfig struct {
 	TextLogging        bool
 	AgentVersion       string
 	PullSecretToken    string
+	InsecureConnection bool
+	CACertificatePath  string
 }
 
 func printHelpAndExit() {
@@ -42,6 +44,8 @@ func ProcessArgs() {
 	flag.StringVar(&ret.InventoryImage, "inventory-image", "quay.io/ocpmetal/inventory:latest", "The image of inventory")
 	flag.BoolVar(&ret.JournalLogging, "with-journal-logging", true, "Use journal logging")
 	flag.BoolVar(&ret.TextLogging, "with-text-logging", false, "Output log to file")
+	flag.StringVar(&ret.CACertificatePath, "cacert", "", "Path to custom CA certificate in PEM format")
+	flag.BoolVar(&ret.InsecureConnection, "insecure", false, "Do not validate TLS certificate")
 	h := flag.Bool("help", false, "Help message")
 	flag.Parse()
 	if h != nil && *h {
--- a/src/session/inventory_session.go
+++ b/src/session/inventory_session.go
@@ -2,9 +2,14 @@ package session
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"fmt"
+	"io/ioutil"
+	"net"
 	"net/http"
 	"net/url"
+	"time"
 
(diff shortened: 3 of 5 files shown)


src/session/inventory_session.go
```diff
@@ -35,21 +40,70 @@ func (i *InventorySession) Client() *client.AssistedInstall {
...
+	certs, err := readCACertificate()
+	if err != nil {
+		return nil, err
+	}
+
+	clientConfig.Transport = requestid.Transport(&http.Transport{
+		Proxy: http.ProxyFromEnvironment,
+		DialContext: (&net.Dialer{
+			Timeout:   30 * time.Second,
+			KeepAlive: 30 * time.Second,
```
mhrivnak, line 60: Any particular reason to change from the default of 15s?

README.md
```diff
@@ -48,6 +50,14 @@ The subsystem tests uses docker-compose to run the agent and wiremock simulating
 
 To perform the subsystem test run: `skipper make subsystem`
 
+> **Note:** If many subsystem tests fail unexpectedly on Fedora/RHEL, it is likely a firewall blocks
```
ori-amizur, line 53: Might be irrelevant now after last changes https://github.com/openshift/assisted-installer-agent/pull/6

src/config/agent_config.go
```diff
@@ -42,6 +44,8 @@ func ProcessArgs() {
 	flag.StringVar(&ret.InventoryImage, "inventory-image", "quay.io/ocpmetal/inventory:latest", "The image of inventory")
 	flag.BoolVar(&ret.JournalLogging, "with-journal-logging", true, "Use journal logging")
 	flag.BoolVar(&ret.TextLogging, "with-text-logging", false, "Output log to file")
+	flag.StringVar(&ret.CACertificate, "cacert", "", "Path to custom CA certificate in PEM format")
+	flag.BoolVar(&ret.InsecureConnection, "insecure", false, "Do not validate TLS certificate")
```
mhrivnak, line 48: I think it's worth a log statement somewhere when this is set to `true`.

src/session/inventory_session.go
```diff
@@ -35,21 +40,70 @@ func (i *InventorySession) Client() *client.AssistedInstall {
...
 	clientConfig := client.Config{}
-	clientConfig.URL, _ = url.Parse(createUrl())
-	clientConfig.Transport = requestid.Transport(http.DefaultTransport)
+	var err error
+	clientConfig.URL, err = url.ParseRequestURI(createUrl())
+	if err != nil {
+		return nil, err
+	}
+
+	certs, err := readCACertificate()
```
ori-amizur, line 51: Why are attempting to read ca certificate if we work in insecure mode?

src/session/inventory_session.go
```diff
@@ -35,21 +40,70 @@ func (i *InventorySession) Client() *client.AssistedInstall {
...
+	})
+
 	clientConfig.AuthInfo = auth.AgentAuthHeaderWriter(config.GlobalAgentConfig.PullSecretToken)
 	bmInventory := client.New(clientConfig)
-	return bmInventory
+	return bmInventory, nil
 }
 
-func New() *InventorySession {
+func readCACertificate() (*x509.CertPool, error) {
```
ori-amizur, line 78: Do we assume that the CACertificate is root CA?

README.md
empovit: Removing, thanks!

src/config/agent_config.go
```diff
@@ -22,6 +22,8 @@ var GlobalAgentConfig struct {
 	TextLogging        bool
 	AgentVersion       string
 	PullSecretToken    string
+	InsecureConnection bool
+	CACertificate      string
```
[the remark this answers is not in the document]
empovit: Good point, will change

src/session/inventory_session.go
empovit: Hmmm, the default is 30 https://github.com/golang/go/blob/release-branch.go1.13/src/net/http/transport.go#L45. I copied DefaultTransport as is, just removed DualStack, which is deprecated.

src/config/agent_config.go
empovit: Will do

src/session/inventory_session.go
empovit: Right, will be fixed.

src/session/inventory_session.go
empovit: Theoretically any server cert can be passed, but it CA makes more sense.

src/session/inventory_session.go
mhrivnak: The [docs here](https://golang.org/pkg/net/#Dialer) say 15s is the default for the `Dialer`. It's interesting that `DefaultTransport` changes that default. Not a big deal, and I'm sure either is fine.

src/session/inventory_session.go
empovit: The corresponding field of `TLSClientConfig` is called `RootCAs`.

src/session/inventory_session.go
empovit: Interesting. In this particular case I will probably keep the behavior that was already tested (i.e. http.DefaultTransport).

PR #16, opened 2020-08-20
Avoid subsystem failure when running on system that only virtual nics have IP addresses

ori-amizur:
/cc @tsorya 
/assign @tsorya


tsorya (commented):
/lgtm

PR #20, opened 2020-08-24
MGMT-1719 create dhcp lease allocator

YuviGold:


src/dhcp_lease_allocator/lease.go
```diff
@@ -0,0 +1,76 @@
+package dhcp_lease_allocator
```
ori-amizur, line 1: The dhclient should be used in synchronous way with "-1" flag.  It will make code simpler and easier to understand, and less error prone.

subsystem/lease_test.go
```diff
@@ -0,0 +1,170 @@
+package subsystem
```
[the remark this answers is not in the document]
YuviGold: I am used to make everything in a container. Since it is already wrapper inside skipper you are right it is unnecessary to wrap it twice.

subsystem/lease_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	BeforeEach(func() {
+		resetAll()
+		hostID = nextHostID()
+
+		apiMac = generateMacString()
+		ingressMac = generateMacString()
+		ifaceName = "eth0"
+	})
+
```
eranco74, line 33: Please add bad flow, should validate exit code and error message

src/dhcp_lease_allocate/dhcp_lease_allocate.go
```diff
@@ -0,0 +1,103 @@
...
+
+	"github.com/go-openapi/strfmt"
+	"github.com/openshift/assisted-installer-agent/src/util"
+	"github.com/openshift/assisted-service/models"
+	"github.com/openshift/baremetal-runtimecfg/pkg/monitor"
+
+	"github.com/sirupsen/logrus"
+)
+
+const configPath string = "/tmp/config.path"
```
ori-amizur, line 15: What id the purpose of configPath?

src/dhcp_lease_allocate/dhcp_lease_allocate.go
```diff
@@ -0,0 +1,103 @@
...
+
+//go:generate mockery -name Executer -inpkg
```
ori-amizur, line 17: Suggest to remove if not used

PR #22, opened 2020-08-25
MGMT-1912 Remove docker install from Dockerfile.assisted_installer_agent

ori-amizur:
… and add Dockerfile for subsystem testing that adds docker installation

   This will cause that docker will be installed in a container only for subsystem testing.

/cc @romfreiman 
/cc @ronniel1 
/cc @eranco74


YuviGold:
Can't we move the subsystem tests to use podman as well as the runtime container?

YuviGold (approved):
LGTM

PR #38, opened 2020-09-22
MGMT-2075 API VIP connectivity check command

danielerez:
* Added apivip_check command:
- For ensuring connectivity between the node to and the API.
- Done by invoking a GET request to <api>/config/worker (to download the worker.ign file).

* Added relevant tests with wiremock stubs in subsystem/apivip_check_test.

### DIFF
--- a/src/apivip_check/apivip_check.go
+++ b/src/apivip_check/apivip_check.go
@@ -1,0 +1,74 @@
+package apivip_check
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/openshift/assisted-service/models"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	WorkerIgnitionPath = "/config/worker"
+)
+
+func CheckAPIConnectivity(checkAPIRequestStr string, log logrus.FieldLogger) (stdout string, stderr string, exitCode int) {
+	var checkAPIRequest models.APIVipConnectivityRequest
+
+	if err := json.Unmarshal([]byte(checkAPIRequestStr), &checkAPIRequest); err != nil {
+		wrapped := errors.Wrap(err, "Error unmarshaling APIVipConnectivityRequest")
+		log.WithError(err).Error(wrapped.Error())
+		return createRepsonse(false), wrapped.Error(), -1
+	}
+
+	if checkAPIRequest.URL == nil {
+		err := errors.New("Missing URL in checkAPIRequest")
+		log.WithError(err).Error(err.Error())
+		return createRepsonse(false), err.Error(), -1
+	}
+	
+	if err := httpDownload(*checkAPIRequest.URL + WorkerIgnitionPath, log); err != nil {
+		wrapped := errors.Wrap(err, "Failed to download worker.ign file")
+		log.WithError(err).Error(wrapped.Error())
+		return createRepsonse(false), wrapped.Error(), -1
+	}
+
+	return createRepsonse(true), "", 0
+}
+
+func createRepsonse(success bool) string {
+	checkAPIResponse := models.APIVipConnectivityResponse{
+		IsSuccess: success,
+	}
+	bytes, err := json.Marshal(checkAPIResponse)
+	if err != nil {
+		return ""
+	}
+	return string(bytes)
+}
+
+func httpDownload(uri string, log logrus.FieldLogger) error {
+    res, err := http.Get(uri)
+    if err != nil {
+		return errors.Wrap(err, "HTTP download failure")
+	}
+	
+    defer res.Body.Close()
+    bytes, err := ioutil.ReadAll(res.Body)
+    if err != nil {
+		return errors.Wrap(err, "File read failure")
+	}
+
+	if len(bytes) == 0 {
+		return errors.New("Empty Ignition file")
+	}
+
+	var js json.RawMessage
+    if err = json.Unmarshal(bytes, &js); err != nil {
+		return errors.Wrap(err, "Error unmarshaling Ignition string")
+	}
+
+    return err
+}
--- a/src/apivip_check/apivip_check_test.go
+++ b/src/apivip_check/apivip_check_test.go
@@ -1,0 +1,96 @@
+package apivip_check
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/openshift/assisted-service/models"
+	"github.com/sirupsen/logrus"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("API connectivity check test", func() {
+	var log logrus.FieldLogger
+	var srv *httptest.Server
+
+	BeforeEach(func() {
+		log = logrus.New()
+	})
+
+	AfterEach(func() {
+		srv.Close()
+	})
+
+	It("HTTP download ignition file", func() {
+		srv = serverMock(ignitionMock)
+		_, _, exitCode := CheckAPIConnectivity(getRequestStr(&srv.URL), log)
+		Expect(exitCode).Should(Equal(0))
+	})
+
+	It("Invalid ignition file format", func() {
+		srv = serverMock(ignitionMockInvalid)
+		_, _, exitCode := CheckAPIConnectivity(getRequestStr(&srv.URL), log)
+		Expect(exitCode).Should(Equal(-1))
+	})
+
+	It("Empty ignition", func() {
+		srv = serverMock(ignitionMockEmpty)
+		_, _, exitCode := CheckAPIConnectivity(getRequestStr(&srv.URL), log)
+		Expect(exitCode).Should(Equal(-1))
+	})
+
+	It("Invalid API URL", func() {
+		url := "http://127.0.0.1:2345"
+		_, _, exitCode := CheckAPIConnectivity(getRequestStr(&url), log)
+		Expect(exitCode).Should(Equal(-1))
+	})
+
+	It("Missing API URL", func() {
+		_, _, exitCode := CheckAPIConnectivity(getRequestStr(nil), log)
+		Expect(exitCode).Should(Equal(-1))
+	})
+})
+
+func getRequestStr(url *string) string {
+	request := models.APIVipConnectivityRequest{
+		URL: url,
+	}
+
+	requestBytes, err := json.Marshal(request)
+	if err != nil {
+		return ""
+	}
+	return string(requestBytes)
+}
+
+func serverMock(mock func(w http.ResponseWriter, r *http.Request)) *httptest.Server {
+	handler := http.NewServeMux()
+	handler.HandleFunc(WorkerIgnitionPath, mock)
+	srv := httptest.NewServer(handler)
+	return srv
+}
+
+func ignitionMock(w http.ResponseWriter, r *http.Request) {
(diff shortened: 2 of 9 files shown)


danielerez:
@ori-amizur @yevgeny-shnaidman

src/apivip_check/apivip_check.go
```diff
@@ -0,0 +1,48 @@
...
+	"io/ioutil"
+	"net/http"
+
+	"github.com/openshift/assisted-installer-agent/src/util"
+
+	"github.com/sirupsen/logrus"
+)
+
+//go:generate mockery -name Executer -inpkg
+type Executer interface {
```
ori-amizur, line 14: This is used to enable both unit tests and actual flow, but there are no unit tests here.  It is also for running sub-process which is not done here.
Mocking should be done on http.Get function in the same way that on other place we did for Execute

src/apivip_check/apivip_check.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+type ProcessExecuter struct{}
+
+func (e *ProcessExecuter) Execute(command string, args ...string) (stdout string, stderr string, exitCode int) {
+	return util.Execute(command, args...)
+}
+
+func CheckAPIConnectivity(apiVip string, e Executer, log logrus.FieldLogger) (stdout string, stderr string, exitCode int) {
+	workerURL := fmt.Sprintf("%s/config/worker", apiVip)
```
ori-amizur, line 25: apiVip is usually an IP address.  Probably a URL is provided here since it should contain also a scheme (http/https) and maybe a port number

subsystem/apivip_check_test.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+type APIConnectivityCheckVerifier struct{}
+
+func (i *APIConnectivityCheckVerifier) verify(actualReply *models.StepReply) bool {
+	if actualReply.ExitCode != 0 {
+		log.Errorf("APIConnectivityCheckVerifier returned with exit code %d. error: %s", actualReply.ExitCode, actualReply.Error)
+		return false
+	}
+	if actualReply.StepType != models.StepTypeConnectivityCheck {
```
ori-amizur, line 69: Are we using the step-type StepTypeConnectivityCheck? It is used for something else

subsystem/utils.go
```diff
@@ -22,6 +23,21 @@ var (
 	MappingsURL   = fmt.Sprintf("%s/__admin/mappings", WireMockURL)
 )
 
+const (
+	workerIgnitionURL = "/config/worker"
+
+	nodeIgnitionFormat = `{
```
ori-amizur, line 29: Are we checking ignition format?  The code does not verify it

ori-amizur (commented):
Please add unit tests

eranco74:
When is apivip_check is triggered?

src/apivip_check/apivip_check.go
yevgeny-shnaidman: @ori-amizur in our case api vip is a hostname, not an IP

yevgeny-shnaidman: the complete URL should be:  http://hostname:22624/config/worker

src/apivip_check/apivip_check.go
```diff
@@ -0,0 +1,48 @@
...
+
+func httpDownload(uri string, log logrus.FieldLogger) ([]byte, error) {
+	fmt.Printf("CheckAPIConnectivity From: %s\n", uri)
+    res, err := http.Get(uri)
+    if err != nil {
+		log.WithError(err).Errorf("CheckAPIConnectivity: HTTP download failure")
+		return nil, err
+    }
+    defer res.Body.Close()
+    d, err := ioutil.ReadAll(res.Body)
```
yevgeny-shnaidman, line 42: maybe we can add here a check that the Body we received is of json format, just to make sure

subsystem/utils.go
yevgeny-shnaidman: @ori-amizur i think that we just need to check that what we received is in json format. No need to check the correctness of the dta itself

yevgeny-shnaidman:
> When is apivip_check is triggered?

it will be triggered along with connectivity check, in insufficient and pending for input states

src/apivip_check/apivip_check.go
ori-amizur: > @ori-amizur in our case api vip is a hostname, not an IP

Probably dns name

subsystem/utils.go
ori-amizur: @yevgeny-shnaidman In the code itself we do not verify it in any way.  So why in the tests?  If it needed, please verify it first in the code.

src/apivip_check/apivip_check.go
danielerez: ok, added tests in apivip_check_test.go

src/apivip_check/apivip_check.go
danielerez: right, added ConnectivityCheckAPIRequest with scheme/host/port instead of  apiVip string

subsystem/apivip_check_test.go
danielerez: Added a new step - StepTypeConnectivityCheckAPI

subsystem/utils.go
danielerez: Added verifications by json marshalling against NodeIgnition struct.

src/apivip_check/apivip_check.go
danielerez: sure, checked by json marshalling on NodeIgnition struct

src/apivip_check/apivip_check.go
```diff
@@ -0,0 +1,103 @@
...
+
+    return b, err
+}
+
+
+func GetWorkerIgnitionURL(scheme, host, port string) string {
+	return fmt.Sprintf("%s://%s:%s" + WorkerIgnitionPath, scheme, host, port)
+}
+
+func FormatNodeIgnitionFile(source string) ([]byte, error) {
```
ori-amizur, line 89: Please don't put functions or any code that are used only for testing in the main code.

ori-amizur (commented):
Please add unit-tests and test also various failure scenarios

src/apivip_check/apivip_check.go
ori-amizur: Still how do we verify that the dns name is resolvable?

src/apivip_check/apivip_check_test.go
```diff
@@ -0,0 +1,62 @@
...
+	"net/http/httptest"
+	"testing"
+
+	"github.com/sirupsen/logrus"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("API connectivity check test", func() {
```
ori-amizur, line 14: Please test bad scenarios

src/apivip_check/apivip_check.go
danielerez: We are verifying connectivity by invoking a request for downloading the work file, so if the DNS is not resolvable we should fail on the http GET request. I'm not sure we can ensure whether the specified API VIP is resolvable through a public DNS or internally via DHCP options.

@yevgeny-shnaidman @ronniel1 @avishayt What do you think? Is it sufficient for checking connectivity?

src/apivip_check/apivip_check.go
danielerez: ok, moved to utils.

src/apivip_check/apivip_check_test.go
danielerez: added

src/apivip_check/apivip_check.go
yevgeny-shnaidman: it is users responsibility that nodes will be able to resolve dns names. otherwise even kubelet won't be able to connect since kubelet config include dns name and not ip. In case correct dns respolution is not available, then VIP connnectivity will fails, which is ok

ori-amizur: @yevgeny-shnaidman Is it something that was a requirement for day1, or is it new requirement for day2?

src/apivip_check/apivip_check.go
```diff
@@ -0,0 +1,78 @@
...
+func CheckAPIConnectivity(checkAPIRequestStr string, log logrus.FieldLogger) (stdout string, stderr string, exitCode int) {
+	var checkAPIRequest models.ConnectivityCheckAPIRequest
+
+	err := json.Unmarshal([]byte(checkAPIRequestStr), &checkAPIRequest)
+	if err != nil {
+		log.WithError(err).Errorf("Error unmarshaling ConnectivityCheckAPIRequest")
+		return "", err.Error(), -1
+	}
+	
+	_, err = httpDownload(*checkAPIRequest.URL + WorkerIgnitionPath, log)
```
ori-amizur, line 48: Maybe check that URL is not nil ?

src/apivip_check/apivip_check.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	
+    defer res.Body.Close()
+    b, err := ioutil.ReadAll(res.Body)
+    if err != nil {
+		log.WithError(err).Errorf("File read failure")
+		return nil, err
+	}
+
+	var nodeIgnition NodeIgnition
```
ori-amizur, line 70: If you really want to verify that you got the correct data, maybe check it is not empty

src/apivip_check/apivip_check.go
danielerez: sure, added

src/apivip_check/apivip_check.go
danielerez: good idea, added.

src/apivip_check/apivip_check.go
```diff
@@ -0,0 +1,77 @@
...
+	var checkAPIRequest models.APIVipConnectivityRequest
+
+	if err := json.Unmarshal([]byte(checkAPIRequestStr), &checkAPIRequest); err != nil {
+		log.WithError(err).Errorf("Error unmarshaling APIVipConnectivityRequest")
+		return createRepsonse(false), err.Error(), -1
+	}
+
+	if checkAPIRequest.URL == nil {
+		err := errors.New("Missing URL in checkAPIRequest")
+		log.WithError(err)
```
ori-amizur, line 27: This doesn't do anything

src/apivip_check/apivip_check.go
```diff
@@ -0,0 +1,77 @@
...
+
+const (
+	WorkerIgnitionPath = "/config/worker"
+)
+
+func CheckAPIConnectivity(checkAPIRequestStr string, log logrus.FieldLogger) (stdout string, stderr string, exitCode int) {
+	var checkAPIRequest models.APIVipConnectivityRequest
+
+	if err := json.Unmarshal([]byte(checkAPIRequestStr), &checkAPIRequest); err != nil {
+		log.WithError(err).Errorf("Error unmarshaling APIVipConnectivityRequest")
```
ori-amizur, line 21: In general, all returned errors in stderr should contain full information about what was the error.  If assisted-service receives the error text the json produces without context, it will be meaningless

src/apivip_check/apivip_check.go
danielerez: right, added actual log.

src/apivip_check/apivip_check.go
danielerez: wrapped errors with the additional info.

ori-amizur (approved):
/approve
/lgtm

PR #40, opened 2020-10-01
MGMT-2216 Store locally obtained host ID instead response from server

empovit:
It's supposed to simplify the flow, reduce the amount of redundant info we store, and allow more flexibility in the API because we don't use a part of the response we don't need.


empovit:
Local subsystem tests didn't pick up the right agent image. Retested and found the cause of failures. Closing until I can figure out how to mock a host ID in subsystem tests.

PR #41, opened 2020-10-04
MGMT-2216 Store locally obtained host ID instead response from server

empovit:


empovit:
Closing until I can figure out how to mock host ID in subsystem tests.

PR #46, opened 2020-10-07
MGMT-2318 added network CIDR verification to API VIP check

danielerez:
Added network CIDR verification to API VIP connectivity check:
- Parse IP from API VIP.
- Get interfaces of the node.
- Calculate CIDRs by interfaces.
- API is verified if contained in one of the CIDRs.


danielerez:
@ori-amizur @yevgeny-shnaidman

ori-amizur (approved):
/lgtm

PR #47, opened 2020-10-07
MGMT-2213 Split agent into register and next step

empovit:


src/commands/register_node.go
```diff
@@ -17,7 +17,7 @@ import (
 	"github.com/openshift/assisted-service/models"
 )
 
-var CurrentHost *models.Host
+var CurrentHostID *strfmt.UUID
```
ori-amizur, line 20: Where is this used?

src/next_step_runner/main/main.go
```diff
@@ -0,0 +1,27 @@
...
+	"fmt"
+
+	"github.com/openshift/assisted-installer-agent/src/commands"
+	"github.com/openshift/assisted-installer-agent/src/config"
+	"github.com/openshift/assisted-installer-agent/src/util"
+)
+
+func main() {
+	config.ProcessArgs()
+	util.SetLogging("agent_next_step_runner", config.GlobalAgentConfig.TextLogging, config.GlobalAgentConfig.JournalLogging)
```
ori-amizur, line 13: Since it is not already part of the agent, the invocation of inventory and connectivity can be removed

src/commands/register_node.go
empovit: It's not in use anymore. Nice catch!

src/next_step_runner/main/main.go
empovit: Will do, thanks for confirming that.

src/agent/main/main.go
```diff
@@ -1,28 +1,42 @@
...
+				reRegistrerDelay = time.Duration(agentCommand.RetrySeconds) * time.Second
+			} else {
+				reRegistrerDelay = defaultRetryDelay
+			}
+
+			log.WithError(err).Errorf("Failed to start next step runner, going to retry in %s", reRegistrerDelay)
+			time.Sleep(reRegistrerDelay)
+			continue
+		}
+
```
eranco74, line 39: Log the error

src/commands/register_node.go
```diff
@@ -31,17 +29,18 @@ func createRegisterParams() *installer.RegisterHostParams {
...
+func RegisterHostWithRetry() *models.HostRegistrationResponseAO1NextStepRunnerCommand {
+
 	for {
 		s, err := session.New(config.GlobalAgentConfig.TargetURL, config.GlobalAgentConfig.PullSecretToken)
 		if err != nil {
 			logrus.Fatalf("Failed to initialize connection: %e", err)
 		}
 		registerResult, err := s.Client().Installer.RegisterHost(s.Context(), createRegisterParams())
 		if err == nil {
-			CurrentHost = registerResult.Payload
```
eranco74, line 42: Since it's all wrapped in retries consider logging the error and returning nil instead of logging.Fatal (line 37)

src/util/execute.go
```diff
@@ -58,3 +58,10 @@ func ExecuteOutputToFile(outputFilePath string, command string, args ...string)
 func ExecuteShell(command string) (stdout string, stderr string, exitCode int) {
 	return Execute("bash", "-c", command)
 }
+
+func ExecutePrivileged(command string, args ...string) (stdout string, stderr string, exitCode int) {
+	commandBase := "nsenter"
```
eranco74, line 63: Just to make sure, the agent is running in a container?

src/commands/start_agent.go
```diff
@@ -0,0 +1,14 @@
+package commands
+
+import (
+	"github.com/openshift/assisted-installer-agent/src/util"
+	"github.com/pkg/errors"
+)
+
+func StartAgent(command string, args ...string) error {
+	_, stderr, exitCode := util.Execute(command, args...)
```
eranco74, line 9: Please add log of the command and args

src/agent/main/main.go
```diff
@@ -1,28 +1,42 @@
...
+		if agentCommand == nil {
+			log.Errorf("Incompatible server version, going to retry in %s", defaultRetryDelay)
+			time.Sleep(defaultRetryDelay)
+			continue
 		}
-	} else {
-		commands.RegisterHostWithRetry()
-		commands.ProcessSteps()
+
+		if err := commands.StartAgent(agentCommand.Command, agentCommand.Args...); err != nil {
```
eranco74, line 26: So start agent run a single (generic) command?
Will this command hang as long the nextStepRunner is running?

empovit: Yes, it will run in the foreground per design.

src/agent/main/main.go
empovit: I think I'm logging it in line 35 https://github.com/openshift/assisted-installer-agent/pull/47/files/994e877c9a1a9c04037d6968c7e4c78156213893#diff-3fc728280f29db82b2a2bd7a811c209dR35

src/commands/register_node.go
empovit: I also thought so, but this error is not recoverable unlike others that can be fixed by retrying. E.g. it will be caused by a malformed URL.

src/commands/start_agent.go
empovit: Will do.

src/util/execute.go
empovit: The "next step runner" part is running as a container. Since it may be required to run other containers (the tools), it will use `nsenter` to do so.

src/agent/main/main.go
empovit: It sort of replaces the agent - runs continuously performing get-next-step and post-reply. With the difference that it's executed by what remains of the agent (host registration), while the agent is executed by systemd.

empovit:
@mhrivnak I'd appreciate it if you could review this PR for any impact on on-prem.

mhrivnak:
> @mhrivnak I'd appreciate it if you could review this PR for any impact on on-prem.

I'm happy to, but can you first describe the motivation for this change? What problem is it solving?

empovit:
> > @mhrivnak I'd appreciate it if you could review this PR for any impact on on-prem.
> 
> I'm happy to, but can you first describe the motivation for this change? What problem is it solving?

The idea is to have a lean and dumb agent that registers with the assisted service and in return receives a command for running the main part (the next step + post reply loop). This is supposed to address situations when a host is booted from an ISO generated long ago, so that it doesn't contain recent changes and may even become incompatible with current server version.

mhrivnak:
On-prem use will often also be disconnected. When the live ISO boots, at that moment it will need access to a container image registry that has the images it is hard-coded to run. In the scenario you're working on, those would be image references dating back to when the live ISO was created.

Based on this change, those old images AND any newer version will all need to be mirrored into the user's local private registry. I don't think anything in this change makes that more difficult, but it is a constraint worth keeping in mind.

src/util/execute.go
eranco74: Seems that the nest step runner can die and no one will know.
So while we think the agent is running technically nothing in the host will query for next steps

empovit: It's executed by the host registration part in the foreground. So if it exits for whatever reason (non-zero exit code will be reported to the log), the registration part will resume and retry after a while in this loop https://github.com/openshift/assisted-installer-agent/pull/47/files#diff-80523fa9fafe15707751db5e800200a7b4a51fc32755bbc0b760b1c6490c9da0R18. After re-registering, this part again will receive a new command from the server and run it. There is another PR coming that will allow us to initiate normal termination of the next step runner and force re-register. This can be useful if we want to support initiated agent updates after registration (but before installation), between two "next step" cycles.

empovit: Clarification: The next-step/post-reply loop logic remains unchanged. It just runs as a separate executable now (the next step runner). It handles command (next step) failures just as before.

ori-amizur (commented):
/lgtm

PR #48, opened 2020-10-09
[WIP] Use centos image rather than fedora

carbonin:
This should fix the compatibility problems with systemd libs

Previously we got a new version of the fedora image and it was failing
to collect logs with the following error:

`Output Failed to run send logs journalctl: error while loading shared libraries: libsystemd-shared-245.so: cannot open shared object file: No such file or directory`

This seems to be because the version of systemd is much newer on the
fedora container than it is on the RHCOS host

### DIFF
--- a/Dockerfile.assisted_installer_agent
+++ b/Dockerfile.assisted_installer_agent
@@ -1,5 +1,5 @@
-FROM registry.fedoraproject.org/fedora-minimal:31
-RUN microdnf install \
+FROM quay.io/app-sre/centos:8
+RUN dnf -y install \
     kmod findutils iputils \
     # inventory
     dmidecode ipmitool lshw biosdevname \
@@ -8,6 +8,6 @@ RUN microdnf install \
     # dhcp_lease_allocate
     dhclient \
     # logs_sender
-    tar && microdnf update systemd \
-    && microdnf clean all
+    tar && dnf -y update systemd \
+    && dnf -y clean all
 ADD build/agent build/connectivity_check build/free_addresses build/inventory build/logs_sender build/dhcp_lease_allocate build/apivip_check /usr/bin/
--- a/subsystem/Dockerfile.agent_test
+++ b/subsystem/Dockerfile.agent_test
@@ -1,2 +1,3 @@
 FROM quay.io/ocpmetal/assisted-installer-agent:latest
-RUN microdnf install docker && microdnf clean all
+RUN curl --output /etc/yum.repos.d/docker-ce.repo https://download.docker.com/linux/centos/docker-ce.repo
+RUN dnf install -y docker-ce docker-ce-cli containerd.io && dnf -y clean all


carbonin:
This also moves us away from the minimal image, but it actually only increases the final image size by ~70MB
```
$ docker images
REPOSITORY                                     TAG           IMAGE ID         CREATED             SIZE
quay.io/carbonin/assisted-installer-agent      latest        16c014a1980e     21 minutes ago      481MB
quay.io/ocpmetal/assisted-installer-agent      latest        1abe3ceea216     17 hours ago        414MB
```

romfreiman:
The best is to align to what we had in prod. I think there was a discussion about it a while ago
@gamli75  @djzager

mhrivnak:
This results in an image where some of the packages are from RHEL (UBI) and some from centos. AIUI once you have to install non-UBI packages, it is better to not mix like this and to go 100% centos upstream and 100% RHEL downstream. That said, it's not easy to find formal guidance on that.

Is there an advantage to using UBI+centos rather than just centos?

carbonin:
> Is there an advantage to using UBI+centos rather than just centos?

I didn't think so, but while trying to check the image size I realized that the only place I seem to have permissions to pull a even semi-official centos image from is docker.io. With the new rate limits I figured we shouldn't be doing that, right?

I found https://quay.io/repository/app-sre/centos, but when changing this image to use that as the base I'm getting permissions errors. I doubt that will work for us in CI.

mhrivnak:
> > Is there an advantage to using UBI+centos rather than just centos?
> 
> I didn't think so, but while trying to check the image size I realized that the only place I seem to have permissions to pull an even semi-official centos image from is docker.io. With the new rate limits I figured we shouldn't be doing that, right?

That's a good question to ask the centos project. I'm sure they want to have their official images hosted where people can use them, so it's possible they could be planning a move or already mirroring them elsewhere.

carbonin:
> That's a good question to ask the centos project

Great suggestion. I found this https://github.com/CentOS/sig-cloud-instance-images/issues/160 so I'll probably do the same. (use `registry.centos.org/centos/centos:8`)

carbonin:
The branch name is no longer accurate, but I'd rather keep this PR because of the discussion.

mhrivnak:
> > That's a good question to ask the centos project
> 
> Great suggestion. I found this [CentOS/sig-cloud-instance-images#160](https://github.com/CentOS/sig-cloud-instance-images/issues/160) so I'll probably do the same. (use `registry.centos.org/centos/centos:8`)

I asked on freenode in #centos-devel and got this: "<Arrfab> mhrivnak: registry.centos.org isn't probably maintained so not up2date"

Not sure how authoritative that assessment is, but it's worth investigating.

carbonin:
Yikes. Did they offer an alternative?

mhrivnak:
> Yikes. Did they offer an alternative?

Nope. :(

carbonin:
Guess this was done in https://github.com/openshift/assisted-installer-agent/pull/63.
Thanks @osherdp :tada:

PR #52, opened 2020-10-14
MGMT-2069: handle getting 409 from server when host tries to register…

osherdp:
… to a cluster

/hold
(until PR on assisted-service is merged, then go.mod should be updated)

### DIFF
--- a/src/commands/register_node.go
+++ b/src/commands/register_node.go
@@ -45,6 +45,10 @@ func RegisterHostWithRetry() {
 		// stop register in case of forbidden reply.
 		switch errValue := err.(type) {
 		case *installer.RegisterHostForbidden:
+			s.Logger().Warn("Host will stop trying to register; host is not allowed to perform the requested operation")
+			// wait forever
+			select {}
+		case *installer.RegisterHostConflict:
 			s.Logger().Warn("Host will stop trying to register; cluster cannot accept new hosts in its current state")
 			// wait forever
 			select {}
--- a/subsystem/agent_test.go
+++ b/subsystem/agent_test.go
@@ -50,31 +50,43 @@ var _ = Describe("Agent tests", func() {
 		Expect(deleteStub(nextStepsStubID)).NotTo(HaveOccurred())
 	})
 
-	It("register forbidden", func() {
-		hostID := nextHostID()
-		registerStubID, err := addRegisterStub(hostID, http.StatusForbidden)
-		Expect(err).NotTo(HaveOccurred())
+	Context("register action makes the agent to wait forever", func() {
+		var status int
 
-		nextStepsStubID, err := addNextStepStub(hostID, defaultnextInstructionSeconds)
-		Expect(err).NotTo(HaveOccurred())
+		It("register conflicted - cluster does not accept hosts", func() {
+			status = http.StatusConflict
+		})
 
-		Expect(startAgent()).NotTo(HaveOccurred())
-		time.Sleep(10 * time.Second)
+		It("register forbidden", func() {
+			status = http.StatusForbidden
+		})
 
-		// validate only register request was called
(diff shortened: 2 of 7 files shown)


osherdp:
depends on openshift/assisted-service#534

src/commands/register_node.go
```diff
@@ -44,6 +44,10 @@ func RegisterHostWithRetry() *models.HostRegistrationResponseAO1NextStepRunnerCo
 		// stop register in case of forbidden reply.
 		switch errValue := err.(type) {
 		case *installer.RegisterHostForbidden:
+			s.Logger().Warn("Host will stop trying to register; host is forbidden performing the asked operation")
```
ori-amizur, line 47: I think "requested" is better than "asked"

subsystem/agent_test.go
```diff
@@ -129,7 +129,34 @@ var _ = Describe("Agent tests", func() {
...
+		for _, r := range requests.Requests {
+			req = append(req, r)
+		}
+		Expect(len(req)).Should(Equal(1))
+		Expect(deleteStub(registerStubID)).NotTo(HaveOccurred())
+		Expect(deleteStub(nextStepsStubID)).NotTo(HaveOccurred())
+	})
+
+	It("register conflicted - cluster does not accept hosts", func() {
+		hostID := nextHostID()
```
ori-amizur, line 142: Maybe have the same test run with just a different param which is the status code? It can be in a loop

subsystem/utils.go
```diff
@@ -260,6 +260,15 @@ func addRegisterStub(hostID string, reply int, clusterID string) (string, error)
 		if err != nil {
 			return "", err
 		}
+	case http.StatusConflict:
+		errorReply := &models.InfraError{
```
ori-amizur, line 264: Are you sure it is InfraError ?

osherdp: yes
https://github.com/openshift/assisted-service/blob/master/swagger.yaml#L1271

subsystem/agent_test.go
osherdp: done

subsystem/utils.go
ori-amizur: Maybe this is wrong too. Infra error should be generated by infrastructure.  This is generated in the API call. I think the swagger has to change.

ori-amizur: @avishayt ^ ?

osherdp: oh I get what you're saying
does it have any functional difference? or mainly conventional?

avishayt: Yes, swagger should be error, not infra_error.  infra_error is only 401/403 and is returned by assisted-service code that we don't fully control, hence it returns something a bit different.  error should be used for everything other than 401/403.

osherdp: Alright, made a fix here:
https://github.com/openshift/assisted-service/pull/566

osherdp: Fixed here too

src/commands/register_node.go
```diff
@@ -45,6 +45,10 @@ func RegisterHostWithRetry() {
 		// stop register in case of forbidden reply.
 		switch errValue := err.(type) {
 		case *installer.RegisterHostForbidden:
+			s.Logger().Warn("Host will stop trying to register; host is forbidden performing the requested operation")
```
ronniel1, line 48: Sentence does not seem right.
"...; host is not allowed to perform the requested operation"

ronniel1 (approved):
/lgtm

src/commands/register_node.go
osherdp: done

ronniel1 (approved):

PR #53, opened 2020-10-14
MGMT-2347: trim the free-addresses message

osherdp:
```
> exit-code <255>Sending step <free-addresses> reply output <[{"free_addresses...> error <> exit-code <0>Sending step <free-addresses> reply output <> error <nmap: Target.cc:503: void Target::stopTimeOutClock(const timeval*): Assertion `htn.toclock_running == true' failed.
```

### DIFF
--- a/src/commands/step_processor.go
+++ b/src/commands/step_processor.go
@@ -38,7 +38,14 @@ func (s *stepSession) sendStepReply(stepType models.StepType, stepID, output, er
 	if exitCode != 0 {
 		logFunc = s.Logger().Warnf
 	}
-	logFunc("Sending step <%s> reply output <%s> error <%s> exit-code <%d>", stepID, output, errStr, exitCode)
+
+	if stepType == models.StepTypeFreeNetworkAddresses {
+		// the free-addresses step's output spams the log too much
+		logFunc("Sending step <%s> reply error <%s> exit-code <%d>", stepID, errStr, exitCode)
+	} else {
+		logFunc("Sending step <%s> reply output <%s> error <%s> exit-code <%d>", stepID, output, errStr, exitCode)
+	}
+
 	params := installer.PostStepReplyParams{
 		HostID:                *CurrentHost.ID,
 		ClusterID:             strfmt.UUID(config.GlobalAgentConfig.ClusterID),


YuviGold:
Unfortunately message like `<[{"free_addresses...>` doesn't tell anything..
In addition I think it would make us problems with logger parsers while there's only an opening double-quotes  without the closing ones.

I would appreciate if the list of free-addresses could be grouped into ranges as I suggested in the ticket.
Best case - The range would have 1 item.
Worst case - The range would be decreased by half.
In this scenario the message would be still informative and readable.

osherdp:
> Unfortunately message like `<[{"free_addresses...>` doesn't tell anything..
> In addition I think it would make us problems with logger parsers while there's only an opening double-quotes without the closing ones.
> 
> I would appreciate if the list of free-addresses could be grouped into ranges as I suggested in the ticket.
> Best case - The range would have 1 item.
> Worst case - The range would be decreased by half.
> In this scenario the message would be still informative and readable.

BTW logging the count of the addresses is not enough?
I'm not afraid from implementing the ranges solution, I just want to state that it seems like a bit of an overkill solution for a log message

YuviGold:
> > Unfortunately message like `<[{"free_addresses...>` doesn't tell anything..
> > In addition I think it would make us problems with logger parsers while there's only an opening double-quotes without the closing ones.
> > I would appreciate if the list of free-addresses could be grouped into ranges as I suggested in the ticket.
> > Best case - The range would have 1 item.
> > Worst case - The range would be decreased by half.
> > In this scenario the message would be still informative and readable.
> 
> BTW logging the count of the addresses is not enough?
> I'm not afraid from implementing the ranges solution, I just want to state that it seems like a bit of an overkill solution for a log message

I agree that it is a bit of an overkill just in logging perspective. I was thinking it should be also the data structure communicating with the assisted-service to reduce the traffic size (this list is being sent a lot of times).
@ori-amizur said that this method wasn't picked, I don't remember why. Ori, can you elaborate?

About a counter - Although I still prefer the ranges one, the counter would be a satisfying clear solution.

osherdp:
@ori-amizur and I talked about it yesterday
I also thought we're going to change it on the API level, but it's a major change of API which shouldn't be made without a good reason
Most importantly the API might change in the future to include assigned-addresses as opposed to free-addresses, as the assigned addresses are acquired by entities that may have other interesting information

YuviGold (approved):
LGTM

YuviGold:
> /hold

@osherdp  Why is this on hold?

PR #54, opened 2020-10-15
MGMR-2420 Add bootable flag to disk in inventory

ori-amizur:
/cc @ronniel1 
/cc @eranco74

### DIFF
--- a/src/inventory/disks.go
+++ b/src/inventory/disks.go
@@ -2,6 +2,7 @@ package inventory
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/jaypipes/ghw"
 	"github.com/openshift/assisted-service/models"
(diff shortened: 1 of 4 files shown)


src/inventory/disks_test.go
```diff
@@ -73,6 +74,53 @@ var _ = Describe("Disks test", func() {
 				SizeBytes: 5555,
 				Vendor:    "disk1-vendor",
 				Wwn:       "disk1-wwn",
+				Bootable:  true,
+			},
+		}))
+	})
```
eranco74, line 80: What about multiple bootable disks?
That is what we actually care about.

src/inventory/disks.go
```diff
@@ -56,6 +57,18 @@ func (d *disks) getHctl(name string) string {
 	return files[0].Name()
 }
 
+func (d *disks) getBootable(path string) bool {
+	if path == "" {
```
eranco74, line 61: When will it get empty path?

ori-amizur: Then we cannot figure it out

src/inventory/disks_test.go
ori-amizur: The information is per disk.  If we have multiple disks, then each of them has such flag

PR #55, opened 2020-10-15
MGMT-2447 Exit next step loop on flag from API

empovit:


osherdp (approved):

src/next_step_runner/main/main.go
```diff
@@ -4,10 +4,12 @@ import (
...
 	"github.com/openshift/assisted-installer-agent/src/config"
 	"github.com/openshift/assisted-installer-agent/src/util"
+	log "github.com/sirupsen/logrus"
 )
 
 func main() {
 	config.ProcessArgs()
 	util.SetLogging("agent_next_step_runner", config.GlobalAgentConfig.TextLogging, config.GlobalAgentConfig.JournalLogging)
 	commands.ProcessSteps()
+	log.Info("Stopping to process steps and exiting")
```
tsorya, line 14: agent will not be just restarted? or this is what u want?

src/commands/step_processor.go
```diff
@@ -93,27 +93,26 @@ func (s *stepSession) processSingleSession() int64 {
 		switch errValue := err.(type) {
 		case *installer.GetNextStepsNotFound:
 			s.Logger().WithError(err).Errorf("Cluster %s was not found in inventory or user is not authorized, going to sleep forever", params.ClusterID)
-			return -1
+			return -1, false
 		case *installer.GetNextStepsUnauthorized:
 			s.Logger().WithError(err).Errorf("User is not authenticated to perform the operation, going to sleep forever")
-			return -1
+			return -1, false
```
tsorya, line 99: maybe we can just return -2 ? create some consts that says sleepForever=-1 Exit=-2 ?

src/commands/step_processor.go
```diff
@@ -93,27 +93,26 @@ func (s *stepSession) processSingleSession() int64 {
...
+	return result.Payload.NextInstructionSeconds, result.Payload.ExitOnCompletion
 }
 
 func ProcessSteps() {
 	var nextRunIn int64
-	for {
+	for exitOnCompletion := false; !exitOnCompletion; {
 		s := newSession()
-		nextRunIn = s.processSingleSession()
+		nextRunIn, exitOnCompletion = s.processSingleSession()
```
tsorya, line 115: why just not to return? if somebody will add by mistake "-1,true" ? Not sure what's better

tsorya: to say the truth that's why o prefer to return one value if possible

src/next_step_runner/main/main.go
empovit: It might be restarted by the process that runs it, if needed. I want next_step_runner to exit when it receives a flag in an server response to do so.

src/commands/step_processor.go
empovit: I'm not sure. I don't like "magic" values, prefer explicit. We could create a special enumeration type, but it would be a mixture of flags and duration - a bit ugly IMO.

src/commands/step_processor.go
empovit: Something will have to take precedence. I replied earlier why I'm against using "magic" sleep time values.

src/commands/step_processor.go
empovit: Also, we might want to wait *and then* exit. It won't be possible with a single value. The agent might need to run a command, allow some time for it to finish (we run commands in the background, so the agent won't block), collect the output and send to the server (post reply). If we don't wait, the agent will exit while the command is still running in a container. So sleep and exit in this case are independent. Unlike sleep for X sec or sleep forever.

ori-amizur: In general the change looks OK to me, but instead a boolean, I think that should be enum that signifies what should be done by the agent. One value is process steps. Another value is exit.  We could add additional values later.

src/commands/step_processor.go
```diff
@@ -104,27 +104,26 @@ func (s *stepSession) processSingleSession() int64 {
...
 		default:
 			s.Logger().WithError(err).Warn("Could not query next steps")
 		}
-		return int64(config.GlobalAgentConfig.IntervalSecs)
-	} else {
-		s.handleSteps(result.Payload)
+		return int64(config.GlobalAgentConfig.IntervalSecs), ""
 	}
-	return result.Payload.NextInstructionSeconds
+	s.handleSteps(result.Payload)
```
ori-amizur, line 118: In case there are steps to handle they are processed in async manner.  It means that the process will not wait for their completion.  Is this what we want?

empovit: Yes, I discussed this with @ronniel1. This problem exists in a couple of other scenarios as well. We're OK with it for now. It's also not that bad because such async command will finish and exit, their container automatically removed by Podman (we run them with `--rm`). The server keeps track of steps and their responses, and will compensate for missing replies. We'll also need to enhance the state machine to know when it's safe to send "exit".

src/next_step_runner/main/main.go
eranco74: Consider changing the log to:
```suggestion
	log.Info("next step runner exiting")
```

eranco74 (commented):
/lgtm
/approve

PR #59, opened 2020-10-20
Revert "MGMT-2069: handle getting 409 from server when host tries to …

tsorya:
…register to a cluster (#52)"

This reverts commit cffd8954c255559fd56495c03f56c7e98c9877e3.


carbonin:
cc @osherdp @ronniel1 @ori-amizur @YuviGold 
Any of you know why the fedora image was changed with the rest of this?
Maybe just an old branch from before we had to move it back or is it somehow related?

Ideally we should just move fedora back to 31.

osherdp:
@carbonin reverted to fedora 31 on #60

PR #62, opened 2020-10-21
MGMT-2213,MGMT-2380 Split agent into registration and next-step parts

empovit:


empovit:
I'm closing this because it's too old and the CI seems to refuse to build it. Opened https://github.com/openshift/assisted-installer-agent/pull/69 instead.

PR #63, opened 2020-10-22
MGMT-2578 use centos 8 as base image

osherdp:
* update base image to centos 8
* not skipping the test with free-addresses verification


eranco74 (commented):
/lgtm

PR #64, opened 2020-10-22
MGMT-2431: Adding 'installer-gather.sh' to logs_sender on failure.

ybettan:
We will now also send to `assisted-service` `installer-gather.sh` logs
on cluster installation cancel of error.

In case the installation succeed, those logs won't be sent, it is
parametrized and can be changed.

### DIFF
--- a/src/logs_sender/send_logs.go
+++ b/src/logs_sender/send_logs.go
@@ -20,6 +20,7 @@ import (
 //go:generate mockery -name LogsSender -inpkg
 type LogsSender interface {
 	Execute(command string, args ...string) (stdout string, stderr string, exitCode int)
+	ExecutePrivilege(command string, args ...string) (stdout string, stderr string, exitCode int)
 	ExecuteOutputToFile(outputFilePath string, command string, args ...string) (stderr string, exitCode int)
 	CreateFolderIfNotExist(folder string) error
 	FileUploader(filePath string, clusterID strfmt.UUID, hostID strfmt.UUID,
@@ -30,6 +31,14 @@ type LogsSenderExecuter struct{}
 
 func (e *LogsSenderExecuter) Execute(command string, args ...string) (stdout string, stderr string, exitCode int) {
 	return util.Execute(command, args...)
+}
+
+// ExecutePrivilege execute a command in the host environment via nsenter
+func (e *LogsSenderExecuter) ExecutePrivilege(command string, args ...string) (stdout string, stderr string, exitCode int) {
+	commandBase := "nsenter"
+	arguments := []string{"-t", "1", "-m", "-i", "--", command}
+	arguments = append(arguments, args...)
+	return e.Execute(commandBase, arguments...)
 }
 
 func (e *LogsSenderExecuter) ExecuteOutputToFile(outputFilePath string, command string, args ...string) (stderr string, exitCode int) {
@@ -58,10 +67,10 @@ func (e *LogsSenderExecuter) FileUploader(filePath string, clusterID strfmt.UUID
 	}
 
 	params := installer.UploadHostLogsParams{
-		Upfile:    uploadFile,
-		ClusterID: clusterID,
+		Upfile:                uploadFile,
+		ClusterID:             clusterID,
 		DiscoveryAgentVersion: &agentVersion,
-		HostID:    hostID,
+		HostID:                hostID,
 	}
 	_, err = invSession.Client().Installer.UploadHostLogs(invSession.Context(), &params)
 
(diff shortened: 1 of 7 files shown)


src/logs_sender/send_logs.go
```diff
@@ -142,6 +151,22 @@ func SendLogs(l LogsSender) error {
 		}
 	}
 
+	if withInstallerGatherlogging && isBootstrap {
+		log.Info("Running installer-gather.sh")
+		_, stdErr, exitCode := l.ExecutePrivilege("installer-gather.sh")
+		if exitCode != 0 {
+			err := errors.Errorf(stdErr)
+			log.WithError(err).Errorf("Failed to run installer-gather.sh command: %v", err)
```
tsorya, line 159: u are already logging error, no need to do it twice.

src/logs_sender/send_logs.go
```diff
@@ -142,6 +151,22 @@ func SendLogs(l LogsSender) error {
...
+			return err
+		}
+		_, stdErr, exitCode = l.ExecutePrivilege("mv", "/root/log-bundle-.tar.gz", fmt.Sprintf("%s/installer_gather.tar.gz", logsTmpFilesDir))
```
tsorya, line 162: why not to have it as part gather logs command? just with && ?

src/logs_sender/send_logs.go
```diff
@@ -142,6 +151,22 @@ func SendLogs(l LogsSender) error {
...
+		if exitCode != 0 {
+			err := errors.Errorf(stdErr)
+			log.WithError(err).Errorf("Failed to mv log-bundle-.tar.gz: %v", err)
```
tsorya, line 165: same here, u use log.WithError(err) no need to add %v, err

src/logs_sender/main/main.go
```diff
@@ -13,7 +13,7 @@ import (
 func main() {
 	config.ProcessLogsSenderConfigArgs(true, true)
 	util.SetLogging("logs-sender", config.LogsSenderConfig.TextLogging, config.LogsSenderConfig.JournalLogging)
-	err := logs_sender.SendLogs(&logs_sender.LogsSenderExecuter{})
+	err := logs_sender.SendLogs(&logs_sender.LogsSenderExecuter{}, config.LogsSenderConfig.IsBootstrap, config.LogsSenderConfig.InstallerGatherlogging)
```
tsorya, line 16: u don't need onfig.LogsSenderConfig.IsBootstrap, config.LogsSenderConfig.InstallerGatherlogging to be provided.
look on how all other config values are used

ybettan: Done.

src/logs_sender/send_logs.go
ybettan: I can but then we won't get different error messages for the different errors.

src/logs_sender/send_logs.go
ybettan: Done

src/logs_sender/send_logs.go
```diff
@@ -142,6 +151,22 @@ func SendLogs(l LogsSender) error {
 		}
 	}
 
+	if withInstallerGatherlogging && isBootstrap {
+		log.Info("Running installer-gather.sh")
+		_, stdErr, exitCode := l.ExecutePrivilege("installer-gather.sh")
```
romfreiman, line 156: mm full path?

src/logs_sender/send_logs.go
romfreiman: what is this '-' in the filename?

src/logs_sender/send_logs.go
tsorya: agree

src/logs_sender/send_logs.go
tsorya: i think it will be enough to get error from one command, if u do it with && u will get relevant error in any case

tsorya: cause second one will not run if the first one failed

ybettan: Ack, I will do it

ybettan: This is happening because if failed to gather other masters data (it should be followed by the master names), it will be updated on the second phase, when we add ssh key to ignition and gather master's logs in addition to bootstrap's logs.
@romfreiman

src/logs_sender/send_logs.go
ybettan: This is a command located in `/usr/local/bin`, do you still thing we need full path?

ybettan: Done.

PR #68, opened 2020-11-02
MGMT-2440 Fix the errors in inventory when trying to get BMC information

ori-amizur:
- Loading kernel modules in a container is meaningless and produces errors in logs - so it was removed
/cc @ronniel1 
/cc @tsorya


ronniel1 (commented):
@ori-amizur  why did we try to install the ipmi tools?
Don't we use them?

ronniel1 (approved):
/lgtm

PR #69, opened 2020-11-03
MGMT-2213,MGMT-2380 Bring back split into registration and next step

empovit:
The agent systemd service has been changed to only register its host with assisted service. The service replies with a command for running the core part of the agent as a container, so that its image is always up to date. This core part polls for next steps and posts replies as before. 

Also, README has been updated and cleaned up.


osherdp (approved):

ori-amizur (commented):
/approve
/lgtm

PR #83, opened 2020-11-09
MGMT-2322 Add ntp_synchronizer executable

YuviGold:
/cc @ori-amizur @ronniel1


subsystem/ntp_test.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	var response models.NtpSynchronizationResponse
+	err := json.Unmarshal([]byte(actualReply.Output), &response)
+	if err != nil {
+		log.Errorf("NTPSynchronizerVerifier failed to unmarshal")
+		return false
+	}
+
+	for _, source := range response.NtpSources {
+		fmt.Println(*source)
```
ori-amizur, line 100: Remove Println

subsystem/ntp_test.go
```diff
@@ -0,0 +1,148 @@
...
+		if source.SourceState == models.SourceStateSynced {
```
ori-amizur, line 101: The veirifier should just identify the reply.  It should not check states

src/ntp_synchronizer/ntp_synchronizer.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/openshift/assisted-service/models"
+	"github.com/pkg/errors"
+
+	"github.com/sirupsen/logrus"
+)
+
+const ChronyTimeoutSeconds = 10
+
+//go:generate mockery -name INTPDependencies -inpkg
+type INTPDependencies interface {
```
ori-amizur, line 18: The name is a bit strange - maybe more meaningful ?

src/ntp_synchronizer/ntp_synchronizer.go
```diff
@@ -0,0 +1,119 @@
...
+
+	if exitCode == 0 {
+		return nil
+	} else {
+		return errors.Errorf("chronyc exited with non-zero exit code %d: %s\n%s", exitCode, stdout, stderr)
+	}
+}
+
+func GetNTPSources(e INTPDependencies) (*[]*models.NtpSource, error) {
+	stdout, stderr, exitCode := e.Execute("timeout", strconv.FormatInt(ChronyTimeoutSeconds, 10), "chronyc", "-c", "sources")
```
ori-amizur, line 58: Why this command with timeout and the previous not? Is the timeout needed?

src/ntp_synchronizer/ntp_synchronizer.go
```diff
@@ -0,0 +1,119 @@
...
+	case 0:
+		return &sources, nil
+	case util.TimeoutExitCode:
+		return nil, errors.Errorf("chronyc was timed out after %d seconds", ChronyTimeoutSeconds)
+	default:
+		return nil, errors.Errorf("chronyc exited with non-zero exit code %d: %s\n%s", exitCode, stdout, stderr)
+	}
+}
+
+func NtpSync(ntpSyncRequestStr string, executer INTPDependencies, log logrus.FieldLogger) (stdout string, stderr string, exitCode int) {
```
ori-amizur, line 84: Maybe create a struct with functions that the executer is a member?

src/ntp_synchronizer/ntp_synchronizer.go
```diff
@@ -0,0 +1,119 @@
...
+		return errors.Errorf("chronyc exited with non-zero exit code %d: %s\n%s", exitCode, stdout, stderr)
+	}
+}
+
+func GetNTPSources(e INTPDependencies) (*[]*models.NtpSource, error) {
+	stdout, stderr, exitCode := e.Execute("timeout", strconv.FormatInt(ChronyTimeoutSeconds, 10), "chronyc", "-c", "sources")
+
+	sources := make([]*models.NtpSource, 0)
+
+	for _, line := range strings.Split(stdout, "\n") {
```
ori-amizur, line 62: It makes more sense to first handle the exitCode and if it is OK format the sources.  The formatting can be in its own function

src/ntp_synchronizer/ntp_synchronizer.go
YuviGold: I was inspired by `Dependencies` from `dhcp_lease_allocator`.

src/ntp_synchronizer/ntp_synchronizer.go
YuviGold: Not really. I just think it is safer to add a timeout. WDYT?

src/ntp_synchronizer/ntp_synchronizer.go
ori-amizur: INTPDependecies is a bit strange (I think)

YuviGold: Ok changing to `NtpSynchronizerDependencies`

src/ntp_synchronizer/ntp_synchronizer.go
YuviGold: What for? I can just replace the executer, I don't have any other functions

YuviGold:
@gamli75 Where do I update the downstream image?

subsystem/ntp_test.go
```diff
@@ -0,0 +1,168 @@
...
+)
+
+const (
+	stepNTPID = "ntp-synchronizer-step"
+)
+
+var _ = Describe("NTP tests", func() {
+	var (
+		hostID          string
+		numberOfSources int
```
osherdp, line 20: why should it be global for the whole describe scope, and not used in this scope?
can be local variable in each ``It``

subsystem/ntp_test.go
```diff
@@ -0,0 +1,168 @@
...
+	})
+
+	It("add_server", func() {
+		server := "1.1.1.1"
+
+		setNTPSyncRequestStub(hostID, models.NtpSynchronizationRequest{NtpSource: &server})
+
+		ntpResponse := getNTPResponse(hostID)
+		Expect(ntpResponse).ShouldNot(BeNil())
+		Expect(len(ntpResponse.NtpSources)).Should(BeNumerically(">", numberOfSources))
```
osherdp, line 48: Cannot be asserted against a concrete value?

osherdp (changes requested):

src/ntp_synchronizer/main/main.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/openshift/assisted-installer-agent/src/util"
+	log "github.com/sirupsen/logrus"
+)
+
+func main() {
+	config.ProcessSubprocessArgs(config.DefaultLoggingConfig)
+	util.SetLogging("ntp_synchronizer", config.SubprocessConfig.TextLogging, config.SubprocessConfig.JournalLogging)
+	if flag.NArg() != 1 {
+		log.Warnf("Expecting exactly single argument to ntp_synchronizer. Received %d", len(os.Args)-1)
+		os.Exit(-1)
```
empovit, line 19: Why not use log.Fatalf() whose purpose seems exactly "cannot continue, print error and exit"? https://github.com/sirupsen/logrus/blob/d131c24e23baaa812461202af6d7cfa388e2d292/logger.go#L186

src/ntp_synchronizer/ntp_synchronizer.go
```diff
@@ -0,0 +1,162 @@
...
+	stdout, stderr, exitCode := e.Execute("chronyc", "add", "server", ntpSource, "iburst")
+
+	if exitCode == 0 {
+		return nil
+	} else {
+		return errors.Errorf("chronyc exited with non-zero exit code %d: %s\n%s", exitCode, stdout, stderr)
+	}
+}
+
+func formatChronySourcesOutput(output string) *[]*models.NtpSource {
```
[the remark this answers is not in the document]
YuviGold: In order to return `nil` and being able to validate it.
In this situation where I'm also retrieving an error (for `GetNTPSources`) maybe it isn't necessary.

src/ntp_synchronizer/main/main.go
empovit: It could also help to tell what the program expects, "exactly single argument" is too general (I know we do it in other commands, too).

subsystem/ntp_test.go
```diff
@@ -0,0 +1,168 @@
...
+		setReplyStartAgent(hostID)
+		waitforChronyDaemonToStart(hostID)
+	})
+
+	It("get_sources", func() {
+		setNTPSyncRequestStub(hostID, models.NtpSynchronizationRequest{})
+
+		ntpResponse := getNTPResponse(hostID)
+		Expect(ntpResponse).ShouldNot(BeNil())
+		numberOfSources = len(ntpResponse.NtpSources)
```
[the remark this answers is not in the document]
YuviGold: Because the NTP executable on subsystem runs with `--net=host` I'm unable to know the amount of expected NTP sources (it is dependent on the running machine. It might be zero).
So I just wanted to make sure I can get a response, and in the next test I'm verifying that when adding a new server - the amount of NTP sources is increased.

subsystem/ntp_test.go
YuviGold: Explained in the previous comment.

subsystem/ntp_test.go
YuviGold: Explained in the previous comment.

src/ntp_synchronizer/main/main.go
YuviGold: As you said, I was just copying other commands' behavior. But your suggestion seems cleaner :) 
@ori-amizur Do we care if the exit error would be `-1` as we currently do or `1` as returned by `log.Fatalf()`?

subsystem/ntp_test.go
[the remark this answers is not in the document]
osherdp: then use _ so it will be clear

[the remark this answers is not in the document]
osherdp: So there is dependencies between the tests?
I don't think it's good...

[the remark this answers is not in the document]
empovit: I agree with @osherdp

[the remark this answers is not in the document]
YuviGold: Alright

subsystem/ntp_test.go
```diff
@@ -0,0 +1,168 @@
...
+		return isReplyFound(hostID, &NTPSynchronizerVerifier{})
+	}, 30*time.Second, 5*time.Second).Should(BeTrue())
+
+	stepReply := getSpecificStep(hostID, &NTPSynchronizerVerifier{})
+	return getNTPResponseFromStepReply(stepReply)
+}
+
+type NTPSynchronizerVerifier struct{}
+
+func (i *NTPSynchronizerVerifier) verify(actualReply *models.StepReply) bool {
```
osherdp, line 110: We need in other issue to replace those verifiers with ones that return errors
It's bad that we separate between validation failure and it's reason (logs are not displayed as the bottom line)

osherdp (changes requested):

subsystem/ntp_test.go
ori-amizur: Don't agree.  Verifiers are just for indication if a message is from a particular type.  error is not relevant here

src/ntp_synchronizer/main/main.go
ori-amizur: We need to output to stderr in this case.  This way, it will be logged and transferred to assisted-service.

src/ntp_synchronizer/ntp_synchronizer.go
ori-amizur: Do you want to fail it on timeout?

subsystem/ntp_test.go
osherdp: We're writing logs and returning a general boolean indication that something bad happened. The first thing being shown if the test fails is that generic message, and the real root cause needs to be then tracked
Why not just error or alternatively use the same assertion-like mechanism of ginkgo?

src/ntp_synchronizer/main/main.go
ori-amizur: @YuviGold Fatalf is OK.  Just make sure to output the error to stderr.

src/dhcp_lease_allocate/lease.go
```diff
@@ -50,7 +51,7 @@ func LeaseVIP(d Dependencies, log logrus.FieldLogger, leaseFile, masterDevice, n
 	switch exitCode {
 	case 0:
 		return nil
-	case 124:
+	case util.TimeoutExitCode:
```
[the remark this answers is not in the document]
YuviGold: Previously we resolved the hostnames, now it's only IPs.
Anyway, I think it is better to wrap a timeout scenario just in case.

subsystem/ntp_test.go
YuviGold: I agree with @osherdp to propagate the error instead of boolean. But it would require to change all of the interface implementations so that's for another PR. :)

src/ntp_synchronizer/ntp_synchronizer.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/pkg/errors"
+	"github.com/thoas/go-funk"
+
+	"github.com/sirupsen/logrus"
+)
+
+const ChronyTimeoutSeconds = 10
+
+//go:generate mockery -name NtpSynchronizerDependencies -inpkg
+type NtpSynchronizerDependencies interface {
```
carbonin, line 20: An interface should be named after what it does. So in this case I would suggest `Executor`.

Also all of this seems like it shouldn't be the ntp-specific package's problem, I think the `util` package should provide this interface.

src/ntp_synchronizer/ntp_synchronizer.go
```diff
@@ -0,0 +1,162 @@
...
+			continue
+		}
+
+		sources = append(sources, &models.NtpSource{SourceName: fields[1], SourceState: convertSourceState(string(fields[0][1]))})
+	}
+
+	return sources
+}
+
+func GetNTPSources(e NtpSynchronizerDependencies) ([]*models.NtpSource, error) {
```
carbonin, line 81: Why is this exported?

src/ntp_synchronizer/ntp_synchronizer.go
```diff
@@ -0,0 +1,162 @@
...
+		if funk.Contains(names, source.SourceName) {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+func NtpSync(ntpSyncRequestStr string, executer NtpSynchronizerDependencies, log logrus.FieldLogger) (stdout string, stderr string, exitCode int) {
+	var ntpSyncRequest models.NtpSynchronizationRequest
```
carbonin, line 125: These variable names could be simplified. We're in a function called `NtpSync` I don't think all the variables should be prefixed with `ntpSync`.

`requestStr` and `request` would be better.
Also `s/executer/executor/`

src/ntp_synchronizer/ntp_synchronizer.go
```diff
@@ -0,0 +1,162 @@
...
+
+	err := json.Unmarshal([]byte(ntpSyncRequestStr), &ntpSyncRequest)
+	if err != nil {
+		log.WithError(err).Errorf("NTPSync: json.Unmarshal")
```
carbonin, line 129: Can these be more like sentences? Maybe something like:

```suggestion
		log.WithError(err).Errorf("Failed to unmarshal ntp request string %s", ntpSyncRequestStr)
```

Same goes for pretty much every error message in this function

src/ntp_synchronizer/ntp_synchronizer.go
```diff
@@ -0,0 +1,162 @@
...
+	for _, source := range sources {
+		if funk.Contains(names, source.SourceName) {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+func NtpSync(ntpSyncRequestStr string, executer NtpSynchronizerDependencies, log logrus.FieldLogger) (stdout string, stderr string, exitCode int) {
```
carbonin, line 124: We only ever return `0` or `-1` for `exitCode`, can we remove it and decide what code to use in the caller?

carbonin: I think a more standard go return value would be `(string, error)`.
Then in the caller we would print the error text to stderr (if we got an error) and the string to stdout.

src/ntp_synchronizer/ntp_synchronizer.go
YuviGold: Yes

Makefile
```diff
@@ -2,21 +2,22 @@ TAG := $(or $(TAG),latest)
 ASSISTED_INSTALLER_AGENT := $(or $(ASSISTED_INSTALLER_AGENT),quay.io/ocpmetal/assisted-installer-agent:$(TAG))
 
 export ROOT_DIR = $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
+BIN = $(ROOT_DIR)/build
 REPORTS = $(ROOT_DIR)/reports
 GOTEST_PUBLISH_FLAGS = --junitfile-testsuite-name=relative --junitfile-testcase-classname=relative --junitfile $(REPORTS)/$(TEST_SCENARIO)_test.xml
-GOTEST_FLAGS = --format=pkgname $(GOTEST_PUBLISH_FLAGS) -- -count=1 -cover -coverprofile=$(REPORTS)/$(TEST_SCENARIO)_coverage.out
+GOTEST_FLAGS = --format=standard-verbose $(GOTEST_PUBLISH_FLAGS) -- -count=1 -cover -coverprofile=$(REPORTS)/$(TEST_SCENARIO)_coverage.out
```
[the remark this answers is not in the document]
YuviGold: Its output is more verbose.

src/ntp_synchronizer/ntp_synchronizer.go
YuviGold: No certain reason. Just thought it might be useable in the future as a utility.
But I guess it would never be used, so I changed it back to be private.

src/ntp_synchronizer/ntp_synchronizer.go
YuviGold: This is how we create the dependencies mostly to be generated for mocks in the UT.

src/ntp_synchronizer/ntp_synchronizer.go
YuviGold: Just copied the skeleton from another utility. :) I agree with you. I'll make these changes on a different PR.

src/ntp_synchronizer/main/main.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/openshift/assisted-installer-agent/src/config"
+	"github.com/openshift/assisted-installer-agent/src/ntp_synchronizer"
+	"github.com/openshift/assisted-installer-agent/src/util"
+	log "github.com/sirupsen/logrus"
+)
+
+func main() {
+	config.ProcessSubprocessArgs(config.DefaultLoggingConfig)
+	util.SetLogging("ntp_synchronizer", config.SubprocessConfig.TextLogging, config.SubprocessConfig.JournalLogging)
+	if flag.NArg() != 1 {
```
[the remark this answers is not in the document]
YuviGold: I created a JIRA ticket [MGMT-2961](https://issues.redhat.com/browse/MGMT-2961) with utility enhancements.

PR #84, opened 2020-11-11
MGMT-2470 wait for all step goroutines to finish before asking for more steps

omertuc:
[Ronnie suggested](https://issues.redhat.com/browse/MGMT-2470?focusedCommentId=15374668&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15374668) we suppress the message, but from what I've seen simply suppressing this specific message is not trivial because it's raised by a generic command runner on the agent that sends a generic StepReply back to the service with the error, the server in turn logs that error message and this is what we see in those logs. I believe suppressing this specific message means giving it special treatment in this generic code and that would be a bit awkward.

I think the root of the issue comes from [this line](https://github.com/openshift/assisted-installer-agent/blob/3aea65b0ba71f6634939e6233184445066bd93a6/src/commands/step_processor.go#L91) that's running the steps received from the service concurrently. I think it makes sense for the function to `WaitGroup.Wait` on the these goroutines and let them finish, instead of what it does right now which is simply wait the server-dictated timeout and then ask for more commands which it then tries to run concurrently while the previous commands may still be running (which is also the source of the error - the previous free_addresses command is running for longer than a minute as Ori Amizur said and we can't create another container with the same name).

This is a draft PR that shows what I mean, please let me know what you think @ori-amizur @YuviGold

### DIFF
--- a/src/commands/step_processor.go
+++ b/src/commands/step_processor.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"net/http"
+	"sync"
 	"time"
 
 	"github.com/go-openapi/swag"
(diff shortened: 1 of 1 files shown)


omertuc:
Draft no longer irrelevant, a better solution was suggested. Will open a PR with the new solution

PR #89, opened 2020-11-17
MGMT-792 - Updated assisted-service, replaced lshw with ghw, populated is-vm field

omertuc:
Continuing https://github.com/openshift/assisted-service/pull/688

- Updated assisted-service to the latest version that includes the PR above

- Replaced lshw with simpler ghw (I saw lshw outputting invalid JSON on some platforms, and while it may be a problem with the lshw version on those platforms and not in the version we put in our container, out of fear that some hardware causes it to misbehave, I thought it was best to just replace it with something simpler like ghw which we already depend on and use in other places in this code)

- Populated new models.SystemVendor.Virtual field with true/false according to whether the ProductName field looks like a known VM technology

### DIFF
--- a/src/inventory/system_vendor.go
+++ b/src/inventory/system_vendor.go
@@ -1,33 +1,42 @@
 package inventory
 
 import (
-	"encoding/json"
-
+	"github.com/jaypipes/ghw"
 	"github.com/openshift/assisted-service/models"
 	"github.com/sirupsen/logrus"
+
+	"strings"
 )
 
-type systemVendor struct {
-	Product string
-	Serial  string
-	Vendor  string
+func isVirtual(product string) bool {
+	for _, vmTech := range []string {
+		"KVM",
+		"VirtualBox",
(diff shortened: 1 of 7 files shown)


src/inventory/system_vendor.go
```diff
@@ -1,33 +1,42 @@
...
-	o, e, exitCode := dependencies.Execute("lshw", "-quiet", "-json")
 	var ret models.SystemVendor
-	if exitCode != 0 {
-		logrus.Errorf("Error running lshw:: %s", e)
-		return &ret
-	}
-	var sv systemVendor
-	err := json.Unmarshal([]byte(o), &sv)
+
+	product, err := dependencies.Product(ghw.WithChroot("/host"))
```
eranco74, line 29: Nice

eranco74 (commented):
Looks good

ori-amizur:
If lshw is not needed, please remove it from the dockerfile

PR #90, opened 2020-11-18
MGMT-2678: Supplying masters host names to installer-gather.sh.

ybettan:
If not supplied, `installer-gather.sh` will try to get this information
using `oc get nodes` but the use case we are trying to get logs for is
the one in which part of the masters didn't join the cluster so `oc get
nods` won't work.

Masters host names are retrieved from `assisted-service` API.

Also added gatherID to the log-bundle created.

### DIFF
--- a/src/logs_sender/send_logs.go
+++ b/src/logs_sender/send_logs.go
@@ -5,6 +5,8 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strings"
+	"time"
 
 	"github.com/openshift/assisted-installer-agent/src/config"
 	"github.com/pkg/errors"
@@ -17,14 +19,21 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const (
+	logsDir                      = "/var/log"
+	installerGatherBin           = "/usr/local/bin/installer-gather.sh"
+	installerGatherArchivePreifx = "/root/log-bundle-"
+)
+
 //go:generate mockery -name LogsSender -inpkg
 type LogsSender interface {
 	Execute(command string, args ...string) (stdout string, stderr string, exitCode int)
-	ExecutePrivilege(command string, args ...string) (stdout string, stderr string, exitCode int)
+	ExecutePrivileged(command string, args ...string) (stdout string, stderr string, exitCode int)
 	ExecuteOutputToFile(outputFilePath string, command string, args ...string) (stderr string, exitCode int)
 	CreateFolderIfNotExist(folder string) error
 	FileUploader(filePath string, clusterID strfmt.UUID, hostID strfmt.UUID,
 		inventoryUrl string, pullSecretToken string, agentVersion string) error
+	GatherInstallerLogs(targetDir string) error
 }
 
 type LogsSenderExecuter struct{}
@@ -33,12 +42,9 @@ func (e *LogsSenderExecuter) Execute(command string, args ...string) (stdout str
 	return util.Execute(command, args...)
 }
 
-// ExecutePrivilege execute a command in the host environment via nsenter
-func (e *LogsSenderExecuter) ExecutePrivilege(command string, args ...string) (stdout string, stderr string, exitCode int) {
-	commandBase := "nsenter"
-	arguments := []string{"-t", "1", "-m", "-i", "-n", "--", command}
-	arguments = append(arguments, args...)
-	return e.Execute(commandBase, arguments...)
+// ExecutePrivileged execute a command in the host environment via nsenter
+func (e *LogsSenderExecuter) ExecutePrivileged(command string, args ...string) (stdout string, stderr string, exitCode int) {
+	return util.ExecutePrivileged(command, args...)
 }
 
 func (e *LogsSenderExecuter) ExecuteOutputToFile(outputFilePath string, command string, args ...string) (stderr string, exitCode int) {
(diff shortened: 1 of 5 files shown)


src/logs_sender/send_logs.go
```diff
@@ -77,7 +85,53 @@ func (e *LogsSenderExecuter) FileUploader(filePath string, clusterID strfmt.UUID
...
 
-const logsDir = "/var/log"
+func (e *LogsSenderExecuter) GetNonBootstrapMasters() ([]string, error) {
+
+	invSession, err := session.New(config.LogsSenderConfig.TargetURL, config.LogsSenderConfig.PullSecretToken)
+	if err != nil {
+		log.Errorf("Failed to initialize connection to service: %e", err)
+		return []string{}, err
+	}
+	listHostsOk, err := invSession.Client().Installer.ListHosts(context.Background(),
```
tsorya, line 95: why not to provide all data from outside? I think much easier will be just to have flags for master ips and provide it from the service

src/logs_sender/send_logs.go
```diff
@@ -77,7 +85,53 @@ func (e *LogsSenderExecuter) FileUploader(filePath string, clusterID strfmt.UUID
...
+		if host.Role == models.HostRoleMaster && host.ID.String() != config.LogsSenderConfig.HostID {
+			res = append(res, host.RequestedHostname)
+		}
+	}
+	return res, nil
+}
+
+func (e *LogsSenderExecuter) GatherInstallerLogs(targetDir string) error {
+	log.Info("Getting non-bootstrap master nodes from service API")
+	listMasters, err := e.GetNonBootstrapMasters()
```
tsorya, line 113: and why hostname? maybe while running on the same l2 dhcp will provide relevant dns resolution but while running on l3 it will not, I think we need to be able to get ip.

src/logs_sender/send_logs.go
```diff
@@ -77,7 +85,53 @@ func (e *LogsSenderExecuter) FileUploader(filePath string, clusterID strfmt.UUID
...
+	if err != nil {
+		log.WithError(err).Error("Failed to get non-bootstrap masters")
+		return err
+	}
+	//FIXME:ybettan: Format(layout?)
+	gatherID := time.Now().Format("20060102150405")
```
tsorya, line 119: just to understand what is that id?

src/logs_sender/send_logs.go
```diff
@@ -77,7 +85,53 @@ func (e *LogsSenderExecuter) FileUploader(filePath string, clusterID strfmt.UUID
...
+	installerGatherArgs := append([]string{"--id", gatherID}, listMasters...)
+	log.Infof("Running %s %v", installerGatherBin, installerGatherArgs)
+	// installer-gather.sh is written in such a way it always return 0.
+	stdOut, stdErr, _ := e.ExecutePrivileged(installerGatherBin, installerGatherArgs...)
+	log.Info(stdOut)
```
tsorya, line 124: is stdout long? do we need to log it? not sure why

src/logs_sender/send_logs.go
```diff
@@ -77,7 +85,53 @@ func (e *LogsSenderExecuter) FileUploader(filePath string, clusterID strfmt.UUID
...
+	if stdErr != "" {
+		log.WithError(err).Warnf("Failed to run %s %v", installerGatherBin, installerGatherArgs)
```
tsorya, line 126: what is err here? maybe stdErr?

src/logs_sender/send_logs.go
```diff
@@ -77,7 +85,53 @@ func (e *LogsSenderExecuter) FileUploader(filePath string, clusterID strfmt.UUID
...
+	}
+	_, stdErr, exitCode := e.ExecutePrivileged("mv", installerGatherArchivePreifx+gatherID+".tar.gz", targetDir)
```
tsorya, line 128: maybe fmt.Sprintf is better that concatenation ?

src/logs_sender/send_logs.go
ybettan: This is the ID of archive created for log-bundle (output of installer-gather.sh).
This is the way it is used in installer: https://github.com/openshift/installer/blob/dfdc5ef6361b649dc8229afd5189d1f3e2ffc3a9/cmd/openshift-install/gather.go#L155-L156

If we don't supply any ID we will get an archive with the name `log-bundle-.tar.gz`

src/logs_sender/send_logs.go
ybettan: Not long:
```Gathering bootstrap systemd summary ...\nGathering bootstrap failed systemd unit status ...\nGathering bootstrap journals ...\nGathering bootstrap containers ...\nGathering rendered assets...\nGathering cluster resources ...\nWaiting for logs ...\nGather remote logs\nCollecting info from test-infra-cluster-assisted-installer-master-1\nGathering master systemd summary ...\nGathering master failed systemd unit status ...\nGathering master journals ...\nGathering master containers ...\nWaiting for logs ...\nCollecting info from test-infra-cluster-assisted-installer-master-0\nGathering master systemd summary ...\nGathering master failed systemd unit status ...\nGathering master journals ...\nGathering master containers ...\nWaiting for logs ...\nLog bundle written to /root/log-bundle-20201118073800.tar.gz\n```

I think it is informational data, don't you think so?!

src/logs_sender/send_logs.go
ybettan: This was a typo, was updated.

src/logs_sender/send_logs.go
ybettan: done.

src/logs_sender/send_logs.go
tsorya: ok, not sure if u need to parse it

src/logs_sender/send_logs.go
ybettan: done.

src/logs_sender/send_logs.go
ybettan: Switched to IPs.

PR #98, opened 2020-11-24
MGMT-2951 - Will log reasons for why disks were skipped

omertuc:
We have many reasons to skip disks as installation candidates - we will now log all the reasons why skip a certain disk when we do, to aid in troubleshooting installation problems

### DIFF
--- a/src/inventory/disks.go
+++ b/src/inventory/disks.go
@@ -2,11 +2,11 @@ package inventory
 
 import (
 	"fmt"
-	"strings"
-
 	"github.com/jaypipes/ghw"
 	"github.com/openshift/assisted-service/models"
 	"github.com/sirupsen/logrus"
+	"github.com/thoas/go-funk"
+	"strings"
 )
 
 type disks struct {
@@ -92,15 +92,37 @@ func unknownToEmpty(value string) string {
 	return value
 }
 
-// isInstallationMedia returns whether this disk is an ISO media or not
-func isInstallationMedia(d *ghw.Disk) bool {
-	for _, p := range d.Partitions {
-		if p.Type == "iso9660" || strings.HasSuffix(p.MountPoint, "iso") {
-			return true
-		}
+// diskIsRelevant checks if a disk is relevant for installation by testing
+// it against a list of predicates. Also returns all the reasons the disk
+// was found to be irrelevant
+func diskIsRelevant(disk *ghw.Disk) (relevant bool, skipReasons []string) {
+	if disk.IsRemovable {
+		skipReasons = append(skipReasons, "Disk is removable")
 	}
 
-	return false
+	if disk.SizeBytes == 0 {
+		skipReasons = append(skipReasons, "Disk has size 0")
+	}
+
+	if disk.BusType == ghw.BUS_TYPE_UNKNOWN && disk.StorageController == ghw.STORAGE_CONTROLLER_UNKNOWN {
+		skipReasons = append(skipReasons, "Disk has unknown bus type and storage controller")
+	}
+
+	if funk.Contains(funk.Map(disk.Partitions, func(p *ghw.Partition) bool {
+		return p.Type == "iso9660"
+	}), true) {
+		skipReasons = append(skipReasons, "Disk appears to be an ISO installation media (has partition with type iso9660)")
+	}
+
+	if funk.Contains(funk.Map(disk.Partitions, func(p *ghw.Partition) bool {
+		return strings.HasSuffix(p.MountPoint, "iso")
+	}), true) {
+		skipReasons = append(skipReasons, "Disk appears to be an ISO installation media (has partition with mountpoint suffix iso)")
+	}
+
+	relevant = len(skipReasons) == 0
+
+	return
 }
 
 func (d *disks) getDisks() []*models.Disk {
@@ -111,8 +133,12 @@ func (d *disks) getDisks() []*models.Disk {
 		return ret
 	}
 	for _, disk := range blockInfo.Disks {
-		if disk.IsRemovable || disk.SizeBytes == 0 || isInstallationMedia(disk) ||
-			(disk.BusType == ghw.BUS_TYPE_UNKNOWN && disk.StorageController == ghw.STORAGE_CONTROLLER_UNKNOWN) {
+		isRelevant, reasons := diskIsRelevant(disk)
+		if !isRelevant {
+			reasonsLines := strings.Join(reasons, "\n")
+			logrus.Infof(
+				"Disk (name %s bus path %s vendor %s model %s partitions %s) was found to be irrelevant for the following reasons:\n %s",
+				disk.Name, disk.BusPath, disk.Vendor, disk.Model, disk.Partitions, reasonsLines)
 			continue
 		}
 


src/inventory/disks.go
```diff
@@ -95,8 +121,12 @@ func (d *disks) getDisks() []*models.Disk {
 		return ret
 	}
 	for _, disk := range blockInfo.Disks {
-		if disk.IsRemovable || disk.SizeBytes == 0 || isInstallationMedia(disk) ||
-			(disk.BusType == ghw.BUS_TYPE_UNKNOWN && disk.StorageController == ghw.STORAGE_CONTROLLER_UNKNOWN) {
+		isRelevant, reasons := diskIsRelevant(disk)
+		if !isRelevant {
+			reasonsLines := strings.Join(reasons, "\n")
+			logrus.Infof(
+				"Disk (name %s bus path %s vendor %s model %s) was found to be irrelevant for the following reasons:\n %s",
```
osherdp, line 128: Maybe you can get along with only ``%s`` the disk model? I remember most of the information is there
But check me, I might be wrong

src/inventory/disks.go
```diff
@@ -87,6 +87,32 @@ func isInstallationMedia(d *ghw.Disk) bool {
...
 
+// diskIsRelevant checks if a disk is relevant for installation by testing
+// it against a list of predicates. Also returns all the reasons the disk
+// was found to be irrelevant
+func diskIsRelevant(disk *ghw.Disk) (relevant bool, reasons []string) {
+	for _, check := range []struct {
+		predicate func() bool
+		reason    string
+	}{
+		{func() bool { return disk.IsRemovable }, "Disk is removeable"},
```
osherdp, line 98: Maybe replace the functions usage with something like
```
if disk.IsRemovable {
    disqualifyingReasons = append(disqualifyingReasons, "Disk is removable")
}
if ...
return disqualifyingReasons
```

src/inventory/disks.go
```diff
@@ -95,8 +121,12 @@ func (d *disks) getDisks() []*models.Disk {
 		return ret
 	}
 	for _, disk := range blockInfo.Disks {
-		if disk.IsRemovable || disk.SizeBytes == 0 || isInstallationMedia(disk) ||
-			(disk.BusType == ghw.BUS_TYPE_UNKNOWN && disk.StorageController == ghw.STORAGE_CONTROLLER_UNKNOWN) {
+		isRelevant, reasons := diskIsRelevant(disk)
+		if !isRelevant {
+			reasonsLines := strings.Join(reasons, "\n")
+			logrus.Infof(
```
osherdp, line 127: replace with debug

src/inventory/disks.go
```diff
@@ -95,8 +121,12 @@ func (d *disks) getDisks() []*models.Disk {
...
+				"Disk (name %s bus path %s vendor %s model %s) was found to be irrelevant for the following reasons:\n %s",
+				disk.Name, disk.BusPath, disk.Vendor, disk.Model, reasonsLines)
```
osherdp, line 129: Is it possible to squeeze in partitions data as well?

src/inventory/disks.go
```diff
@@ -87,6 +87,32 @@ func isInstallationMedia(d *ghw.Disk) bool {
 	return false
```
osherdp, line 87: We should consider splitting the conditions of fstype and mountpoint, I'm not happy with what I did :)

osherdp (changes requested):

osherdp:
also add some tests

src/inventory/disks.go
omertuc: What's the reason? I assumed debug logs are turned off by default. Do we usually have debug logs enabled/stored? This seems important enough to always be available in the logs. It should only happen once anyway so it doesn't really spam.

osherdp: You're right, it seems like debug is turned off
Nice catch :)

src/inventory/disks.go
omertuc: Haven't messed around with the disks in practice, which partition information do you think is interesting to include in this log? For each partition we have the following fields:
```
type Partition struct {
	Disk       *Disk  `json:"-"`
	Name       string `json:"name"`
	Label      string `json:"label"`
	MountPoint string `json:"mount_point"`
	SizeBytes  uint64 `json:"size_bytes"`
	Type       string `json:"type"`
	IsReadOnly bool   `json:"read_only"`
}
```

osherdp: default format seems fine I think
https://github.com/jaypipes/ghw/blob/master/pkg/block/block.go#L209

Maybe @yevgeny-shnaidman have other ideas

src/inventory/disks.go
omertuc: Doesn't seem so, here's an example from my laptop:
```
+block storage (5 disks, 713GB physical storage)
Disk &block.Disk{Name:"dm-0", SizeBytes:0x3b37d00000, PhysicalBlockSizeBytes:0x200, DriveType:4, IsRemovable:false, StorageController:0, BusPath:"unknown", NUMANodeID:0, Vendor:"unknown", Model:"unknown", SerialNumber:"unknown", WWN:"unknown", Partitions:[]*block.Partition{}}
Disk &block.Disk{Name:"dm-1", SizeBytes:0x117f400000, PhysicalBlockSizeBytes:0x200, DriveType:4, IsRemovable:false, StorageController:0, BusPath:"unknown", NUMANodeID:0, Vendor:"unknown", Model:"unknown", SerialNumber:"unknown", WWN:"unknown", Partitions:[]*block.Partition{}}
Disk &block.Disk{Name:"dm-2", SizeBytes:0x3e8000000, PhysicalBlockSizeBytes:0x200, DriveType:4, IsRemovable:false, StorageController:0, BusPath:"unknown", NUMANodeID:0, Vendor:"unknown", Model:"unknown", SerialNumber:"unknown", WWN:"unknown", Partitions:[]*block.Partition{}}
Disk &block.Disk{Name:"dm-3", SizeBytes:0x25d0800000, PhysicalBlockSizeBytes:0x200, DriveType:4, IsRemovable:false, StorageController:0, BusPath:"unknown", NUMANodeID:0, Vendor:"unknown", Model:"unknown", SerialNumber:"unknown", WWN:"unknown", Partitions:[]*block.Partition{}}
Disk &block.Disk{Name:"nvme0n1", SizeBytes:0x3b9e656000, PhysicalBlockSizeBytes:0x200, DriveType:4, IsRemovable:false, StorageController:3, BusPath:"pci-0000:2e:00.0-nvme-1", NUMANodeID:0, Vendor:"unknown", Model:"SAMSUNG MZVLB256HBHQ-000L7", SerialNumber:"S4ELNX0N663432", WWN:"eui.0025388601b396fe", Partitions:[]*block.Partition{(*block.Partition)(0xc00015ab60), (*block.Partition)(0xc00015abd0), (*block.Partition)(0xc00015ac40)}}
	Partition  nvme0n1p1 (600MB) [vfat] mounted@/boot/efi
	Partition  nvme0n1p2 (1GB) [ext4] mounted@/boot
	Partition  nvme0n1p3 (237GB) 

```

src/inventory/disks.go
omertuc: Will use default format then

src/inventory/disks.go
osherdp: What your thought of this matter?
The code is currently not readable

omertuc: I've modified the formatting to make this more readable for now. If you still think it's not readable I'll change it to `if`'s

omertuc:
> also add some tests

I think adding tests to make sure the correct `info` log is being logged is a bit excessive. There are already existing tests to make sure disks are skipped when they need to be skipped, and these tests are passing.

src/inventory/disks.go
osherdp: I think it's still not readable

PR #99, opened 2020-11-24
MGMT-2777 Adapt IPv6 network prefix by using RA routes

ori-amizur:
IPV6 addresses many times don't have the correct network prefix associated.  Instead,
    a prefix values of 128 is attached.  Since we need this prefix to calculate host-networks
    and Machine Network CIDR, this value needs to be fixed.
    The change takes the correct network prefix from Route Advertisement records
    and fixes them in the IPv6 addresses that are returned as part of the inventory.


empovit:
Could you please briefly describe the logic of these changes in the commit message? They very hard to understand without special knowledge.

PR #102, opened 2020-11-24
MGMT-3033 - Use newer smartmontools from the Fedora 32 repositories in upstream Dockerfile

omertuc:
Since centos:8 is using an older version of `smartmontools` (`smartctl` needed for the inventory command), we download the fedora 32 RPM package for `smartmontools` and install it instead of using the centos repos. (We need version 7.1+ for the `--json=c` flag to work).

Note that doing this is NOT needed in the downstream Dockerfile, the base image used there is `openshift/ose-base:ubi8` and it has the 7.1 version of `smartmontools` in the repos

### DIFF
--- a/Dockerfile.assisted_installer_agent
+++ b/Dockerfile.assisted_installer_agent
@@ -3,7 +3,11 @@ RUN dnf install -y \
 		findutils iputils \
 		podman \
 		# inventory
-		smartmontools dmidecode ipmitool biosdevname file \
+		dmidecode ipmitool biosdevname file \
+        # Since centos:8 is using an older version of smartmontools (smartctl needed for the inventory command), we download the fedora 32 RPM package and
+        # install it instead of using the centos repos. (We need version 7.1+ for the `--json=c` flag to work)
+        # Note that doing this is NOT needed in the downstream Dockerfile, the base image used there is openshift/ose-base:ubi8 and it has the 7.1 version of smartmontools in the repos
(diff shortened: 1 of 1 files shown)


omertuc:

As the comment explains, there's no need to do this in downstream

Dockerfile.assisted_installer_agent
```diff
@@ -16,3 +16,11 @@ RUN dnf install -y \
 ADD build/agent build/connectivity_check build/free_addresses build/inventory \
 	build/logs_sender build/dhcp_lease_allocate build/apivip_check build/next_step_runner build/ntp_synchronizer \
 	/usr/bin/
+
+# Since centos:8 is using an older version of smartmontools (smartctl needed for the inventory command),
+# we download the fedora 32 RPM package and install it instead of using the centos repos.
+# (We need version 7.1+ for the `--json=c` flag to work)
+# Note that doing this is NOT needed in the downstream Dockerfile, the base image used
+# there is openshift/ose-base:ubi8 and it has the 7.1 version of smartmontools in the repos
+ADD https://download-cc-rdu01.fedoraproject.org/pub/fedora/linux/releases/32/Everything/x86_64/os/Packages/s/smartmontools-7.1-8.fc32.x86_64.rpm smartmontools-7.1.rpm
```
tsorya, line 25: can we delete it at the end? Maybe it can be oneliner

omertuc:

As the comment explains, there's no need to do this in downstream

omertuc:

As the comment explains, there's no need to do this in downstream

PR #106, opened 2020-11-30
MGMT-3120 IPv6 connectivity check

empovit:
Use ndisc6 to check L2 connectivity between nodes and receive a remote's MAC address.
For now there is a limitation that the outgoing IP address cannot be detected.

### DIFF
--- a/src/commands/connectivity_check.go
+++ b/src/commands/connectivity_check.go
@@ -7,6 +7,7 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/openshift/assisted-installer-agent/src/util"
 	"github.com/openshift/assisted-service/models"
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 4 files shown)


src/commands/connectivity_check.go
```diff
@@ -118,6 +118,15 @@ func macInDstMacs(mac string, allDstMacs []string) bool {
 
 func l2CheckAddressOnNic(dstAddr string, dstMac string, allDstMacs []string, srcNic string, l2chan chan Any) {
 	defer sendDone(l2chan)
+	if net.ParseIP(dstAddr).To4() != nil {
```
ori-amizur, line 121: To4 is not necessarily a good choice because it tries to convert to IPv4.  Instead we should check if it is IPv4 address.  Maybe this is a better test https://github.com/openshift/assisted-service/blob/b49914fce0793cc3ebca2fb182a25212c3ec8950/internal/network/family_utils.go#L8

empovit: I added a util function like the one you showed

ori-amizur (approved):
/lgtm

PR #107, opened 2020-12-01
MGMT-2858 LogsSender: added coredump and dmesg

danielerez:
Added system-coredump file and dmesg output to logs collector
(should be added to each host logs tar in case of installation error).

### DIFF
--- a/src/logs_sender/send_logs.go
+++ b/src/logs_sender/send_logs.go
@@ -34,6 +34,7 @@ type LogsSender interface {
 	FileUploader(filePath string, clusterID strfmt.UUID, hostID strfmt.UUID,
 		inventoryUrl string, pullSecretToken string, agentVersion string) error
 	GatherInstallerLogs(targetDir string) error
+	GatherErrorLogs(targetDir string) error
 }
 
 type LogsSenderExecuter struct{}
@@ -105,6 +106,58 @@ func (e *LogsSenderExecuter) GatherInstallerLogs(targetDir string) error {
 	return nil
 }
 
+func (e *LogsSenderExecuter) GatherErrorLogs(targetDir string) error {
+	// Write the entire output of dmesg
+	outputFile := path.Join(targetDir, "dmesg.logs")
+	err := getDmesgLogs(e, outputFile)
+	if err != nil {
+		return err
+	}
+
+	// Write coredump files
+	err = getCoreDumps(e, targetDir)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func getDmesgLogs(l LogsSender, outputFilePath string) error {
+	log.Infof("Running dmesg")
+	stderr, exitCode := l.ExecuteOutputToFile(outputFilePath, "dmesg")
+	if exitCode != 0 {
+		err := errors.Errorf(stderr)
+		log.WithError(err).Errorf("Failed to run dmesg command")
+		return err
+	}
+	return nil
+}
+
+func getCoreDumps(l LogsSender, targetDir string) error {
+	log.Infof("Get coredump files")
+	coredumpPath := "/var/lib/systemd/coredump/"
+	stdout, stderr, exitCode := l.Execute("ls", coredumpPath)
+	if exitCode != 0 {
+		err := errors.Errorf(stderr)
+		log.WithError(err).Errorf("Failed to list coredump files")
+		return err
+	}
+
+	files := strings.Split(strings.TrimSuffix(stdout, "\n"), "\n")
+	for _, file := range files {
+		outputFile := path.Join(targetDir, fmt.Sprintf("coredump_%s", file))
+		stderr, exitCode := l.ExecuteOutputToFile(outputFile, "cat", coredumpPath + file)
+		if exitCode != 0 {
+			err := errors.Errorf(stderr)
+			log.WithError(err).Errorf("Failed to read coredump file: %s", file)
+			return err
+		}
+	}
+
+	return nil
+}
+
 func getJournalLogsWithFilter(l LogsSender, since string, outputFilePath string, journalFilterParams []string) error {
 	log.Infof("Running journalctl with filters %s", journalFilterParams)
 	args := []string{"-D", "/var/log/journal/", "--since", since, "--all"}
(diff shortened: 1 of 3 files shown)


danielerez:
@YuviGold @tsorya @yevgeny-shnaidman @osherdp @ybettan PTAL

src/logs_sender/send_logs.go
```diff
@@ -118,6 +118,47 @@ func getJournalLogsWithFilter(l LogsSender, since string, outputFilePath string,
...
+	return nil
+}
+
+func getCoreDump(l LogsSender, outputFilePath string) error {
+	log.Infof("Copy coredump file")
+	cmd := "cat $(sysctl kernel.core_pattern | cut -d '|' -f2 | cut -d ' ' -f1)"
+	stderr, exitCode := l.ExecuteOutputToFile(outputFilePath, "bash", "-c", cmd)
+	if exitCode != 0 {
+		err := errors.Errorf(stderr)
+		log.WithError(err).Errorf("Failed to copy coredump file")
```
YuviGold, line 138: Failed to get core dump file. This isn't really copying anything.
Does this file always exist?

src/logs_sender/send_logs.go
```diff
@@ -118,6 +118,47 @@ func getJournalLogsWithFilter(l LogsSender, since string, outputFilePath string,
...
+		log.WithError(err).Errorf("Failed to run dmesg command")
+		return err
+	}
+	return nil
+}
+
+func getCoreDump(l LogsSender, outputFilePath string) error {
+	log.Infof("Copy coredump file")
+	cmd := "cat $(sysctl kernel.core_pattern | cut -d '|' -f2 | cut -d ' ' -f1)"
+	stderr, exitCode := l.ExecuteOutputToFile(outputFilePath, "bash", "-c", cmd)
```
ybettan, line 135: Is this working well? I know for sure that not separating the args in `ExecutePrivilegedCommand` isn't working well (maybe because on `nsester`)

src/logs_sender/send_logs.go
danielerez: ExecuteOutputToFile actually copies the output of the command into the specified file, but I can rephrase.
The coredump file should always exist in the host, but I can ignore the errors just to be on the safe side.

src/logs_sender/send_logs.go
```diff
@@ -118,6 +118,47 @@ func getJournalLogsWithFilter(l LogsSender, since string, outputFilePath string,
...
+	if exitCode != 0 {
+		err := errors.Errorf(stderr)
+		log.WithError(err).Errorf("Failed to copy coredump file")
+		return err
+	}
+	return nil
+}
+
+func gatherErrorLogs(l LogsSender, logsTmpFilesDir string) error {
```
ybettan, line 144: If you make this function and interface method you will be able to mock it entirely instead of mocking all it internal separately

src/logs_sender/send_logs_test.go
```diff
@@ -93,7 +102,7 @@ var _ = Describe("logs sender", func() {
 
 	It("ExecuteOutputToFile failed", func() {
 		folderSuccess()
-		gatherInstallerLogsSuccess()
+		config.LogsSenderConfig.InstallerGatherlogging = false
```
ybettan, line 105: Maybe you can add a different `It` for testing `GatherErrorLogs` once you make it part of the interfce, same way as we do for `GatherInstallerLogs`

src/logs_sender/send_logs.go
danielerez: Yes seems to work well, so should have enough permissions to run this bash command.

src/logs_sender/send_logs.go
```diff
@@ -118,6 +118,47 @@ func getJournalLogsWithFilter(l LogsSender, since string, outputFilePath string,
...
+		err := errors.Errorf(stderr)
+		log.WithError(err).Errorf("Failed to run dmesg command")
+		return err
+	}
+	return nil
+}
+
+func getCoreDump(l LogsSender, outputFilePath string) error {
+	log.Infof("Copy coredump file")
+	cmd := "cat $(sysctl kernel.core_pattern | cut -d '|' -f2 | cut -d ' ' -f1)"
```
osherdp, line 134: do we have ``coredumpctl`` on our system?
might save a bit of the bash magic here

src/logs_sender/send_logs.go
danielerez: Indeed, just wasn't sure it worth exposing as interface just for tests.

src/logs_sender/send_logs.go
```diff
@@ -118,6 +118,47 @@ func getJournalLogsWithFilter(l LogsSender, since string, outputFilePath string,
...
+	if exitCode != 0 {
+		err := errors.Errorf(stderr)
+		log.WithError(err).Errorf("Failed to run dmesg command")
+		return err
+	}
+	return nil
+}
+
+func getCoreDump(l LogsSender, outputFilePath string) error {
+	log.Infof("Copy coredump file")
```
osherdp, line 133: It might be harder, but...
can we store all the core dumps?

src/logs_sender/send_logs.go
```diff
@@ -118,6 +118,47 @@ func getJournalLogsWithFilter(l LogsSender, since string, outputFilePath string,
...
+		return err
+	}
+
+	// Write system coredump file
+	outputFile = path.Join(logsTmpFilesDir, "system-coredump")
+	err = getCoreDump(l, outputFile)
+	if err != nil {
+		return err
+	}
+
```
osherdp, line 158: can we get ``/var/log/messages`` as well?
I think it's different from ``dmesg`` and can be interesting as well

osherdp (changes requested):

src/logs_sender/send_logs_test.go
danielerez: Right, but not sure we have much to test here though. I'll give it a try.

src/logs_sender/send_logs.go
danielerez: You mean besides the system coredump? Which ones would be interesting?

src/logs_sender/send_logs.go
danielerez: Yeah actually tried with coredumpctl, but seems 'No coredumps found.' on the machine,
so I've just included the system-coredump.

src/logs_sender/send_logs.go
danielerez: yeah looked for it,  but seems that rsyslog isn't running in the hosts.

src/logs_sender/send_logs.go
osherdp: What happens when two segfaults happen? Will the second core dump override the first one?
I don't think so, so we should bring them both

Also, does ``coreos-installer`` will dump in the system's location or any other location?
It's very important because we have unexplained behavior when using ``coreos-installer`` that can be investigated only by core dumps

src/logs_sender/send_logs.go
osherdp: k

src/logs_sender/send_logs.go
osherdp: ok
there's journalctl BTW?

src/logs_sender/send_logs.go
danielerez: sure, I'll just include all coredump files.
iinm, core-installer should dump in same locationn. @YuviGold

YuviGold: Unfortunately, I have no idea where `coreos-installer` dumps to.

romfreiman: https://coreos.slack.com/archives/CUPJTHQ5P/p1605165310224300?thread_ts=1605057411.176200&cid=CUPJTHQ5P

danielerez: ok, so seems it's indeed the same location.

YuviGold (approved):
LGTM

PR #108, opened 2020-12-02
MGMT-2858: LogSender - use coredumpctl

danielerez:
Used coredumpctl to fetch all coresumps.

### DIFF
--- a/src/logs_sender/send_logs.go
+++ b/src/logs_sender/send_logs.go
@@ -136,21 +136,22 @@ func getDmesgLogs(l LogsSender, outputFilePath string) error {
 
 func getCoreDumps(l LogsSender, targetDir string) error {
 	log.Infof("Get coredump files")
-	coredumpPath := "/var/lib/systemd/coredump/"
-	stdout, stderr, exitCode := l.Execute("ls", coredumpPath)
+	stdout, stderr, exitCode := l.Execute("coredumpctl", "list", "--no-legend")
 	if exitCode != 0 {
-		err := errors.Errorf(stderr)
-		log.WithError(err).Errorf("Failed to list coredump files")
-		return err
+		log.Infof("Couldn't fetch coredump list: %s", stderr)
+		return nil
 	}
 
-	files := strings.Split(strings.TrimSuffix(stdout, "\n"), "\n")
-	for _, file := range files {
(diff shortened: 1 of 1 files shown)


src/logs_sender/send_logs.go
```diff
@@ -136,21 +136,20 @@ func getDmesgLogs(l LogsSender, outputFilePath string) error {
 
 func getCoreDumps(l LogsSender, targetDir string) error {
 	log.Infof("Get coredump files")
-	coredumpPath := "/var/lib/systemd/coredump/"
-	stdout, stderr, exitCode := l.Execute("ls", coredumpPath)
+	stdout, _, exitCode := l.Execute("coredumpctl", "list", "--no-legend")
 	if exitCode != 0 {
```
osherdp, line 140: It doesn't return specific ``exitCode`` when there are not matches?
We won't return "No coredump files available" on legit errors?

src/logs_sender/send_logs.go
```diff
@@ -136,21 +136,20 @@ func getDmesgLogs(l LogsSender, outputFilePath string) error {
...
 	}
 
-	files := strings.Split(strings.TrimSuffix(stdout, "\n"), "\n")
-	for _, file := range files {
-		outputFile := path.Join(targetDir, fmt.Sprintf("coredump_%s", file))
-		stderr, exitCode := l.ExecuteOutputToFile(outputFile, "cat", coredumpPath + file)
+	dumps := strings.Split(strings.TrimSuffix(stdout, "\n"), "\n")
+	for _, dump := range dumps {
+		pid := strings.Fields(dump)[4]
+		outputFile := path.Join(targetDir, fmt.Sprintf("coredump_pid_%s", pid))
```
osherdp, line 148: is it possible to embed the executable basename path as well in the target path?

osherdp (changes requested):

src/logs_sender/send_logs.go
danielerez: It returns a generic error(1) so I'll just add stderr to the log.

src/logs_sender/send_logs.go
danielerez: sure, added to the filename

PR #109, opened 2020-12-06
Use quay.io/app-sre/golang for build container

YuviGold:
To avoid docker hub limitations


YuviGold:
Images weren't changed, just their base image to be from quay.io registry

eranco74 (approved):
/lgtm

PR #111, opened 2020-12-09
MGMT-3012 - Add mount and volume display logs

slaviered:


src/logs_sender/send_logs.go
```diff
@@ -226,24 +265,31 @@ func SendLogs(l LogsSender) error {
...
 	}
 
 	for _, service := range config.LogsSenderConfig.Services {
 		outputFile := path.Join(logsTmpFilesDir, fmt.Sprintf("%s.logs", service))
-		err := getJournalLogsWithFilter(l, config.LogsSenderConfig.Since, outputFile,
-			[]string{"-u", service})
-		if err != nil {
-			return err
-		}
+		if err := getJournalLogsWithFilter(l, config.LogsSenderConfig.Since, outputFile,
```
yevgeny-shnaidman, line 283: i think that it would be better to get the whole jouralctl and not just journalctl of a specific services. In case we have filesystem/disks/HW errors you want to look at the entires journalctl to understand what happened. @tsorya ?

slaviered: Anyhow, This should be done in a separate PR. Here I only aggregate errors, not changing existing logic. As for the suggestion: It's difficult to read the entire journal with no filtering by sources.

PR #113, opened 2020-12-13
MGMT-3257: discarding disks of type ODD (Optical Disk Drive)

osherdp:
/hold


src/inventory/disks_test.go
```diff
@@ -169,6 +170,22 @@ var _ = Describe("Disks test", func() {
 						},
 					},
 				},
+				{
```
omertuc, line 173: This test is for checking ISO filtering, I think it's better if you create a separate test / change the name of the test to make it clear you're testing something else

PR #116, opened 2020-12-14
MGMT-2858: coredumpctl - use priviliged

danielerez:
Use ExecutePrivileged (i.e. nsenter1) for coredumpctl execution.
This is needed for getting the core dumps in host context,
otherwise, coredumpctl fails to get sd-bus ("Failed to acquire bus")

### DIFF
--- a/src/logs_sender/send_logs.go
+++ b/src/logs_sender/send_logs.go
@@ -173,7 +173,7 @@ func getDmesgLogs(l LogsSender, outputFilePath string) error {
 
 func getCoreDumps(l LogsSender, targetDir string) error {
 	log.Infof("Get coredump files")
-	stdout, stderr, exitCode := l.Execute("coredumpctl", "list", "--no-legend")
+	stdout, stderr, exitCode := l.ExecutePrivileged("coredumpctl", "list", "--no-legend")
 	if exitCode != 0 {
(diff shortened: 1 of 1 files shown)


osherdp:
And coreos-installer dumps can be found via coredumpctl? Even though it's not being invoked by systemd?

danielerez:
> And coreos-installer dumps can be found via coredumpctl? Even though it's not being invoked by systemd?

Tried it locally and seems the dump file is indeed generated:
$ coreos-installer download
gpg: Signature made Tue Nov 17 12:51:30 2020 UTC
gpg:                using RSA key 97A1AE57C3A2372CCA3A4ABA6C13026D12C944D0
Quit (core dumped)
[core@test-infra-cluster-assisted-installer-master-0 ~]$ gpg: BAD signature from "Fedora (32) <fedora-32-primary@fedoraproject.org>" [ultimate]
$ killall -3 coreos-installer
$ ll /var/lib/systemd/coredump/
-rw-r-----+ 1 root root 2179996 Dec 14 17:24 core.coreos-installe.1000.d65d2602ae21422b9a20d2d20b68d4bf.314986.1607966658000000.lz4

But, still not sure whether any seg fault will invoke a dump. As we don't see in coreos-installer logs "Segmentation fault _(core dumped)_" as expected. So will see in the next log...

PR #117, opened 2020-12-15
OCPBUGSM-21434 NTP: Ignore adding server failures

YuviGold:
In case could not determine whether a new server is already configured -
just log.
In case adding a new server failed - just log as well.
We always want to receive the current NTP sources from the agent.


ori-amizur (approved):
/lgtm

PR #119, opened 2020-12-15
MGMT-3026: skip slow disks using fio output

danielerez:
- According to the following blog[*], the 99th percentile of fdatasync durations should be less than 10ms for running etcd on the cluster.
- Hence, using fio tool output, skip disks that are considered too slow according to the aforementioned metric.
- Added fio_perf_check command for invoking fio perf test.

[*] https://www.ibm.com/cloud/blog/using-fio-to-tell-whether-your-storage-is-fast-enough-for-etcd


danielerez:
@filanov @omertuc PTAL

src/inventory/disks.go
```diff
@@ -84,6 +87,41 @@ func (d *disks) getSMART(path string) string {
...
 
+// Returns the 99th percentile of fdatasync durations in milliseconds
+func (d *disks) getDiskPerf(disk *ghw.Disk, path string) (int64, error) {
+	if path == "" {
+		return -1, errors.New("Missing disk path")
+	}
+
+	args := []string{"--filename", path, "--name=test", "--rw=write", "--ioengine=sync",
+		"--size=22m", "-bs=2300", "--fdatasync=1", "--output-format=json"}
+	stdout, stderr, exitCode := d.dependencies.Execute("fio", args...)
```
omertuc, line 98: How long does that take?

src/inventory/disks.go
```diff
@@ -123,9 +161,23 @@ func diskIsRelevant(disk *ghw.Disk) (relevant bool, skipReasons []string) {
...
 
-	return
+	if (isRelevant()) {
+		diskPerf, err := d.getDiskPerf(disk, path)
+		if err != nil {
+			logrus.WithError(err).Warnf("Failed to get disk's I/O performance: %s", disk.Name)
+		} else if diskPerf > 10 {
+			// If the 99th percentile of fdatasync durations is more than 10ms, it's not fast enough for etcd.
+			// See: https://www.ibm.com/cloud/blog/using-fio-to-tell-whether-your-storage-is-fast-enough-for-etcd
+			skipReasons = append(skipReasons, "Disk is not fast enough for installation")
```
omertuc, line 175: %d diskPerf?

omertuc (commented):
Looks good otherwise

omertuc:
> According to the following blog[*], the 99th percentile of
> fdatasync durations should be less than 10ms for running etcd
> on the cluster.
> Hence, using fio tool output, skip disks that are considered
> too slow according to the aforementioned metric.
> 
> [*] https://www.ibm.com/cloud/blog/using-fio-to-tell-whether-your-storage-is-fast-enough-for-etcd

Do we care about etcd in worker hosts? i.e is that limitation fair to them?

src/inventory/disks.go
```diff
@@ -137,7 +189,8 @@ func (d *disks) getDisks() []*models.Disk {
 	}
 
 	for _, disk := range blockInfo.Disks {
-		isRelevant, reasons := diskIsRelevant(disk)
+		path := d.getPath(disk.BusPath, disk.Name)
+		isRelevant, reasons := d.diskIsRelevant(disk, path)
```
filanov, line 193: maybe it should be done in parallel?

src/inventory/disks.go
```diff
@@ -94,7 +132,7 @@ func unknownToEmpty(value string) string {
 // diskIsRelevant checks if a disk is relevant for installation by testing
 // it against a list of predicates. Also returns all the reasons the disk
 // was found to be irrelevant
-func diskIsRelevant(disk *ghw.Disk) (relevant bool, skipReasons []string) {
+func (d *disks) diskIsRelevant(disk *ghw.Disk, path string) (relevant bool, skipReasons []string) {
```
romfreiman, line 135: what will happen if we'll not find any disk? what the UX for that?

filanov: it should be a as any other validation failure.

romfreiman: so what will the user do? I mean speed is negotiable - depending on the cluster usage, right?

filanov: Not at this stage, we don't support cluster profile yet. so it should fail, the same as if the user don't have enough disk space, RAM or CPU.

src/inventory/disks.go
danielerez: Depends on the hw, but seems ~10sec:
$time fio --filename=/dev/sda --name=perftest --rw=write --ioengine=sync --size=22m --bs=2300 --fdatasync=1 --output-format=json
...
real	0m8.397s
user	0m0.218s
sys	0m0.536s

danielerez:
> > According to the following blog[*], the 99th percentile of
> > fdatasync durations should be less than 10ms for running etcd
> > on the cluster.
> > Hence, using fio tool output, skip disks that are considered
> > too slow according to the aforementioned metric.
> > [*] https://www.ibm.com/cloud/blog/using-fio-to-tell-whether-your-storage-is-fast-enough-for-etcd
> 
> Do we care about etcd in worker (non-master) hosts? i.e. is that limitation fair to them?

I guess it's worth checking performance anyway, as the i/o check seems sensible for modern disks.
@filanov @ronniel1 what do you think? should we skip the validation for workers?

src/inventory/disks.go
omertuc: The inventory command typically runs every 60 seconds. That means a host with >7~8 disks, if I understand it correctly, will start to slowly create a backlog of inventory command containers.

Probably not a typical occurrence, but maybe we should take that into account. Maybe run the check for all disks in parallel like @filanov suggested

src/inventory/disks.go
romfreiman: so at least we leave some spares on the disk performance? Or we're strict?
What about the new api that lets the user to choose a disk?

omertuc: Strict. The new API will not allow the user to choose a disk that has "skip reasons".

romfreiman: Won't we run into some issues there that users will not be manage to install?
What will be blocked for example?
cc: @avishayt @ronniel1

src/inventory/disks.go
omertuc: Might be best to run it in parallel like @filanov suggested then, to decrease the time it takes to return an inventory from ~n*8.3 to ~8.3

src/inventory/disks.go
danielerez: Right, for now it's validating for <10ms (as suggested in https://www.ibm.com/cloud/blog/using-fio-to-tell-whether-your-storage-is-fast-enough-for-etcd). We can increase it a bit if makes sense.
We simply skip the disk if doesn't meet the requirements. Do you think we should just let the user select instead?
@ronniel1 @filanov @avishayt

filanov: What is the point having this validation if we allow to install anyway?

filanov:
> > > According to the following blog[_], the 99th percentile of
> > > fdatasync durations should be less than 10ms for running etcd
> > > on the cluster.
> > > Hence, using fio tool output, skip disks that are considered
> > > too slow according to the aforementioned metric.
> > > [_] https://www.ibm.com/cloud/blog/using-fio-to-tell-whether-your-storage-is-fast-enough-for-etcd
> > 
> > 
> > Do we care about etcd in worker (non-master) hosts? i.e. is that limitation fair to them?
> 
> I guess it's worth checking performance anyway, as the i/o check seems sensible for modern disks.
> @filanov @ronniel1 what do you think? should we skip the validation for workers?

Maybe instead of having the decision here it should be done in the service? this way if we have workers/masters etc then it will be done in the service.
In addition you will have an issue with auto assign, if role is not set you will probably need to address it as a master.

src/inventory/disks.go
omertuc: Agree with @filanov. Either we add a separate mechanism for "disk warnings" if we want more flexibility, or we allow the user to install on any disk they wish, even if we decided it's uninstallable.

omertuc:
> > > > According to the following blog[_], the 99th percentile of
> > > > fdatasync durations should be less than 10ms for running etcd
> > > > on the cluster.
> > > > Hence, using fio tool output, skip disks that are considered
> > > > too slow according to the aforementioned metric.
> > > > [_] [ibm.com/cloud/blog/using-fio-to-tell-whether-your-storage-is-fast-enough-for-etcd](https://www.ibm.com/cloud/blog/using-fio-to-tell-whether-your-storage-is-fast-enough-for-etcd)
> > > 
> > > 
> > > Do we care about etcd in worker (non-master) hosts? i.e. is that limitation fair to them?
> > 
> > 
> > I guess it's worth checking performance anyway, as the i/o check seems sensible for modern disks.
> > @filanov @ronniel1 what do you think? should we skip the validation for workers?
> 
> Maybe instead of having the decision here it should be done in the service? this way if we have workers/masters etc then it will be done in the service.
> In addition you will have an issue with auto assign, if role is not set you will probably need to address it as a master.

Making this restriction on masters-only might complicate things, even if we relay this performance stat to the service. Right now, the service decides the eligibility of each disk based on the information present in the inventory alone. Adding roles into the mix means you need to update the inventory every time the user decides to change roles.

src/inventory/disks.go
romfreiman: I think it can be layered.
Some numbers - yes, dont allow to install, it just will not work
Others. we warn. but the user still can install. 
I just afraid that we'll find ourselves not letting anyone to install. Did we check metrics/elastic for disks types? Will they fit into those numbers?

danielerez:
> > > > > According to the following blog[_], the 99th percentile of
> > > > > fdatasync durations should be less than 10ms for running etcd
> > > > > on the cluster.
> > > > > Hence, using fio tool output, skip disks that are considered
> > > > > too slow according to the aforementioned metric.
> > > > > [_] [ibm.com/cloud/blog/using-fio-to-tell-whether-your-storage-is-fast-enough-for-etcd](https://www.ibm.com/cloud/blog/using-fio-to-tell-whether-your-storage-is-fast-enough-for-etcd)
> > > > 
> > > > 
> > > > Do we care about etcd in worker (non-master) hosts? i.e. is that limitation fair to them?
> > > 
> > > 
> > > I guess it's worth checking performance anyway, as the i/o check seems sensible for modern disks.
> > > @filanov @ronniel1 what do you think? should we skip the validation for workers?
> > 
> > 
> > Maybe instead of having the decision here it should be done in the service? this way if we have workers/masters etc then it will be done in the service.
> > In addition you will have an issue with auto assign, if role is not set you will probably need to address it as a master.
> 
> Warning in advance - making this restriction masters-only might complicate the "disk eligibility" flow, even if we relay this performance stat to the service and let the service decide, as you said.
> 
> This is because right now, the service decides the eligibility of each disk for installation based on information present in the inventory alone, and it updates the inventory accordingly. Adding host roles into the mix means you need to update the inventory every time the user decides to change host roles.

so maybe worth keeping the validation for all disks for now?
or can we do the validation just before installing? e.g. in https://github.com/openshift/assisted-service/blob/1856df753deaf1bb5de79a854b0fdc485a0b093d/internal/bminventory/inventory.go#L3105

omertuc:
> > > > > > According to the following blog[_], the 99th percentile of
> > > > > > fdatasync durations should be less than 10ms for running etcd
> > > > > > on the cluster.
> > > > > > Hence, using fio tool output, skip disks that are considered
> > > > > > too slow according to the aforementioned metric.
> > > > > > [_] [ibm.com/cloud/blog/using-fio-to-tell-whether-your-storage-is-fast-enough-for-etcd](https://www.ibm.com/cloud/blog/using-fio-to-tell-whether-your-storage-is-fast-enough-for-etcd)
> > > > > 
> > > > > 
> > > > > Do we care about etcd in worker (non-master) hosts? i.e. is that limitation fair to them?
> > > > 
> > > > 
> > > > I guess it's worth checking performance anyway, as the i/o check seems sensible for modern disks.
> > > > @filanov @ronniel1 what do you think? should we skip the validation for workers?
> > > 
> > > 
> > > Maybe instead of having the decision here it should be done in the service? this way if we have workers/masters etc then it will be done in the service.
> > > In addition you will have an issue with auto assign, if role is not set you will probably need to address it as a master.
> > 
> > 
> > Warning in advance - making this restriction masters-only might complicate the "disk eligibility" flow, even if we relay this performance stat to the service and let the service decide, as you said.
> > This is because right now, the service decides the eligibility of each disk for installation based on information present in the inventory alone, and it updates the inventory accordingly. Adding host roles into the mix means you need to update the inventory every time the user decides to change host roles.
> 
> so maybe worth keeping the validation for all disks for now?
> or can we do the validation just before installing? e.g. in [openshift/assisted-service@`1856df7`/internal/bminventory/inventory.go#L3105](https://github.com/openshift/assisted-service/blob/1856df753deaf1bb5de79a854b0fdc485a0b093d/internal/bminventory/inventory.go#L3105)

We must perform the validation in advance. The user must be aware of all disks and whether they're fit for installation even before they press install.

src/inventory/disks.go
danielerez: Not really, just followed the suggested threshold for etcd, but it seems reasonable for modern disks.
Should we simply increase it for now, and maybe add some warnings for borderline values?

filanov:
> > > > > > > According to the following blog[_], the 99th percentile of
> > > > > > > fdatasync durations should be less than 10ms for running etcd
> > > > > > > on the cluster.
> > > > > > > Hence, using fio tool output, skip disks that are considered
> > > > > > > too slow according to the aforementioned metric.
> > > > > > > [_] [ibm.com/cloud/blog/using-fio-to-tell-whether-your-storage-is-fast-enough-for-etcd](https://www.ibm.com/cloud/blog/using-fio-to-tell-whether-your-storage-is-fast-enough-for-etcd)
> > > > > > 
> > > > > > 
> > > > > > Do we care about etcd in worker (non-master) hosts? i.e. is that limitation fair to them?
> > > > > 
> > > > > 
> > > > > I guess it's worth checking performance anyway, as the i/o check seems sensible for modern disks.
> > > > > @filanov @ronniel1 what do you think? should we skip the validation for workers?
> > > > 
> > > > 
> > > > Maybe instead of having the decision here it should be done in the service? this way if we have workers/masters etc then it will be done in the service.
> > > > In addition you will have an issue with auto assign, if role is not set you will probably need to address it as a master.
> > > 
> > > 
> > > Warning in advance - making this restriction masters-only might complicate the "disk eligibility" flow, even if we relay this performance stat to the service and let the service decide, as you said.
> > > This is because right now, the service decides the eligibility of each disk for installation based on information present in the inventory alone, and it updates the inventory accordingly. Adding host roles into the mix means you need to update the inventory every time the user decides to change host roles.
> > 
> > 
> > so maybe worth keeping the validation for all disks for now?
> > or can we do the validation just before installing? e.g. in [openshift/assisted-service@`1856df7`/internal/bminventory/inventory.go#L3105](https://github.com/openshift/assisted-service/blob/1856df753deaf1bb5de79a854b0fdc485a0b093d/internal/bminventory/inventory.go#L3105)
> 
> We must perform the validation in advance. The user must be aware of all disks and whether they're fit for installation even before they press install.

We can't start writing to users disks before we know that we can start the installation. until the user ask to perform the installation all the system is as a draft.

omertuc:
> > > > > > > > According to the following blog[_], the 99th percentile of
> > > > > > > > fdatasync durations should be less than 10ms for running etcd
> > > > > > > > on the cluster.
> > > > > > > > Hence, using fio tool output, skip disks that are considered
> > > > > > > > too slow according to the aforementioned metric.
> > > > > > > > [_] [ibm.com/cloud/blog/using-fio-to-tell-whether-your-storage-is-fast-enough-for-etcd](https://www.ibm.com/cloud/blog/using-fio-to-tell-whether-your-storage-is-fast-enough-for-etcd)
> > > > > > > 
> > > > > > > 
> > > > > > > Do we care about etcd in worker (non-master) hosts? i.e. is that limitation fair to them?
> > > > > > 
> > > > > > 
> > > > > > I guess it's worth checking performance anyway, as the i/o check seems sensible for modern disks.
> > > > > > @filanov @ronniel1 what do you think? should we skip the validation for workers?
> > > > > 
> > > > > 
> > > > > Maybe instead of having the decision here it should be done in the service? this way if we have workers/masters etc then it will be done in the service.
> > > > > In addition you will have an issue with auto assign, if role is not set you will probably need to address it as a master.
> > > > 
> > > > 
> > > > Warning in advance - making this restriction masters-only might complicate the "disk eligibility" flow, even if we relay this performance stat to the service and let the service decide, as you said.
> > > > This is because right now, the service decides the eligibility of each disk for installation based on information present in the inventory alone, and it updates the inventory accordingly. Adding host roles into the mix means you need to update the inventory every time the user decides to change host roles.
> > > 
> > > 
> > > so maybe worth keeping the validation for all disks for now?
> > > or can we do the validation just before installing? e.g. in [openshift/assisted-service@`1856df7`/internal/bminventory/inventory.go#L3105](https://github.com/openshift/assisted-service/blob/1856df753deaf1bb5de79a854b0fdc485a0b093d/internal/bminventory/inventory.go#L3105)
> > 
> > 
> > We must perform the validation in advance. The user must be aware of all disks and whether they're fit for installation even before they press install.
> 
> We can't start writing to users disks before we know that we can start the installation. until the user ask to perform the installation all the system is as a draft.

Yeah I realized that during the meeting. Didn't know at first that these tests perform actual writes to the disk.

src/fio_perf_check/fio_perf_check.go
```diff
@@ -0,0 +1,96 @@
...
+		err := errors.New("Missing Filename in FioPerfCheckRequest")
+		log.WithError(err).Error(err.Error())
+		return createResponse(-1), err.Error(), -1
+	}
+
+	diskPerf, err := p.getDiskPerf(*fioPerfCheckRequest.Path)
+	if err != nil {
+		log.WithError(err).Warnf("Failed to get disk's I/O performance: %s", *fioPerfCheckRequest.Path)
+		return createResponse(-1), err.Error(), -1
+	} else if diskPerf > *fioPerfCheckRequest.DurationThreshold {
```
ori-amizur, line 40: Why does agent check if performance is sufficient? I think it should pass the details to the assisted-service, and the assisted-service should check it.

ori-amizur (commented):
I think the threshold should not be passed to the agent.  Instead, the check should be done in the assisted service.  The agent should send the raw results of the check to the assisted service.

src/fio_perf_check/fio_perf_check.go
rollandf: In general, I agree with you,  but currently, this check is run as part of the install phase and not preparing phase.
This is the reason we want this command to check and fail if the threshold is not met.

ori-amizur: OK

gamli75:
@danielerez @rollandf Don't forget to change the Docker file in the gerrit: https://code.engineering.redhat.com/gerrit/gitweb?p=assisted-installer-projects.git;a=blob;f=distgit/containers/assisted-installer-agent/Dockerfile.in;h=b1957a07a671d011d5d006e5a939fe869e2c2f62;hb=refs/heads/rhaos-4.6-rhel-8 
and check that it's working when we use DS image.

PR #122, opened 2020-12-22
MGMT-3291 - Show all disks - mark eligibility instead of filtering

omertuc:
Agent will just mark disks as ineligible instead of filtering them out completely.

### DIFF
--- a/src/inventory/disks.go
+++ b/src/inventory/disks.go
@@ -91,41 +91,34 @@ func unknownToEmpty(value string) string {
 	return value
 }
 
-// diskIsRelevant checks if a disk is relevant for installation by testing
-// it against a list of predicates. Also returns all the reasons the disk
-// was found to be irrelevant
-func diskIsRelevant(disk *ghw.Disk) (relevant bool, skipReasons []string) {
-	if disk.IsRemovable {
-		skipReasons = append(skipReasons, "Disk is removable")
-	}
+// checkEligibility checks if a disk is eligible for installation by testing
+// it against a list of predicates. Returns all the reasons the disk
+// was found to be not eligible, or an empty slice if it was found to
+// be eligible
+func checkEligibility(disk *ghw.Disk) []string {
+	var notEligibleReasons []string
 
-	if disk.SizeBytes == 0 {
-		skipReasons = append(skipReasons, "Disk has size 0")
+	if disk.IsRemovable {
+		notEligibleReasons = append(notEligibleReasons, "Disk is removable")
 	}
 
 	if disk.BusType == ghw.BUS_TYPE_UNKNOWN && disk.StorageController == ghw.STORAGE_CONTROLLER_UNKNOWN {
-		skipReasons = append(skipReasons, "Disk has unknown bus type and storage controller")
-	}
-
-	if disk.DriveType == ghw.DRIVE_TYPE_ODD {
-		skipReasons = append(skipReasons, "Disk is an optical disk drive")
+		notEligibleReasons = append(notEligibleReasons, "Disk has unknown bus type and storage controller")
 	}
 
 	if funk.Contains(funk.Map(disk.Partitions, func(p *ghw.Partition) bool {
 		return p.Type == "iso9660"
 	}), true) {
-		skipReasons = append(skipReasons, "Disk appears to be an ISO installation media (has partition with type iso9660)")
+		notEligibleReasons = append(notEligibleReasons, "Disk appears to be an ISO installation media (has partition with type iso9660)")
 	}
 
 	if funk.Contains(funk.Map(disk.Partitions, func(p *ghw.Partition) bool {
 		return strings.HasSuffix(p.MountPoint, "iso")
 	}), true) {
-		skipReasons = append(skipReasons, "Disk appears to be an ISO installation media (has partition with mountpoint suffix iso)")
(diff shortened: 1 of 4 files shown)


src/inventory/disks.go
```diff
@@ -91,41 +91,34 @@ func unknownToEmpty(value string) string {
...
-		skipReasons = append(skipReasons, "Disk has size 0")
+	if disk.IsRemovable {
+		notEligibleReasons = append(notEligibleReasons, "Disk is removable")
 	}
 
 	if disk.BusType == ghw.BUS_TYPE_UNKNOWN && disk.StorageController == ghw.STORAGE_CONTROLLER_UNKNOWN {
-		skipReasons = append(skipReasons, "Disk has unknown bus type and storage controller")
-	}
-
-	if disk.DriveType == ghw.DRIVE_TYPE_ODD {
```
YuviGold, line 110: Was this check removed on purpose?

src/inventory/disks.go
```diff
@@ -137,29 +130,32 @@ func (d *disks) getDisks() []*models.Disk {
...
+		var eligibility models.DiskInstallationEligibility
+		eligibility.NotEligibleReasons = diskIsEligible(disk)
+		eligibility.Eligible = len(eligibility.NotEligibleReasons) == 0
+
+		if !eligibility.Eligible {
+			reasons := strings.Join(eligibility.NotEligibleReasons, ", ")
 			logrus.Infof(
-				"Disk (name %s drive type %v bus path %s vendor %s model %s partitions %s) was found to be irrelevant for the following reasons: %s",
-				disk.Name, disk.DriveType.String(), disk.BusPath, disk.Vendor, disk.Model, disk.Partitions, reasonsLines)
-			continue
```
omertuc, line 146: Instead of skipping, just log

src/inventory/disks_test.go
```diff
@@ -12,6 +13,95 @@ import (
 	"github.com/openshift/assisted-service/models"
 )
 
+// deleteExpectedCall allows deleting call expectations from a mock according to method name and arguments
+func deleteExpectedCall(mockDeps *MockIDependencies, methodName string, arguments ...interface{}) {
```
omertuc, line 17: Having the ability to remove expected calls from a mock simplifies the tests somewhat

src/inventory/disks_test.go
```diff
@@ -12,6 +13,95 @@ import (
...
+	Expect(deleteIndex).ToNot(Equal(-1))
+
+	// Remove
+	mockDeps.ExpectedCalls = append(
+		mockDeps.ExpectedCalls[:deleteIndex],
+		mockDeps.ExpectedCalls[deleteIndex+1:]...,
+	)
+}
+
+func prepareDiskObjects(dependencies *MockIDependencies, diskNum int) (*ghw.Disk, *models.Disk) {
```
omertuc, line 45: This code was repeated too much in the tests, so I put it here in a more "generic" form that most tests now use.

src/inventory/disks.go
```diff
@@ -91,41 +91,34 @@ func unknownToEmpty(value string) string {
...
-// was found to be irrelevant
-func diskIsRelevant(disk *ghw.Disk) (relevant bool, skipReasons []string) {
-	if disk.IsRemovable {
-		skipReasons = append(skipReasons, "Disk is removable")
-	}
+// DiskIsEligible checks if a disk is eligible for installation by testing
+// it against a list of predicates. Returns all the reasons the disk
+// was found to be not eligible, or an empty slice if it was found to
+// be eligible
+func diskIsEligible(disk *ghw.Disk) []string {
```
YuviGold, line 98: I would expect from `Is` as a function name to return a boolean value.
While this function returns a strings array, what about `getEligibilityReasons`?

src/inventory/disks_test.go
```diff
@@ -12,6 +13,95 @@ import (
...
+
+	return mockDisk, expectedDisk
+}
+
+func prepareOneDiskTest(dependencies *MockIDependencies) (*ghw.BlockInfo, []*models.Disk) {
+	disk1Ghw, disk1Models := prepareDiskObjects(dependencies, 1)
+	return &ghw.BlockInfo{Disks: []*ghw.Disk{disk1Ghw}}, []*models.Disk{disk1Models}
+}
+
+func prepareTwoDisksTest(dependencies *MockIDependencies) (*ghw.BlockInfo, []*models.Disk) {
```
YuviGold, line 99: And when we would like 3 disks test? Please change it to be generic, getting N number of disks.

src/inventory/disks.go
omertuc: Yes. This check and the size 0 check are performed by the service anyway, so there's no need to perform them here as well.

src/inventory/disks.go
omertuc: It used to be `bool, []string` returned here, but I removed the bool cause it was redundant, forgot to change the name.

The service PR has a very similar function with the same name, and I would like to keep them similar because they're related. I do not wish to make any further to the service PR, so I'll fix this in a future, dedicated, separate PR.

omertuc: Fixed

src/inventory/disks_test.go
omertuc: > And when we would like 3 disks test? Please change it to be generic, getting N number of disks.

Fixed

YuviGold (approved):
LGTM

YuviGold (approved):

PR #124, opened 2020-12-24
MGMT-3194 Use nmap for L2 connectivity check

empovit:
Replace ndisc6 because it cannot handle concurrent discovery


ori-amizur (commented):
Seems OK.  Can you add some tests?

ori-amizur (approved):
/lgtm

PR #127, opened 2020-12-29
Print NTP sources on NTP subsystem tests

YuviGold:


subsystem/ntp_test.go
```diff
@@ -140,6 +150,12 @@ var _ = Describe("NTP tests", func() {
 	})
 })
 
+func printNtpSources(ntpResponse *models.NtpSynchronizationResponse) {
+	for _, source := range ntpResponse.NtpSources {
+		fmt.Println(source.SourceName, source.SourceState)
```
ori-amizur, line 155: Maybe formatted message to explain what is printed?

YuviGold:
This isn't necessary anymore since we found the NTP subsystem tests bug on https://github.com/openshift/assisted-installer-agent/pull/128

PR #134, opened 2021-01-03
MGMT-3456 - Agent will now report whether a disk appears to be an installation media (ISO)

omertuc:


src/inventory/disks.go
```diff
@@ -95,8 +95,9 @@ func unknownToEmpty(value string) string {
 // it against a list of predicates. Returns all the reasons the disk
 // was found to be not eligible, or an empty slice if it was found to
 // be eligible
-func checkEligibility(disk *ghw.Disk) []string {
+func checkEligibility(disk *ghw.Disk) ([]string, bool) {
```
ronniel1, line 98: Please name the return values.
This way there is no need to read the entire function code, to understand what it returns
since the return values are counter-intuitive

src/inventory/disks_test.go
```diff
@@ -220,6 +221,7 @@ var _ = Describe("Disks test", func() {
 		expectedDisks[1].InstallationEligibility.NotEligibleReasons = []string{
 			"Disk appears to be an ISO installation media (has partition with type iso9660)",
 		}
+		expectedDisks[1].IsInstallationMedia = true
```
ronniel1, line 224: Also add a check in other cases where installation media is false

omertuc:
/unhold

openshift/assisted-service#862 has been merged

ronniel1 (commented):
/lgtm

YuviGold (approved):

PR #133, opened 2021-01-03
MGMT-3458 Send step reply from FIO perf check

rollandf:

PR #136, opened 2021-01-04
MGMT-3458: added device path to fio response

danielerez:


danielerez:
@rollandf

danielerez:
pushed to https://github.com/openshift/assisted-installer-agent/pull/133

PR #145, opened 2021-01-19
Add generateNsenterStep and generateDockerStep to subsystem utils

YuviGold:


subsystem/utils.go
```diff
@@ -488,3 +494,38 @@ func setReplyStartAgent(hostID string) {
...
+	}
+}
+
+func generateDockerStep(stepType models.StepType, podmanArgs []string, commandArgs []string) *models.Step {
+	podmanOptions := []string{
+		"run", "--privileged", "--rm",
+		"-v", "/var/log:/var/log",
+		"-v", "/run/systemd/journal/socket:/run/systemd/journal/socket",
+	}
+	podmanOptions = append(podmanOptions, podmanArgs...)
```
empovit, line 517: `podmanOptions` and `podmanArgs` are confusing. Maybe `podmanArgs` and `prodmanAdditionalArgs`, or `podmanCoreArgs` and `podmanArgs`.

subsystem/utils.go
```diff
@@ -12,16 +12,23 @@ import (
...
+	WireMockURLFromSubsystemHost = fmt.Sprintf("http://127.0.0.1:%s", os.Getenv("WIREMOCK_PORT"))
+	AssistedServiceURLFromAgent  = fmt.Sprintf("http://wiremock:%s", os.Getenv("WIREMOCK_PORT"))
+	RequestsURL                  = fmt.Sprintf("%s/__admin/requests", WireMockURLFromSubsystemHost)
+	MappingsURL                  = fmt.Sprintf("%s/__admin/mappings", WireMockURLFromSubsystemHost)
+)
+
+const (
+	maxTimeout  = 300 * time.Second
+	agentDocker = "agent"
+	agentImage  = "quay.io/ocpmetal/assisted-installer-agent:latest"
```
empovit, line 31: It's a known problem that this suit always test an image with this particular name and tag. If build with a custom name (which is possible), the official latest will be tested instead. Can you think of a way to run these on `$ASSISTED_INSTALLER_AGENT`?

Makefile
```diff
@@ -11,7 +14,6 @@ GIT_REVISION := $(shell git rev-parse HEAD)
 PUBLISH_TAG := $(or ${GIT_REVISION})
 
 DOCKER_COMPOSE=docker-compose -f ./subsystem/docker-compose.yml
-export WIREMOCK_PORT = 8362
```
empovit, line 14: Why did you need a separate file for that? IMO you could've declared a variable with a default value just like we do for the other parameters.

subsystem/utils.go
```diff
@@ -488,3 +494,38 @@ func setReplyStartAgent(hostID string) {
...
+
+	args := make([]string, 0)
+	args = append(args, podmanOptions...)
+	args = append(args, agentImage)
+	args = append(args, commandArgs...)
+
+	return &models.Step{
+		StepType: stepType,
+		StepID:   stepID,
+		Command:  "docker",
```
empovit, line 528: `podmanArguments` but the command is `docker`?

Makefile
YuviGold: It is needed in order to call `docker-compose` without the Makefile.
By default, docker-compose looks for [.env file](https://docs.docker.com/compose/env-file/)

PR #146, opened 2021-01-20
MGMT-3526 Add container_image_availability tool

YuviGold:
The new container_image_avilability gets a list of images and checks if they are pullable.
This tool doesn't have a statemachine but it counts on 'pull' mechanism.
In case an image has been pulled before, we would still check that the remote image is pullable.

Each pull is restricted within a timeout. Eventually, each image will return an indication if it succeeded and the size/time rate it took to download the image.

/cc @avishayt @ori-amizur @filanov


src/container_image_availability/container_image_availability.go
```diff
@@ -0,0 +1,94 @@
...
+	Execute(command string, args ...string) (stdout string, stderr string, exitCode int)
+}
+
+type ProcessExecuter struct{}
+
+func (e *ProcessExecuter) Execute(command string, args ...string) (stdout string, stderr string, exitCode int) {
+	return util.Execute(command, args...)
+}
+
+func isImageAvailable(executer ImageAvailabilityDependencies, containerTool, image string) bool {
```
romfreiman, line 28: why do u need this logic at all? Why not just to pull and rely on the fact that podman knows what to do?
Why to add extra logic?

YuviGold: Before pulling the image layers, `pull` operation checks for metadata, and only then in case no differences with the local layers - the pull return success.

In case of a network failure, the metadata will fail to get downloaded, and the pull operation will fail - even if the image already exists locally.

For example, here I'm tagging an image with a different name to demonstrate an image that can't be pulled from the registry but does exist in the local registry.
```bash
➜  ~ docker tag quay.io/ocpmetal/assisted-installer-controller:stable quay.io/ocpmetal/assisted-installer-controller:abc
➜  ~ docker images quay.io/ocpmetal/assisted-installer-controller:abc                                             
REPOSITORY                                       TAG                 IMAGE ID            CREATED             SIZE
quay.io/ocpmetal/assisted-installer-controller   abc                 9aa9bd22db5a        2 days ago          510MB
➜  ~ docker pull quay.io/ocpmetal/assisted-installer-controller:abc  
Error response from daemon: manifest for quay.io/ocpmetal/assisted-installer-controller:abc not found: manifest unknown: manifest unknown

```

ori-amizur: Aren't we interested to know about the fact that image is pullable? If yes then it is not interesting if it exists locally

YuviGold: We want to know if the image is ready for installation.

romfreiman:
no it's not good. If there is no network, you'll fail anyhow. so better not
to start, right?

YuviGold:
> no it's not good. If there is no network, you'll fail anyhow. so better not
> to start, right?
> 
> On Wed, Jan 20, 2021, 18:38 Yuval Goldberg <notifications@github.com> wrote:
> 
> > *@YuviGold* commented on this pull request.
> > ------------------------------
> >
> > In src/container_image_availability/container_image_availability.go
> > <https://github.com/openshift/assisted-installer-agent/pull/146#discussion_r561110916>
> > :
> >
> > > +)
> > +
> > +var supportedContainerTools = [...]string{"podman", "docker"}
> > +
> > +//go:generate mockery -name ImageAvailabilityDependencies -inpkg
> > +type ImageAvailabilityDependencies interface {
> > +	Execute(command string, args ...string) (stdout string, stderr string, exitCode int)
> > +}
> > +
> > +type ProcessExecuter struct{}
> > +
> > +func (e *ProcessExecuter) Execute(command string, args ...string) (stdout string, stderr string, exitCode int) {
> > +	return util.Execute(command, args...)
> > +}
> > +
> > +func isImageAvailable(executer ImageAvailabilityDependencies, containerTool, image string) bool {
> >
> > We want to know if the image is ready for installation.
> > So if the image exists locally that's good.
> >
> > —
> > You are 

@eranco74 @avishayt WDYT?

src/container_image_availability/container_image_availability.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/pkg/errors"
+	"github.com/thoas/go-funk"
+
+	"github.com/sirupsen/logrus"
+)
+
+var supportedContainerTools = [...]string{"podman", "docker"}
+
+//go:generate mockery -name ImageAvailabilityDependencies -inpkg
+type ImageAvailabilityDependencies interface {
```
eranco74, line 18: Why Dependencies?

src/container_image_availability/container_image_availability.go
eranco74: I agree with Rom and Ori.
Having the image under a different name won't help us when podman run will fail.
If we can't pull the image we want to know.

src/container_image_availability/container_image_availability.go
```diff
@@ -0,0 +1,94 @@
...
+	return util.Execute(command, args...)
+}
+
+func isImageAvailable(executer ImageAvailabilityDependencies, containerTool, image string) bool {
+	stdout, _, exitCode := executer.Execute(containerTool, "images", "-q", image)
+	return exitCode == 00 && stdout != ""
+}
+
+func pullImage(executer ImageAvailabilityDependencies, pullTimeoutSeconds int, containerTool, image string) error {
+	stdout, stderr, exitCode := executer.Execute("timeout", strconv.FormatInt(int64(pullTimeoutSeconds), 10), containerTool, "pull", image)
```
eranco74, line 34: Note that in case of slow network, The command will fail, but some of the image layers may get cached, after a few runs this function might return nil although we have a network issue.

src/container_image_availability/main/main.go
```diff
@@ -0,0 +1,40 @@
...
+	Request       string
+}
+
+var executableConfig Config
+
+func processArgs() {
+	ret := &executableConfig
+	flag.StringVar(&ret.ContainerTool, "container-tool", "", "The container tool to be used to pull images")
+	flag.IntVar(&ret.PullTimeout, "timeout", 0, "Timeout in seconds for a pull operation")
+	flag.StringVar(&ret.Request, "request", "", "The request details. See models.ContainerImageAvailabilityRequest")
```
eranco74, line 26: Consider adding example format

src/container_image_availability/main/main.go
```diff
@@ -0,0 +1,40 @@
...
+
+	flag.Parse()
+}
+
+func main() {
+	processArgs()
+	config.ProcessSubprocessArgs(config.DefaultLoggingConfig)
+	util.SetLogging("container_image_availability", config.SubprocessConfig.TextLogging, config.SubprocessConfig.JournalLogging)
```
eranco74, line 34: Since the logging isn't depending on any args, move it up and log: "Starting  image availability"
Consider adding another log for the executableConfig

src/container_image_availability/container_image_availability.go
```diff
@@ -0,0 +1,94 @@
...
+
+func pullImageIfNotExists(executer ImageAvailabilityDependencies, pullTimeoutSeconds int, containerTool, image string) error {
+	if isImageAvailable(executer, containerTool, image) {
+		return nil
+	}
+
+	return pullImage(executer, pullTimeoutSeconds, containerTool, image)
+}
+
+func Run(containerTool string, pullTimeoutSeconds int, requestStr string,
```
eranco74, line 54: I think this method should log how long it took.
This way we can parse the response on the server side and add some metric around it.
There is a big difference between a node that pull the image in5 seconds vs an image that took 90 seconds.
I don't know what timeout we plan to start with (default timeout is 0 == infinite), but I'm guessing that we don't want to fail on this validation yet.
Once we will collect some data we can adjust the timeout.

src/container_image_availability/main/main.go
```diff
@@ -0,0 +1,40 @@
...
+	PullTimeout   int
+	Request       string
+}
+
+var executableConfig Config
+
+func processArgs() {
+	ret := &executableConfig
+	flag.StringVar(&ret.ContainerTool, "container-tool", "", "The container tool to be used to pull images")
+	flag.IntVar(&ret.PullTimeout, "timeout", 0, "Timeout in seconds for a pull operation")
```
eranco74, line 25: Mention that 0 means no timeout

src/container_image_availability/container_image_availability.go
YuviGold: In order to define the dependencies used by this module. It's easier that way to generate mocks.

src/container_image_availability/container_image_availability.go
YuviGold: In case of a network issue, the manifest won't be able to be downloaded and this command will fail.

romfreiman: I would not address all the corner cases in this itiration. Lets understand how we monitor it, and analyze later on how it should (and whether) be imrpoved

src/container_image_availability/container_image_availability.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+
+		if err == nil {
+			result = models.ContainerImageAvailabilityResultSuccess
+		}
+
+		response.Images = append(response.Images, &models.ContainerImageAvailability{
+			Name:   image,
+			Result: result,
+			Time:   int64(diff / time.Second),
```
ronniel1, line 74: I think we should return the rate of the pull - size/time and not the time.
This will provide good results even in the edge cases mentioned above 
also results will be consistent on multiple versions of the image

Also in the metrics we'll want to look at the rate

ronniel1: Maybe return the rate + size 
that will provide us with all the info we can hope for :)

eranco74: Good point, I guess you can get the image size once you pulled it.

src/container_image_availability/container_image_availability.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+type ProcessExecuter struct{}
+
+func (e *ProcessExecuter) Execute(command string, args ...string) (stdout string, stderr string, exitCode int) {
+	return util.Execute(command, args...)
+}
+
+func pullImage(executer ImageAvailabilityDependencies, pullTimeoutSeconds int, containerTool, image string) error {
+	stdout, stderr, exitCode := executer.Execute("timeout", strconv.FormatInt(int64(pullTimeoutSeconds), 10), containerTool, "pull", image)
```
eranco74, line 30: This doesn't handle the case where some of the layers already exist.

src/container_image_availability/main/main.go
```diff
@@ -0,0 +1,43 @@
...
+
+	if executableConfig.ContainerTool == "" || executableConfig.Request == "" {
+		flag.CommandLine.Usage()
+		os.Exit(1)
+	}
+}
+
+func main() {
+	util.SetLogging("container_image_availability", config.SubprocessConfig.TextLogging, config.SubprocessConfig.JournalLogging)
+	processArgs()
```
eranco74, line 36: Log the input
```suggestion
	processArgs()
	log.StandardLogger().Infof("Checking image availability, tool: %s, requested images: %s", executableConfig.ContainerTool, executableConfig.Request)
```

src/container_image_availability/container_image_availability.go
YuviGold: Seems clearer to me

src/container_image_availability/container_image_availability.go
eranco74: @romfreiman, @ronniel1  is this the desired behavior?

ronniel1: We are doing this only once at the end.
Also since the metric we'll look at is the rate of download, even if in the first pull, some layers are cached already, that is ok

src/container_image_availability/container_image_availability.go
```diff
@@ -0,0 +1,138 @@
...
+			log.WithError(err).Warnf("Couldn't get the image size of %s", image)
+		} else {
+			result = models.ContainerImageAvailabilityResultSuccess
+		}
+	}
+
+	return &models.ContainerImageAvailability{
+		Name:         image,
+		Result:       result,
+		Time:         diff / float64(time.Second),
```
eranco74, line 106: If the image is in the local cache, we should not report metric at all

subsystem/container_image_availability_test.go
```diff
@@ -0,0 +1,196 @@
...
+	return true
+}
+
+func pullImage(containerTool, image string) bool {
+	_, _, exitCode := util.Execute(containerTool, "pull", image)
+	return exitCode == 0
+}
+
+func isImageAvailable(containerTool, image string) bool {
+	stdout, _, exitCode := util.Execute(containerTool, "images", "-q", image)
```
eranco74, line 189: Nice.
Please add a comment why the -q is needed.

subsystem/container_image_availability_test.go
```diff
@@ -0,0 +1,196 @@
...
+	return exitCode == 0 && stdout != ""
+}
+
+func removeImage(containerTool, image string) bool {
```
eranco74, line 193: YAGNI

eranco74 (commented):
/lgtm

subsystem/container_image_availability_test.go
omertuc, line 189: Use long form `--quiet` in code. Relevant to other commands in the code as well

src/container_image_availability/container_image_availability.go
```diff
@@ -0,0 +1,153 @@
...
+
+	"github.com/openshift/assisted-installer-agent/src/util"
+	"github.com/openshift/assisted-service/models"
+	"github.com/pkg/errors"
+	"github.com/thoas/go-funk"
+
+	"github.com/sirupsen/logrus"
+)
+
+var supportedContainerTools = [...]string{"podman", "docker"}
```
omertuc, line 18: Do we even have Docker on the agent?

src/container_image_availability/container_image_availability.go
omertuc: > I don't understand where dependencies come from, what's wrong with: `type ImageAvailability interface`

@eranco74 it's already a common practice in this repo to do it in this manner

src/container_image_availability/container_image_availability.go
```diff
@@ -0,0 +1,153 @@
...
+		} else {
+			response.Result = models.ContainerImageAvailabilityResultSuccess
+		}
+
+		response.SizeBytes = sizeInBytes
+	}
+
+	if response.Result == models.ContainerImageAvailabilityResultSuccess && !imageExistLocallyBeforePull {
+		response.Time = diff / float64(time.Second)
+		response.DownloadRate = calcMBps(response.SizeBytes, diff)
```
omertuc, line 121: This is rather misleading - a docker pull command a lot of times skips a lot of work because of cached layers. They should be taken into account if you want to claim that you're calculating DownloadRate

src/container_image_availability/container_image_availability.go
```diff
@@ -0,0 +1,153 @@
...
+	val := strings.TrimSpace(stdout)
+	size, err := strconv.ParseFloat(val, 64)
+	if err != nil {
+		return 0, errors.Wrapf(err, "Failed to convert %s to float", val)
+	}
+
+	return size, nil
+}
+
+func calcMBps(bytes, nanosecond float64) float64 {
```
omertuc, line 60: `calcMiBps`

src/container_image_availability/container_image_availability.go
eranco74: We don't have docker on the iso, I believe this is for tests.

YuviGold: I wrote at the description

>The tool must get container-tool as a parameter in order to be tested with docker in the subsystem and with podman in production. I added a validation inside the code that those 2 are the only options (so no security risk for the executable command).

But I might have an idea how to solve that.
I can try to keep it `podman` and I will just soft link `podman` to `docker` in the subsystem framework.

src/container_image_availability/container_image_availability.go
YuviGold: We decided that maintaining a state-machine for pull operation might be too much work.
Do you have a suggestion on how it can be done without checking each layer?

src/container_image_availability/container_image_availability.go
YuviGold: Why?

omertuc: Because you're using 1 << 20 which is a mebibyte, not a megabyte

src/container_image_availability/container_image_availability.go
omertuc: Check if you can query docker/podman to check which layers it's going to pull and what is the size of each layer before pulling and which of these layers already exists and then you can accurately measure download rate.

or

Don't call it DownloadRate. Remove the DownloadRate field, `Time` is enough.

DownloadRate, as it is now, mostly misleads the user - layers are likely to be shared among the various images and therefore cached. You're going to see unreasonably fast DownloadRate on some of the pulls which is in fact completely wrong.

You're also leaving the user to wonder why the images have such different download speeds, when in fact it was just cached.

src/container_image_availability/container_image_availability.go
YuviGold: Cool that works. Changing.

src/container_image_availability/container_image_availability.go
YuviGold: Once the image got already pulled - it stops calculating the time & download rate.
It is easier than the agent already calculates it, then the service will decide what to do with it.

`podman` doesn't support inspecting remote registries as `skopeo` does. It will require a lot of other changes. Can be done in the future to improve this functionality.

omertuc: > Once the image got already pulled - it stops calculating the time & download rate.

I'm not talking about cases where it's already pulled - I'm talking about cases where two different images share some of the layers.

> It is easier that the agent already calculates it, then the service will decide what to do with it.
The calculation is meaningless / misleading if you assume cached layers on pull. And cached layers are very likely if two images are using the same `FROM`. Reporting this metric to the server with the name `DownloadRate` is very misleading. Right now, me and you, are aware of cached layers distorting it and it being possibly completely wrong / off. But someone months from now will have no idea about this issue, and won't know they can't trust this field and that it might be completely wrong. That's why I'm suggesting removing it all together. You can add the ImageSize field, you can add `boolean used_cache` true or false, but `DownloadRate` is just very possibly wrong. 

> `podman` doesn't support inspecting remote registries as `skopeo` does. It will require a lot of other changes. Can be done in the future to improve this functionality.

So that means the first option I offered is not relevant. The second one (removing the field) still stands.

src/container_image_availability/container_image_availability.go
```diff
@@ -0,0 +1,147 @@
...
+	start := time.Now()
+	err := pullImage(executer, pullTimeoutSeconds, image)
+	diff := float64(time.Since(start))
+
+	if err != nil {
+		log.WithError(err).Warnf("Pulling image %s wasn't available", image)
+	} else {
+		log.Infof("Pulling image %s is available. Took %f seconds", image, diff/float64(time.Second))
+
+		sizeInBytes, err := getImageSizeInBytes(executer, image)
```
eranco74, line 109: There is no need for this if the imageExistLocallyBeforePull is true.
Consider refactoring, and moving it under the if in line 119

src/container_image_availability/container_image_availability.go
YuviGold: Will be done as a separate task.

eranco74: https://issues.redhat.com/browse/MGMT-3852

YuviGold:
This has to be done after this PR will be merged.

PR #147, opened 2021-01-21
Move cloudctl/dhcp:ubi to cloudctl/dhcp:latest

YuviGold:
quay.io/cloudctl/dhcp:ubi tag was deleted. For now moving to latest tag


oshercc (approved):

usrbinkat:
I'm the image maintainer. Recommend attaching to image tag `micro` instead.

PR #149, opened 2021-01-24
MGMT-3390 - Added lint and format - for now, lint not performed automatically

ronniel1:


YuviGold (approved):
LGTM

ronniel1:
Removed the `downstream-change` label since the change is in the build dockerfile.

@gamli75  no need to modify downstream, right?

gamli75:
@ronniel1 no need to change the DS, juts remove the label

Makefile
```diff
@@ -17,12 +17,19 @@ DOCKER_COMPOSE=docker-compose -f ./subsystem/docker-compose.yml
 
 all: build
 
+lint: format
+	golangci-lint run -v
+
+format:
+	goimports -w -l src/ subsystem/ || /bin/true
```
omertuc, line 24: `format` should be like in `assisted-service`:

```
format:
	golangci-lint run --fix -v
```

omertuc:
/hold
Prevent auto-merge until my review is resolved

Makefile
ronniel1: nice.
Changed

omertuc:
Not exactly what I meant, we still have separate `lint:` and `format:` in the service, but there's no real reason to have that separation that I know of, so no big deal.

YuviGold (approved):

PR #153, opened 2021-01-27
MGMT-3390 enable liner as part of build step. On jenkins use concurre…

ronniel1:
…ncy=1


YuviGold:
```
make: golangci-lint: Command not found
```
/hold

PR #157, opened 2021-02-02
MGMT-3526 Support image pull network failures

YuviGold:


src/container_image_availability/container_image_availability.go
```diff
@@ -90,24 +90,37 @@ func pullImage(executer ImageAvailabilityDependencies, pullTimeoutSeconds int, i
...
 		Name:   image,
 		Result: models.ContainerImageAvailabilityResultFailure,
 	}
 
 	start := time.Now()
-	err := pullImage(executer, pullTimeoutSeconds, image)
-	diff := float64(time.Since(start))
+	pullTimeInSeconds := 0.0
+
+	for {
```
ronniel1, line 103: Why are we doing it in a loop?

YuviGold:
Closing since it will ignore failures which we would like to know about immediately.

PR #168, opened 2021-02-14
Fix IPv6 CIDR based on route

empovit:
Find the "real" CIDR of IPv6 addresses, so that checking if an API VIP
belongs to a subnet on the machine works correctly. 
Otherwise all IPv6 addresses are /128 (one host).


src/apivip_check/apivip_check.go
```diff
@@ -115,21 +127,43 @@ func getIPByURI(uri string) (net.IP, error) {
...
+
+		if !isVIPv4 {
+			util.SetV6PrefixesForAddress(intf.Name, &routeFinder{}, log, addrStrs)
+		}
+
+		var ipNet *net.IPNet
+		var err error
+		for _, ipAddr := range addrStrs {
+
+			if isVIPv4 {
```
ori-amizur, line 154: Why is this condition needed? Why not just parse and check the IP is contained in the CIDR?

src/inventory/interfaces.go
```diff
@@ -184,3 +185,15 @@ func (i *interfaces) getInterfaces() []*models.Interface {
 func GetInterfaces(depenndecies IDependencies) []*models.Interface {
 	return newInterfaces(depenndecies).getInterfaces()
 }
+
+func setV6PrefixesForAddresses(interfaces []*models.Interface, dependencies IDependencies) {
+
+	for _, intf := range interfaces {
+
+		if len(intf.IPV6Addresses) == 0 {
```
ori-amizur, line 193: This condition is not needed

src/util/network.go
```diff
@@ -3,9 +3,121 @@ package util
...
+		return err
+	}
+
+	for i, addr := range addresses {
+
+		if addr == "" {
+			continue
+		}
+
+		ip, _, err := ToIPv6CIDR(addr)
```
ori-amizur, line 65: Why was this added?

src/inventory/interfaces.go
empovit: it was there, probably some kind of optimization

src/apivip_check/apivip_check.go
empovit: It is needed to skip trying IPv6 subnets with IPv4 API VIP and vice versa. Notice `ipNet != nil` in line 165.

src/util/network.go
empovit: I don't want to traverse all routes trying to find the correct CIDR when the IP is not even an IPv6. Line 71.

src/inventory/interfaces.go
empovit: I believe it's there to avoid fetching a link by name and collecting usable routes if there are no IP addresses to update

YuviGold:
IPv6 isn't being tested from here. I'll add IPv6 job as non-blocking.

src/apivip_check/apivip_check.go
```diff
@@ -7,15 +7,27 @@ import (
...
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	"github.com/vishvananda/netlink"
 )
 
 const (
 	WorkerIgnitionPath = "/config/worker"
 )
 
+type routeFinder struct{}
```
ori-amizur, line 21: It is strange that the logic moved to util but we call netlink here.  Should be in util as well

empovit: moved RouteFinder based on netlink into utils.

PR #175, opened 2021-02-25
MGMT-3854 report log progress

slaviered:


slaviered:
@ybettan

src/logs_sender/send_logs.go
```diff
@@ -74,19 +86,29 @@ func (e *LogsSenderExecuter) FileUploader(filePath string, clusterID strfmt.UUID
...
+func (e *LogsSenderExecuter) LogProgressReport(clusterID strfmt.UUID, hostID strfmt.UUID, inventoryUrl string, pullSecretToken string, progress models.LogsState) error {
+	params := installer.UpdateHostLogsProgressParams{
+		ClusterID: clusterID,
+		HostID:    hostID,
+		LogsProgressParams: &models.LogsProgressParams{
+			LogsState: progress,
+		},
+	}
 
+	client, ctx := getClient(inventoryUrl, pullSecretToken)
```
eranco74, line 110: I think this method should get the client and not create it every time it updates the log progress.
It will be even better if the LogsSenderExecuter struct will get the client since it seems that few functions call  `getClient`

dhcpd.log
```diff
@@ -0,0 +1,15 @@
+Attaching to dhcpd
+[36mdhcpd       |[0m Internet Systems Consortium DHCP Server 4.3.6
```
tsorya, line 2: what is this file?

wiremock.log
```diff
@@ -0,0 +1,19 @@
+Attaching to subsystem_wiremock_1
+[36mwiremock_1  |[0m SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
```
tsorya, line 2: and this?

dhcpd.log
slaviered: removed

wiremock.log
slaviered: removed

PR #176, opened 2021-02-25
MGMT-3869 - Add domain resolution command to agent

omertuc:
This is part of the epic to add DNS validations to the assisted service.
The domain validation agent is very simple - it receives a list of domains as input, and outputs a list of IPv4/IPv6 addresses for each requested domain.

This agent allows the service to make sure the cluster hosts are capable of resolving some domain names  that are necessary for the installation to complete properly. Sometimes those domains have to be manually configured by the user, and this allows the service to validate that the user configured them correctly.

Example:
```bash
docker run \
quay.io/ocpmetal/assisted-installer-agent:latest \
domain_resolution -request \
'{"domains":[{"domain_name":"example.com"},{"domain_name":"example.net"},{"domain_name":"example.org"}]}' \
| jq
```

Results in:
```json
{
  "resolutions": [
    {
      "domain_name": "example.com",
      "ipv4_addresses": [
        "93.184.216.34"
      ],
      "ipv6_addresses": [
        "2606:2800:220:1:248:1893:25c8:1946"
      ]
    },
    {
      "domain_name": "example.net",
      "ipv4_addresses": [
        "93.184.216.34"
      ],
      "ipv6_addresses": [
        "2606:2800:220:1:248:1893:25c8:1946"
      ]
    },
    {
      "domain_name": "example.org",
      "ipv4_addresses": [
        "93.184.216.34"
      ],
      "ipv6_addresses": [
        "2606:2800:220:1:248:1893:25c8:1946"
      ]
    }
  ]
}
```


src/domain_resolution/domain_resolution.go
```diff
@@ -0,0 +1,77 @@
...
+
+//go:generate mockery -name DomainResolutionDependencies -inpkg
+type DomainResolutionDependencies interface {
+	Resolve(domain string) (ips []net.IP, err error)
+}
+
+type DomainResolver struct{}
+
+func (e *DomainResolver) Resolve(domain string) (ips []net.IP, err error) {
+	return net.LookupIP(domain)
```
omertuc, line 19: In practice we're simply using `net.LookupIP(domain)` which returns a list of `net.IP` addresses

src/domain_resolution/domain_resolution.go
```diff
@@ -0,0 +1,77 @@
...
+
+	ips, err := resolver.Resolve(domain)
+
+	if err != nil {
+		// Error during resolution, just return an empty result without any IP addresses
+		return result
+	}
+
+	for _, ip := range ips {
+		if ip.To4() != nil {
```
omertuc, line 37: `net.IP` objects are just a type alias for `[]byte` - IPv4 is simply a 4 byte array, IPv6 is simply a 16 byte array. `ip.To4` fails for IPv6 addresses, we use that fact to detect whether the IP is an IPv4 or IPv6. We do the same for IPv6 addresses. This is how we know whether to put them inside the IPV4Addresses array vs the IPV6Addresses array.

Note that the output response format expects a strfmt.IPv4 (strfmt is a go-swagger package) which is a type alias for a string.

eranco74: Just curious, why do we care if it's IPv4 or IPv6?

eranco74: Just curious, do we actually care if it's IPv4 or IPv6 in the service side?
Except for the swagger definition (that can be updated to a pattern: '^(?:(?:(?:[0-9]{1,3}\.){3}[0-9]{1,3})|(?:(?:[0-9a-fA-F]*:[0-9a-fA-F]*){2,}))$' just like we have for the VIPs) is there any other reason for it?

src/domain_resolution/domain_resolution.go
```diff
@@ -0,0 +1,77 @@
...
+	result := models.DomainResolutionResponseDomain{
+		DomainName: &domain,
+		IPV4Addresses: make([]strfmt.IPv4, 0),
+		IPV6Addresses: make([]strfmt.IPv6, 0),
+	}
+
+	ips, err := resolver.Resolve(domain)
+
+	if err != nil {
+		// Error during resolution, just return an empty result without any IP addresses
```
eranco74, line 32: Add logging of the error

src/domain_resolution/domain_resolution_test.go
```diff
@@ -0,0 +1,180 @@
...
+				Domains: []*models.DomainResolutionRequestDomain{
+					{DomainName: &domains[0]},
+					{DomainName: &domains[1]},
+					{DomainName: &domains[2]},
+				},
+			}
+			b, err := json.Marshal(request)
+			Expect(err).ShouldNot(HaveOccurred())
+
+			// Arbitrary amounts of ipv4 and ipv6 addresses returned for each domain
```
eranco74, line 113: I think the max possible result may have a single IPv6 and a single IPv4.
No reason to add impossible result in the test (a bit confusing).

src/domain_resolution/domain_resolution_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+			// Make sure all requested domains appear in the response
+			Expect(response.Resolutions).Should(HaveLen(len(domains)))
+
+			// Check resolved IP addresses are the same as what the mock resolver returned for each domain
+			for _, zip := range funk.Zip(domains, response.Resolutions) {
+				requestedDomain, _ := zip.Element1.(string)
+				resolution, _ := zip.Element2.(*models.DomainResolutionResponseDomain)
+
+				Expect(err).ToNot(HaveOccurred())
```
eranco74, line 140: What err are you testing?

src/domain_resolution/domain_resolution.go
omertuc: Don't know if we'll care, but I thought that separation is might be valuable for whatever reasons - I figured it would be easier to do in the agent side while it's still in bytes-form, before it gets turned into strings. T

The service can easily merge those arrays into a single one if it doesn't want to deal with the separation / it doesn't care

eranco74 (commented):
Nice work.
Looks good

src/domain_resolution/domain_resolution_test.go
omertuc: >I think the max possible result may have a single IPv6 and a single IPv4.
Thought so myself, but couldn't find any evidence of that in the go documentation / source code, so I just left the code to assume there can be more than one of each

>No reason to add impossible result in the test (a bit confusing).
Will change it to some lower number, but still >1

src/domain_resolution/domain_resolution_test.go
omertuc: whoops, just some leftover

src/domain_resolution/domain_resolution_test.go
eranco74: Conceptually speaking DNS must resolve a single IP (per protocol).
Yet, In case the net.LookupIP is doing multiple DNS queries for the same domain name on every call (very unlikely) you might get more than a single IP.

omertuc: > Conceptually speaking DNS must resolve a single IP (per protocol).
That's what I know too, but the `LookupIP` documentation not mentioning that fact and simply returning an array made me second guess myself and think that in some edge cases it might not be the case.

Do you propose we change the swagger API to simply be 2 strings - one for IPv4, one for IPv6 instead of arrays?

omertuc:

Will update downstream

PR #179, opened 2021-03-04
MGMT-4052 Add by-id information to the inventory

michaellevy101:


src/inventory/disks.go
```diff
@@ -19,6 +19,34 @@ func newDisks(dependencies IDependencies) *disks {
 	return &disks{dependencies: dependencies}
 }
 
+func (d *disks) getDisksWWNs() map[string]string {
+	stdout, stderr, exitCode := d.dependencies.Execute("bash", "-c", `find /dev/disk/by-id/ \( -name wwn-* -o -name nvme-eui* \)  -type l | xargs -I % bash -c 'basename % | tr '\''\n'\''  '\''\t'\'' && readlink -f %'`)
```
omertuc, line 23: Can you replace this bash one liner with go code?

src/inventory/disks.go
```diff
@@ -130,6 +158,12 @@ func (d *disks) getDisks() []*models.Disk {
 		return ret
 	}
 
+	if len(blockInfo.Disks) == 0 {
```
omertuc, line 161: Is this really necessary? I think it's better getDisksWWNs handles 0 disks more gracedully rather than getDisks having to worry about it

src/inventory/disks.go
```diff
@@ -148,7 +182,10 @@ func (d *disks) getDisks() []*models.Disk {
 		}
 
 		path := d.getPath(disk.BusPath, disk.Name)
+		id := diskPath2diskWWN[path]
```
omertuc, line 185: Doesn't map access potentially return a not found error?

src/inventory/disks.go
omertuc: As discussed, we'll keep this `if` to prevent `getDisksWWNs` from doing unnecessary work

src/inventory/disks.go
omertuc: As discussed, it's intended behavior here to just get the zero-value of the string, `""`, when the key doesn't exist

src/inventory/disks.go
```diff
@@ -19,6 +21,47 @@ func newDisks(dependencies IDependencies) *disks {
...
+	}
+
+	matchingFiles := funk.Filter(filesInfo, func(fileInfo os.FileInfo) bool {
+		basename := filepath.Base(fileInfo.Name())
+
+		if !strings.HasPrefix(basename, "wwn-") && !strings.HasPrefix(basename, "nvme-eui") {
+			return false
+		}
+
+		return fileInfo.Mode() & os.ModeSymlink != 0
```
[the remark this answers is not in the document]
michaellevy101: 1. It should be a symlink so i verified it
2. Why replacing one call to two calls?

src/inventory/mock_FileInfo.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/stretchr/testify/mock"
+	"os"
+	"time"
+)
+
+type FileInfoMock struct {
+	mock.Mock
+}
+
+func (f *FileInfoMock) Name() string {
```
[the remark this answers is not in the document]
michaellevy101: I didn't create it just exported it to a separated file.

src/inventory/disks.go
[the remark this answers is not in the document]
omertuc: 1. What I meant that maybe in some systems it isn't? Is it guaranteed to be a symlink? Do we inherently care about it being one?
2. It's not just one call, it's all this code:
```filesInfo, err := d.dependencies.ReadDir("/dev/disk/by-id")

	if err != nil {
		logrus.Warnf("Cannot get disk/by-id information: %s", err)
		return make(map[string]string)
	}

	matchingFiles := funk.Filter(filesInfo, func(fileInfo os.FileInfo) bool {
		basename := filepath.Base(fileInfo.Name())

		if !strings.HasPrefix(basename, "wwn-") && !strings.HasPrefix(basename, "nvme-eui") {
			return false
		}

		return fileInfo.Mode() & os.ModeSymlink != 0
	})
```

without the symlink part, it can be replaced by something like
`files := append(filepath.Glob("/dev/disk/by-id/wwn-*", filepath.Glob("/dev/disk/by-id/nvme-eui*"))`

src/inventory/disks_test.go
```diff
@@ -118,6 +90,121 @@ func createAWSXenEBSDisk() *ghw.Disk {
...
+		IsRemovable:            false,
+		NUMANodeID:             0,
+		PhysicalBlockSizeBytes: 512,
+		StorageController:      ghw.STORAGE_CONTROLLER_SCSI,
+	}
+}
+
+func createWWNResults() map[string]string{
+	byidmapping := make(map[string]string)
+	byidmapping["/dev/sdb"] = "wwn-0x6141877064533b0020adf3bc0325d664"
```
omertuc, line 188: Maybe put this value (and others that are used in multiple places) in some global constant to to clarify that they're expected to match

src/inventory/mock_FileInfo.go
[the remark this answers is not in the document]
omertuc: But it seems to me that currently it is not being automatically generated, it should be, like all other mocked interfaces using the `//go:generate mockery ...`. I realize it wasn't that way before, but if you're already moving it to its own file - might as well just generate that file automatically. We should not manually maintain this mock code when we have a tool that generates it.

src/inventory/disks_test.go
michaellevy101: will do

src/inventory/mock_FileInfo.go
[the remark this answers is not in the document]
michaellevy101: sure, will do.

PR #182, opened 2021-03-11
MGMT-4048 change disk id to be the full path of the by-id link

michaellevy101:


src/inventory/disks.go
```diff
@@ -43,18 +43,18 @@ func (d *disks) getDisksWWNs() map[string]string {
 	// For example: wwn-0x6141877064533b0020adf3bc0325d664	-> /dev/sdb
 	// "wwn-0x6141877064533b0020adf3bc0325d664" is the disk id and the path is: /dev/sdb
 	return funk.Map(matchingFiles, func(fileInfo os.FileInfo) (string, string) {
-		diskId := fileInfo.Name()
-		diskPath, err := d.dependencies.EvalSymlinks(filepath.Join("/dev/disk/by-id", diskId))
+		diskId := filepath.Join("/dev/disk/by-id", fileInfo.Name())
```
omertuc, line 46: Lines 25 and 46 duplicate this path, it should be in a var

src/inventory/disks.go
```diff
@@ -68,7 +68,7 @@ func (d *disks) getPath(busPath string, diskName string) string {
 	if err == nil {
```
omertuc, line 68: GitHub won't allow me to comment but the first line of this function should be filepath.Join rather than Sprintf

michaellevy101: because i didn't change this method.... but i changed it

src/inventory/disks.go
michaellevy101: done

PR #184, opened 2021-03-17
MGMT-4117 ovs logs gather

slaviered:
/assign @eranco74

### DIFF
--- a/scripts/installer/ovs-installer-gather.sh
+++ b/scripts/installer/ovs-installer-gather.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+
+if test "x${1}" = 'x--id'
+then
+	GATHER_ID="${2}"
+	shift 2
+fi
+
+ARTIFACTS="/tmp/artifacts-${GATHER_ID}"
+mkdir -p "${ARTIFACTS}"
+
+exec &> >(tee "${ARTIFACTS}/ovs-gather.log")
+
+echo "Gather remote logs"
+export MASTERS=()
+if [[ -f ${LOG_BUNDLE_BOOTSTRAP_ARCHIVE_NAME} ]]; then
+    # run on the current node
+    MASTER_GATHER_ID="master-${GATHER_ID}"
+    MASTER_ARTIFACTS="/tmp/artifacts-${MASTER_GATHER_ID}"
+    mkdir -p "${MASTER_ARTIFACTS}/journals"
+    mkdir -p "${ARTIFACTS}/control-plane/master"
+    sudo /usr/local/bin/ovs-master-installer-gather.sh --id "${MASTER_GATHER_ID}" </dev/null
+elif [ "$#" -ne 0 ]; then
+    MASTERS=( "$@" )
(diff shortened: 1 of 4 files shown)


src/logs_sender/send_logs.go
```diff
@@ -123,25 +124,46 @@ func (e *LogsSenderExecuter) LogProgressReport(clusterID strfmt.UUID, hostID str
...
+	//Then, installer-gather.sh runa as overlay and finally bundles
+	//all logs together. Unlike installer-gather.sh, ovs-installer-gather.sh
+	//runs locally in the container rather on the host
+	ovsGatherArgs := append([]string{"--id", gatherID}, mastersIPs...)
+	log.Infof("Running %s %v", ovsGatherBin, ovsGatherArgs)
+	stdOut, stdErr, exitCode := e.Execute(ovsGatherBin, ovsGatherArgs...)
+	for _, so := range strings.Split(stdOut, "\n") {
+		log.Infof("ovs-gather log: %s", so)
+	}
+	if stdErr != "" || exitCode != 0 {
```
eranco74, line 141: Why do you care about the stdErr?
Isn't exitCode enough?

scripts/installer/ovs-installer-gather.sh
```diff
@@ -0,0 +1,37 @@
...
+echo "Gather remote logs"
+export MASTERS=()
+if [[ -f ${LOG_BUNDLE_BOOTSTRAP_ARCHIVE_NAME} ]]; then
+    # run on the current node
+    MASTER_GATHER_ID="master-${GATHER_ID}"
+    MASTER_ARTIFACTS="/tmp/artifacts-${MASTER_GATHER_ID}"
+    mkdir -p "${MASTER_ARTIFACTS}/journals"
+    mkdir -p "${ARTIFACTS}/control-plane/master"
+    sudo /usr/local/bin/ovs-master-installer-gather.sh --id "${MASTER_GATHER_ID}" </dev/null
+    #cp -r "${MASTER_ARTIFACTS}"/* "${ARTIFACTS}/control-plane/master/" - SARAH no need will be done by installer-gather itself
```
eranco74, line 23: I don't think this is required.
This code will be executed from the bootstrap node prior to the reboot, no?

eranco74 (commented):
Seems that it should work :smile_cat: 
Note that this is  a place holder until we deliver these services journal collection in the installer:
https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/usr/local/bin/installer-masters-gather.sh#L25

src/logs_sender/send_logs.go
slaviered: exitCode is always 0 since there is no set -e. See Yoni's comment in the original code

scripts/installer/ovs-installer-gather.sh
slaviered: Wouldn't this branch executes on single node?

slaviered: But I do need to take out the debug comment

PR #185, opened 2021-03-18
MGMT-4397 Agent changes - adding new tool for disk speed check

ori-amizur:
[MGMT-4397](https://issues.redhat.com/browse/MGMT-4397) Agent changes - adding new tool for disk speed check
    This tool should eventually replace fio_perf_check which we do not want to touch until it is not needed.
    2 main changes comparing to fio_perf_check:
    - Communicate through agent instead of directly to service
    - Remove threshold check, and instead pass fio results to the service, which should perform validate it against the threshold.

/cc @ronniel1 
/cc @tsorya


src/disk_speed_check/disk_speed_check.go
```diff
@@ -0,0 +1,93 @@
+package disk_speed_check
```
omertuc, line 1: This is almost 100% duplicated code from https://github.com/openshift/assisted-installer-agent/blob/53bceb5ea502f28530697242f1e4e970a6ed3da2/src/fio_perf_check/fio_perf_check.go, what am I missing?

omertuc:
Please write a commit message with what this new tool does, and why it was added

src/disk_speed_check/disk_speed_check.go
ori-amizur: @omertuc this file is a subset of that file.  The difference is that this one is communicating with the agent as a subprocess, while fio_perf_check.go sends results directly to the service, and this one does not get a threshold.  Instead, the threshold is part of the service.
The plan is eventually to remove the fio package when it is safe to do so.

ori-amizur:
> Please write a commit message with a short summary of what this new tool does, and why it was added.
> 
> The Jira ticket doesn't say anything. EDIT: I now noticed the Jira ticket is part of an epic, but it still doesn't explain why this new tool was added / what's its purpose

OK

PR #187, opened 2021-03-30
MGMT-4718 Capture network latency and packet loss at L3 network connectivity

jordigilh:
This PR extends the L3 connectivity information gathered by the agent by capturing the average latency (RTT) and the packet loss information per host connection.

Code has been refactored and unit tests extended to cover the new functionality as well as the aggregation process to validate the refactored code.

This [PR](https://github.com/openshift/assisted-service/pull/1432) is the counterpart in the assisted-service.

@pkliczewski @jakub-dzon @machacekondra @masayag @YuviGold please review at your discretion in the meanwhile.

/Jordi

### DIFF
--- a/src/commands/connectivity_check.go
+++ b/src/commands/connectivity_check.go
@@ -5,17 +5,16 @@ import (
 	"encoding/xml"
 	"os/exec"
 	"regexp"
+	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/openshift/assisted-installer-agent/src/util"
 	"github.com/openshift/assisted-installer-agent/src/util/nmap"
 	"github.com/openshift/assisted-service/models"
+	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 )
-
-type Done struct{}
-
-type Any interface{}
 
 func getOutgoingNics() []string {
 	ret := make([]string, 0)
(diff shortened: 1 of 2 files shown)


src/commands/connectivity_check.go
```diff
@@ -218,91 +252,96 @@ func analyzeNmap(dstAddr string, dstMac string, allDstMacs []string, srcNic stri
...
-		<-doneChan
+		<-doneCh
 	}
+	sendDone(syncCh)
+}
+
+type hostChecker struct {
+	l2IPv4Cmd    func(string, string, []string, string, chan *models.L2Connectivity)
+	l2IPv6Cmd    func(string, string) func() ([]byte, error)
+	l3Cmd        func(string, string) ([]byte, error)
```
YuviGold, line 310: Could you make it a separate dependencies interface like [NTPSynchornizerDependencies](https://github.com/openshift/assisted-installer-agent/blob/53bceb5ea502f28530697242f1e4e970a6ed3da2/src/ntp_synchronizer/ntp_synchronizer.go#L20-L25) so it would be easily mockable generated?

src/commands/connectivity_check.go
```diff
@@ -55,32 +54,73 @@ func getIPAddressFromCIDR(cidr string) string {
...
+func regexMatchFor(regex, line string) ([]string, error) {
+	r := regexp.MustCompile(regex)
+	p := r.FindStringSubmatch(line)
+	if len(p) < 2 {
+		return nil, errors.Errorf("unable to parse %s with regex %s", line, regex)
+	}
+	return p, nil
+}
+
+func parsePingCmd(conn *models.L3Connectivity, cmdOutput string) error {
```
YuviGold, line 97: Maybe we can use a Golang package instead? i.e. https://github.com/go-ping/ping
It would ease all the regex part

src/commands/connectivity_check.go
```diff
@@ -119,38 +159,39 @@ func macInDstMacs(mac string, allDstMacs []string) bool {
...
-		cmd := exec.Command("nmap", "-6", "-sn", "-n", "-oX", "-", "-e", srcNic, dstAddr)
-		analyzeNmap(dstAddr, dstMac, allDstMacs, srcNic, l2chan, cmd.Output)
+		analyzeNmap(dstAddr, dstMAC, allDstMACs, srcNIC, innerCh, l2IPv6Cmd(srcNIC, dstAddr))
 	}
+	sendDone(doneCh)
 }
 
-func runArping(dstAddr string, dstMac string, allDstMacs []string, srcNic string, l2chan chan Any) {
+func nmapCmd(srcNIC, dstAddr string) func() ([]byte, error) {
+	cmd := exec.Command("nmap", "-6", "-sn", "-n", "-oX", "-", "-e", srcNIC, dstAddr)
```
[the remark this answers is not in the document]
jordigilh: They were there already. I did not change them. This PR only captures the ping results, it does not change the cmd for nmap or arp.

src/commands/connectivity_check.go
jordigilh: Makes sense. I'll wrap it around.

src/commands/connectivity_check.go
jordigilh: We need to be able to specify the interface to use when we ping, I checked this library and the other one mentioned in that project and none looked like it allowed me to specify the network interface. I'd prefer to use a library than a regex for obvious reasons :wink:.

src/commands/connectivity_check.go
jordigilh: Done, let me know what you think :smile:

pkliczewski:
@YuviGold please review

src/commands/connectivity_check.go
```diff
@@ -55,254 +54,304 @@ func getIPAddressFromCIDR(cidr string) string {
...
+			log.Warnf("Unexpected mac address for arping %s on nic %s: %s", dstAddr, srcNIC, remoteMAC)
+		}
+		if strings.ToLower(dstMAC) != remoteMAC {
+			log.Infof("Received remote mac %s different then expected mac %s", remoteMAC, dstMAC)
+		}
+		dataCh <- ret
+	}
+}
+
+func checkConnectivity(conCheck connectivityCmd, outCh chan *models.ConnectivityRemoteHost) {
```
ori-amizur, line 326: The name checkHost was better because it indicates what is being checked

src/commands/connectivity_check.go
```diff
@@ -55,254 +54,304 @@ func getIPAddressFromCIDR(cidr string) string {
...
-		iret := <-ch
-		switch value := iret.(type) {
-		case *models.L2Connectivity:
-			ret.L2Connectivity = append(ret.L2Connectivity, value)
-		case *models.L3Connectivity:
-			ret.L3Connectivity = append(ret.L3Connectivity, value)
-		case Done:
-			numDone++
+		v := <-dataCh
+		if l3, ok := v.(*models.L3Connectivity); ok {
```
ori-amizur, line 338: Why change the switch case ?

src/commands/connectivity_check.go
```diff
@@ -55,254 +54,304 @@ func getIPAddressFromCIDR(cidr string) string {
...
+		v := <-innerChan
+		if l2, ok := v.(*models.L2Connectivity); ok && l2.Successful {
 			received = true
-			l2chan <- ret
-		case Done:
-			numDone++
+			dataCh <- l2
+			continue
 		}
+		numDone++
```
ori-amizur, line 228: Why was the switch case removed?
If Successful is false, are why counting it as done?

jordigilh: you're right. I'll add the check for Done and add the typed struct back to simplify the code.

src/commands/connectivity_check.go
jordigilh: I'll put it back. Easier to read with the switch case.

src/commands/connectivity_check.go
jordigilh: Which one? checkConnectivity -> checkHost or conCheck -> checkHost?

jordigilh: ok got it, checkConnectivity -> checkHost.

src/commands/connectivity_check.go
```diff
@@ -55,254 +54,306 @@ func getIPAddressFromCIDR(cidr string) string {
...
 	received := false
-	for numDone := 0; numDone != len(sourceNics); {
+	for numDone := 0; numDone != len(conCheck.getOutgoingNICs()); {
 		iret := <-innerChan
 		switch ret := iret.(type) {
 		case *models.L2Connectivity:
-			received = true
-			l2chan <- ret
-		case Done:
+			dataCh <- ret
```
ori-amizur, line 228: What about the received flag? When is it set ?

src/commands/connectivity_check.go
```diff
@@ -55,254 +54,307 @@ func getIPAddressFromCIDR(cidr string) string {
...
 	}
 	if !successful {
 		ret := &models.L3Connectivity{
 			RemoteIPAddress: address,
-			Successful:      false,
 		}
-		l3chan <- ret
+		dataCh <- ret
 	}
+	wg.Done()
```
ori-amizur, line 150: Maybe add defer wg.Done() in the beginning of the function instead of the end ?

src/commands/connectivity_check.go
```diff
@@ -55,254 +54,307 @@ func getIPAddressFromCIDR(cidr string) string {
...
-	addresses := getOutgoingAddresses(host.Nics)
-	ch := make(chan Any, 1000)
-	go l3CheckConnectivity(addresses, outgoingNics, ch)
-	go l2CheckConnectivity(host.Nics, outgoingNics, ch)
+		HostID:         conCheck.getHost().HostID,
+		L2Connectivity: []*models.L2Connectivity{},
+		L3Connectivity: []*models.L3Connectivity{},
+	}
+	addresses := getOutgoingAddresses(conCheck.getHost().Nics)
+	dataCh := make(chan any)
```
ori-amizur, line 336: Why channel with size 1 ?

src/commands/connectivity_check.go
```diff
@@ -55,254 +54,307 @@ func getIPAddressFromCIDR(cidr string) string {
...
+			go l2CheckAddress(address, destNic.Mac, allDstMACs, dataCh, doneCh, conCheck)
 		}
 	}
 	for i := 0; i != numAddresses; i++ {
-		<-doneChan
+		<-doneCh
+	}
+}
+
+type hostChecker struct {
```
ori-amizur, line 260: Suggest to put this part not in the middle of the other code.  Maybe at the end or other file?

src/commands/connectivity_check.go
```diff
@@ -55,254 +54,307 @@ func getIPAddressFromCIDR(cidr string) string {
...
 
-func l2CheckConnectivity(destinationNics []*models.ConnectivityCheckNic, sourceNics []string, l2chan chan Any) {
-	defer sendDone(l2chan)
-	doneChan := make(chan Any)
-	allDstMacs := make([]string, 0)
-	for _, destNic := range destinationNics {
-		allDstMacs = append(allDstMacs, destNic.Mac)
+func l2CheckConnectivity(dataCh chan any, conCheck connectivityCmd) {
+	defer sendDone(dataCh)
+	doneCh := make(chan any)
```
ori-amizur, line 243: Suggest to have single method to signal the go-routine terminates - either wait-group or done channel.  
It is not relevant to channels that the done information is sent along with the data.

src/commands/connectivity_check.go
jordigilh: because there is only 1 routine receiving the data from the channel. Having the channel with a bigger queue won't improve the time it takes for this single routine to process all the information generated by all the other routines.

src/commands/connectivity_check.go
jordigilh: sure.

src/commands/connectivity_check.go
jordigilh: ok. I'll move it to the end of the file.

src/commands/connectivity_check.go
jordigilh: ok. I'll refactor this part to use a wait group like in
https://github.com/openshift/assisted-installer-agent/blob/1499058eafcb4d9ddd329c9827fb79047a5a991a/src/commands/connectivity_check.go#L123-L124
It will save the need to use this `doneCh` to sync.

PR #188, opened 2021-04-01
NO-ISSUE: Use 'Errorf' instead of 'Infof' for step failure

pawanpinjarkar:
Logging the step execution failure with Error instead of Info message will help in analyzing logs quickly for failures.


YuviGold (approved):
LGTM

PR #189, opened 2021-04-07
NO-ISSUE Allow @pkliczewski's team to lgtm PRs

ronniel1:


OWNERS
```diff
@@ -6,6 +6,7 @@ filters:
       - code-reviewers
     reviewers:
       - code-reviewers
+      - reviewers-only
```
osherdp, line 9: bad naming
I think we should do ``code-approvers`` and ``code-reviewers`` and put ``code-approvers`` under ``approvers`` and under ``reviewers``

osherdp (changes requested):

OWNERS
YuviGold: Or maybe separate into actual teams like `core` and `edge`, etc

jordigilh:
@pkliczewski can you rebase?

OWNERS
jordigilh: `reviewers` sounds good to me, unless you expect more teams to join and be granted different roles. By the way, are we following a naming convention based on function or team's name? I think function makes more sense in the long run, but just my $0.02 :smile:

jordigilh:
> @jordigilh have a look at #212, let me know what you think

looks good to me. Thanks for taking care of it so swiftly :smile:

osherdp:
don't worry about it :)

PR #192, opened 2021-04-11
OCPBUGSM-27560 Duplicate installation disk

michaellevy101:
Use the regular path in case of duplicate by-path value

### DIFF
--- a/src/inventory/disks.go
+++ b/src/inventory/disks.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/jaypipes/ghw"
+	"github.com/jaypipes/ghw/pkg/block"
 	ghwutil "github.com/jaypipes/ghw/pkg/util"
 	"github.com/openshift/assisted-installer-agent/src/util"
 	"github.com/openshift/assisted-service/models"
@@ -181,7 +182,7 @@ func (d *disks) getDisks() []*models.Disk {
 
 	diskPath2diskWWN := d.getDisksWWNs()
 
-	for _, disk := range blockInfo.Disks {
+	for diskIndex, disk := range blockInfo.Disks {
 		var eligibility models.DiskInstallationEligibility
 		var isInstallationMedia bool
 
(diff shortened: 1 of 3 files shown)


src/inventory/disks.go
```diff
@@ -231,6 +232,26 @@ func (d *disks) getDisks() []*models.Disk {
 	return ret
 }
 
+// getBusPath - Support special case where two disk has the same busType.
+// We reproduce this case by creating a machine using virt-install with the following command which creates a machine
+// with two IDE disks HDD and CDROM and both has the same busType.
+// virt-install --name=master-1 --cdrom=$PATH/cluster-discovery.iso --vcpus=2 --ram=16384 --disk=size="$DISKGIB",pool="$POOL" --os-variant=rhel-unknown --network=bridge=virbr0,model=virtio --graphics=none --noautoconsole
+func getBusPath(d *disks, disks []*block.Disk, index int, busPath string) string {
```
omertuc, line 239: Why is this a function and not a method of disks?

src/inventory/disks.go
```diff
@@ -231,6 +232,26 @@ func (d *disks) getDisks() []*models.Disk {
 	return ret
 }
 
+// getBusPath - Support special case where two disk has the same busType.
```
omertuc, line 235: `have the same busPath`*

src/inventory/disks.go
```diff
@@ -231,6 +232,26 @@ func (d *disks) getDisks() []*models.Disk {
...
+	if busPath == ghwutil.UNKNOWN {
+		return ""
+	}
+
+	for i, disk := range disks {
+		if i == index || disk.BusPath != busPath{
+			continue
+		}
+
+		return ""
```
omertuc, line 249: Maybe add this comment: 
`// when two disks share the same bus path, we prefer they have no bus path at all to avoid confusing them`

src/inventory/disks.go
michaellevy101: no need to mock it...

omertuc: But it's a function that receives an instance of `*disks` as its first parameter, which is pretty much what a method is. How is mocking relevant?

omertuc:
Please edit the commit message to contain explanations on what changed (in general) and why it changed (you can copy most of it from the diff comments)

avishayt (approved):

PR #193, opened 2021-04-12
MGMT-4142 Using ghw to detect GPUs

jakub-dzon:
Because `lshw` is not installed on the agent image and `ghw` provides all the necessary GPU-related functionality, the code had to be migrated to `ghw`.


YuviGold (approved):
LGTM

PR #197, opened 2021-05-09
MGMT-4966 Display a better message when media disconnection failure o…

michaellevy101:
…ccurs

When a media disconnection error occurred the system became unstable and the step execution might be stuck on IO errors or failed with an unfriendly message.
Media disconnection occurred when a network drive (mostly PXE) has any errors (Network issue) and the controller disconnects it automatically.
Since this error is common to all the steps, we add this verification to the agent
The indication to error is reading from the file failed

### DIFF
--- a/src/commands/step_processor.go
+++ b/src/commands/step_processor.go
@@ -1,8 +1,14 @@
 package commands
 
 import (
+	"fmt"
+	"io"
 	"net/http"
+	"os"
+	"strings"
 	"time"
+
+	"github.com/pkg/errors"
 
 	"github.com/go-openapi/swag"
 
@@ -20,6 +26,14 @@ import (
 
 type HandlerType func(string, ...string) (stdout string, stderr string, exitCode int)
 
+type errorCode int
+
+// TODO - add an ErrorCode Enum to the swagger
+const (
+	Undetected        errorCode = 999
+	MediaDisconnected errorCode = 256
+)
+
 type stepSession struct {
 	session.InventorySession
 }
@@ -33,36 +47,30 @@ func newSession() *stepSession {
 	return &ret
 }
 
-func (s *stepSession) sendStepReply(stepType models.StepType, stepID, output, errStr string, exitCode int) {
-	if exitCode == 0 && alreadyExistsInService(stepType, output) {
-		s.Logger().Infof("Result for %s already exists in assisted service", string(stepType))
+func (s *stepSession) sendStepReply(reply models.StepReply) {
+	if reply.ExitCode == 0 && alreadyExistsInService(reply.StepType, reply.Output) {
+		s.Logger().Infof("Result for %s already exists in assisted service", string(reply.StepType))
 		return
 	}
 	logFunc := s.Logger().Infof
-	if exitCode != 0 {
+	if reply.ExitCode != 0 {
 		logFunc = s.Logger().Warnf
 	}
 
-	if stepType == models.StepTypeFreeNetworkAddresses {
+	if reply.StepType == models.StepTypeFreeNetworkAddresses {
 		// the free-addresses step's output spams the log too much
-		logFunc("Sending step <%s> reply error <%s> exit-code <%d>", stepID, errStr, exitCode)
+		logFunc("Sending step <%s> reply error <%s> exit-code <%d>", reply.StepID, reply.Error, reply.ExitCode)
 	} else {
-		logFunc("Sending step <%s> reply output <%s> error <%s> exit-code <%d>", stepID, output, errStr, exitCode)
+		logFunc("Sending step <%s> reply output <%s> error <%s> exit-code <%d>", reply.StepID, reply.Output, reply.Error, reply.ExitCode)
 	}
 
 	params := installer.PostStepReplyParams{
 		HostID:                strfmt.UUID(config.GlobalAgentConfig.HostID),
 		ClusterID:             strfmt.UUID(config.GlobalAgentConfig.ClusterID),
 		DiscoveryAgentVersion: &config.GlobalAgentConfig.AgentVersion,
+		Reply:                 &reply,
 	}
-	reply := models.StepReply{
-		StepType: stepType,
-		StepID:   stepID,
-		ExitCode: int64(exitCode),
-		Output:   output,
-		Error:    errStr,
-	}
-	params.Reply = &reply
+
 	_, err := s.Client().Installer.PostStepReply(s.Context(), &params)
 	if err != nil {
 		switch errValue := err.(type) {
@@ -75,12 +83,22 @@ func (s *stepSession) sendStepReply(stepType models.StepType, stepID, output, er
 		default:
 			s.Logger().WithError(err).Warn("Error posting step reply")
 		}
-	} else if exitCode == 0 {
-		storeInCache(stepType, output)
+	} else if reply.ExitCode == 0 {
+		storeInCache(reply.StepType, reply.Output)
 	}
 }
 
-func (s *stepSession) handleSingleStep(stepType models.StepType, stepID string, command string, args []string, handler HandlerType) {
+func (s *stepSession) createStepReply(stepType models.StepType, stepID string, output string, errStr string, exitCode int) models.StepReply {
+	return models.StepReply{
+		StepType: stepType,
+		StepID:   stepID,
+		ExitCode: int64(exitCode),
+		Output:   output,
+		Error:    errStr,
+	}
+}
+
+func (s *stepSession) handleSingleStep(stepType models.StepType, stepID string, command string, args []string, handler HandlerType) models.StepReply {
 	s.Logger().Infof("Executing step: <%s>, command: <%s>, args: <%v>", stepID, command, args)
 	stdout, stderr, exitCode := handler(command, args...)
 	if exitCode != 0 {
@@ -92,7 +110,8 @@ stderr:
 %v
 `, exitCode, stepID, command, args, stdout, stderr)
 	}
-	s.sendStepReply(stepType, stepID, stdout, stderr, exitCode)
+
+	return s.createStepReply(stepType, stepID, stdout, stderr, exitCode)
 }
 
 func (s *stepSession) handleSteps(steps *models.Steps) {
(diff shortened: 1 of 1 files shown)


src/commands/step_processor.go
```diff
@@ -74,7 +71,17 @@ func (s *stepSession) sendStepReply(stepType models.StepType, stepID, output, er
 	}
 }
 
-func (s *stepSession) handleSingleStep(stepType models.StepType, stepID string, command string, args []string, handler HandlerType) {
+func (s *stepSession) createStepReplay(stepType models.StepType, stepID string, output string, errStr string, exitCode int) models.StepReply {
```
omertuc, line 74: s/Replay/Reply/

src/commands/step_processor.go
```diff
@@ -86,19 +93,78 @@ stderr:
...
+			}
+
+			s.sendStepReply(reply)
+		}(step)
+	}
+}
+
+func (s *stepSession) diagnoseSystem() error {
+	mediaPath := "/run/media/iso"
+	stdout, stderr, exitCode := util.ExecutePrivileged("findmnt", "--raw", "-n", "-o", "SOURCE,TARGET", "--target", mediaPath)
```
omertuc, line 134: Please use long flag forms in scripts/code, e.g. `--noheadings` vs `-n`

src/commands/step_processor.go
```diff
@@ -86,19 +93,78 @@ stderr:
...
+	}
+
+	fields := strings.Fields(stdout)
+
+	if fields[1] != mediaPath {
+		return s.diagnoseError(errors.Errorf("media isn't mounted directly to %s directory", mediaPath))
+	}
+
+	source := fields[0]
+	if source == "" || !strings.HasPrefix(source, "/dev") {
```
omertuc, line 151: What does the SOURCE column mean exactly? What does it mean for it to have the value `/dev`? 

Is there no direct way to reach that conclusion, without going through `findmnt`?

src/commands/step_processor.go
```diff
@@ -86,19 +93,78 @@ stderr:
...
+				}
+			}
+
+			s.sendStepReply(reply)
+		}(step)
+	}
+}
+
+func (s *stepSession) diagnoseSystem() error {
+	mediaPath := "/run/media/iso"
```
omertuc, line 133: There are many ways for the to boot from ISO, are we sure that this directory will be used in all such cases?

src/commands/step_processor.go
```diff
@@ -86,19 +93,78 @@ stderr:
...
+	if fields[1] != mediaPath {
+		return s.diagnoseError(errors.Errorf("media isn't mounted directly to %s directory", mediaPath))
+	}
+
+	source := fields[0]
+	if source == "" || !strings.HasPrefix(source, "/dev") {
+		return s.diagnoseError(errors.Errorf("invalid mountpoint source %s", source))
+	}
+
+	_, stderr, exitCode = util.ExecutePrivileged("head", "-c", "2", source)
```
omertuc, line 155: Is this simply trying to read 2 bytes from the disk? Why use `head` rather than native Go `ioutil.ReadFile`?

src/commands/step_processor.go
```diff
@@ -86,19 +93,78 @@ stderr:
...
+					reply.Error = err.Error()
+				}
+			}
+
+			s.sendStepReply(reply)
+		}(step)
+	}
+}
+
+func (s *stepSession) diagnoseSystem() error {
```
omertuc, line 132: ```// diagnoseSystem runs quick validations that need to need to occur before and after each step. This is in order to detect and report known problems that might otherwise manifest as confusing error messages.```

src/commands/step_processor.go
```diff
@@ -86,19 +93,78 @@ stderr:
...
-			s.sendStepReply(step.StepType, step.StepID, "", errStr, -1)
+			s.sendStepReply(s.createStepReply(step.StepType, step.StepID, "", errStr, -1))
 			continue
 		}
-		go s.handleSingleStep(step.StepType, step.StepID, step.Command, step.Args, util.ExecutePrivileged)
+
+		go func(step *models.Step) {
+			if err := s.diagnoseSystem(); err != nil {
+				s.sendStepReply(s.createStepReply(step.StepType, step.StepID, "", err.Error(), -1))
+				return
```
omertuc, line 112: We should maybe consider having this as best effort. Run diagnose. Get failures. Try to run the step anyway. Step fails? Append the error from the pre-diagnostics. Step succeeds? Great, our diagnostics were wrong

src/commands/step_processor.go
```diff
@@ -86,19 +93,78 @@ stderr:
...
+			}
+
+			reply := s.handleSingleStep(step.StepType, step.StepID, step.Command, step.Args, util.ExecutePrivileged)
+
+			if err := s.diagnoseSystem(); err != nil {
```
omertuc, line 117: Should post-step-diagnostics even be ran if the step succeeded?

src/commands/step_processor.go
michaellevy101: It's a little bit tricky
When a media is disconnected the CoreOS might have the device file for example: /dev/sr1
and even the mounted files (like /boot) might be exist after accessing them.
To identify the disconnection event we should read from the device itself (not it's mountpoints)
In CoreOs the ISO mounts to the /run/media/iso directory and this code resolves the device file (/dev/sr1) from this mountpoint.  
Here we validating that the the source is a device

src/commands/step_processor.go
michaellevy101: how is it different? (We already used findmnt)

src/commands/step_processor.go
michaellevy101: yes

src/commands/step_processor.go
michaellevy101: If the disk was disconnected the OS still might has a device file for example: /dev/sr1
CoreOS has a directory located at /run/media 
https://support.oracle.com/knowledge/Sun%20Microsystems/1316513_1.html

src/commands/step_processor.go
michaellevy101: Media disconnection might stuck the whole system
https://support.oracle.com/knowledge/Sun%20Microsystems/1316513_1.html

src/commands/step_processor.go
michaellevy101: add a guard

src/commands/step_processor.go
omertuc: Sorry, didn't mean `ReadFile`, I meant `f, err := os.Open(...)` and `f.Read`. Go has native functions to read from files, it seems odd to delegate that to a third-party utility, standard as it may be.

> open a file, check for errors, read two bytes, check for error...

Well that's just what Go code looks like, 
image
Don't think we should do things in odd ways just to avoid those very standard Go `if` statements.

And with regards to `findmnt`, I personally don't think we should have used `findmnt`, I think a loop looking through `ghw` disks mounts would've been better and much more clear, because the behavior of `findmnt` can only be understood by trial and error and reading through a not so clear manpage. But that implementation is a bit more involved, so I didn't comment about that. I don't feel it's relevant to the discussion of whether to use `head` or not.

src/commands/step_processor.go
omertuc: I see. Then can you maybe hide all this logic behind a "getMountpointSourceDeviceFile" function? It would self-document the general goal of this `findmnt` call and would make `diagnoseSystem` a bit less cluttered for future additions

src/commands/step_processor.go
```diff
@@ -86,21 +94,84 @@ stderr:
...
+	_, stderr, exitCode = util.ExecutePrivileged("head", "-c", "2", source)
+
+	if exitCode != 0 {
+		return s.diagnoseError(errors.New(stderr))
+	}
+
+	return nil
+}
+
+func (s *stepSession) diagnoseError(specificError error) error {
```
omertuc, line 169: `s/diagnoseError/getDiagnoseError` or `s/diagnoseError/getDiagnosticError` because `diagnoseError` sounds like a verb+noun (i.e. "This function diagnoses the error")

src/commands/step_processor.go
omertuc: Failing disk, ZFS, zpool - how is this related?

Is a "stuck" system anything we've ever seen? Are you implying we should not run the inventory/dhcp/etc step because that step might cause the machine to freeze? Is that something we've ever seen happen?

It just feels odd that we're blocking all of our steps because we noticed the virtual media looks like it has been disconnected - we should raise a flag somehow, UI warning, an event, even block the installation - but I don't see how actively blocking individual steps because of disconnected media makes sense.

Most steps don't have anything to do with virtual media.

src/commands/step_processor.go
omertuc: I see. The question was due to the same attitude as [my best-effort comment](https://github.com/openshift/assisted-installer-agent/pull/197#discussion_r628890982) - it seems odd we are actively failing and hiding step results because of completely unrelated diagnostics, when the steps themselves were completely.

Maybe this entire diagnostics mechanism should be done separately from the steps flow, it feels completely unrelated. Feels like it should be done in a background thread, and report results to the service API, unrelated to steps.

src/commands/step_processor.go
michaellevy101: after investigating it: We have two cases full-iso and minimal-iso the full-iso contains this directory but the minimal-iso doesn't - implemented

src/commands/step_processor.go
michaellevy101: I don't agree and feel comfortable to use the head function instead of writing 6 lines of code, 
our code is low level and broadly uses os command.
Anyway i addressed your comment and changed the code.

src/commands/step_processor.go
michaellevy101: In the agent, we don't know what's the actual step does, and can't rely on it. (It's impossible to decide which step to stop).
For a minimal-iso, the media disconnection event doesn't stack the server because we read the small ISO to the RAM and don't use the device anymore. 
For the full ISO, the root-fs is a part of the ISO file and after disconnecting it everything might happen.
@ronniel1 told me that media disconnection event is unpredictable and might stack the server.
He saw it in the past and guided me to block the step.

Anyway, I don't understand why to continue if the installation itself will fail for sure(I tested it so many times and it's a common issue). IMO blocking the step is a reasonable action

src/commands/step_processor.go
michaellevy101: That was my first suggestion, but i decided to implement it this way to ensure the system stability before and after each step. (Media disconnection is only one example)

michaellevy101: The ultimate solution is extending the reply object but i decided to delay this decision to the next diagnosis

src/commands/step_processor.go
```diff
@@ -86,21 +96,109 @@ stderr:
...
+	source, err := s.getMountpointSourceDeviceFile()
+
+	if err != nil {
+		log.Warn(err)
+		return nil
+	}
+
+	file, err := os.Open(source)
+	if err != nil {
+		err := errors.Wrap(err, "cannot access the media(ISO) probably media was disconnected")
```
avishayt, line 148:
```suggestion
		err := errors.Wrap(err, "cannot access the media (ISO) - media was likely disconnected")
```

src/commands/step_processor.go
```diff
@@ -86,21 +96,109 @@ stderr:
...
+		log.Error(err)
+		return err
+	}
+
+	defer file.Close()
+
+	_, err = io.ReadFull(file, make([]byte, 2, 2))
+	if err != nil {
+		err := errors.Wrap(err, "cannot access the media(ISO) probably media was disconnected")
```
avishayt, line 157:
```suggestion
		err := errors.Wrap(err, "cannot read from the media (ISO) - media was likely disconnected")
```

(Better to have different error messages - in the future we might learn that they are different cases?)

src/commands/step_processor.go
```diff
@@ -86,21 +96,109 @@ stderr:
...
+		log.Error(err)
+		return err
+	}
+
+	return nil
+}
+
+func (s *stepSession) getMountpointSourceDeviceFile() (string, error) {
```
avishayt, line 165: Nice defensive programming in this method

src/commands/step_processor.go
omertuc: > In the agent, we don't know what's the actual step does, and can't rely on it.

That's the main part of my point - there's no direct relation between steps and ISO disconnection, so I don't see why blocking steps because of disconnection makes sense

> For a minimal-iso, the media disconnection event doesn't stack the server because we read the small ISO to the RAM and don't use the device anymore.

Great to hear that

> For the full ISO, the root-fs is a part of the ISO file and after disconnecting it, everything might happen.
@ronniel1 told me that media disconnection event is unpredictable and might stack the server.

Okay, if that's the case, and we're sure what @ronniel1 said happened, and random steps causing freezes when ISO is disconnected is something we know that happens often, then blocking all steps may be reasonable. I still think it's not the right way (depends on how often it happens), but this is at-least a valid reason.

> Anyway, I don't understand why to continue if the installation itself will fail for sure(I tested it so many times and it's a common issue). IMO blocking the step is a reasonable action

I'm with you on that. I'm definitely against continuing installation if we detect ISO disconnection. I even think that blocking the installation in general, not just the installation step, is the right way to go. I think we should even warn the user about it in the UI.

But I just don't think doing the detection in some kind of blocking pre-step and post-step hook is the right way to detect ISO disconnection / block the installation. The reason I think it's not the right way is because we'll miss a lot of steps that might provide us with important information.

I think we should have a dedicated background thread that periodically performs these `diagnoseSystem` checks (we could have a special step for this instead of a thread but steps don't run often enough), and report to a new service API endpoint which will store the latest result for each host, and add a proper host validation that checks those results.

src/commands/step_processor.go
```diff
@@ -20,6 +26,14 @@ import (
 
 type HandlerType func(string, ...string) (stdout string, stderr string, exitCode int)
 
+type errorCode int
+
+// TODO - add an ErrorCode Enum to the swagger
+const (
+	Undetected        errorCode = 999
```
omertuc, line 33: Do we have any way to tell these apart from regular error codes? Assuming we'll want to add UI support for this error, we'd like to distinguish there errors from regular step errors... So maybe prefix the errors with some "magic" - `0xa1a1a100`, `0xa1a1a101`, `0xa1a1a102` etc

michaellevy101: It's not a common exist code but i changed it to 256 which is out of range 
https://tldp.org/LDP/abs/html/exitcodes.html

PR #198, opened 2021-05-18
MGMT-6484 [WIP] Create cache for post step reply api calls

ori-amizur:
In case a previous result exists in the cache and it is the same as the current result, do not send to the assisted service

/cc @tsorya

### DIFF
--- a/src/commands/reply_cache.go
+++ b/src/commands/reply_cache.go
@@ -1,0 +1,21 @@
+package commands
+
+import (
+	"time"
+
+	"github.com/patrickmn/go-cache"
+
+	"github.com/openshift/assisted-service/models"
+	"github.com/thoas/go-funk"
+)
+
+var replyCache = cache.New(time.Hour, time.Hour)
+
+func alreadyExistsInService(stepType models.StepType, value string) bool {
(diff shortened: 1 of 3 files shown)


src/commands/reply_cache.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+var replyCache sync.Map
+
+func alreadyExistsInService(stepType models.StepType, value string) bool {
+	storedValue, ok := replyCache.Load(stepType)
+	return ok && funk.Equal(storedValue, value)
+}
+
+func storeInCache(stepType models.StepType, value string) {
```
tsorya, line 17: don't we wanted to have cache with ttl?

src/commands/reply_cache.go
```diff
@@ -0,0 +1,19 @@
+package commands
+
+import (
+	"sync"
+
+	"github.com/openshift/assisted-service/models"
+	"github.com/thoas/go-funk"
+)
+
+var replyCache sync.Map
```
ronniel1, line 10: I think we should use go-cache instead and have TTL on the entries

PR #199, opened 2021-05-18
MGMT-6498 Capture routing information in inventory

jordigilh:
* Populates the routing information into the `inventory` structure for both IPv4 and IPv6. 
* The code is a copy & paste from the [source code](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/net/interface.go) of Kubernetes for both the business logic and the test scenarios, but adapted for the agent's use case.
* The PR depends on this [PR](https://github.com/openshift/assisted-service/pull/1752) that adds the Routing fields in the swagger schema.


avishayt (changes requested):
Need to add routes to "inventory" in swagger

src/inventory/routes.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+	return rf.parse(file)
+}
+
+func GetRoutes(dependencies util.IDependencies) []*models.Route {
+
+	routes, err := v4File.extract()
```
ori-amizur, line 48: Is there a reason why not to use existing library such as netlink ?
https://pkg.go.dev/github.com/vishvananda/netlink#RouteList

jordigilh:
> Need to add routes to "inventory" in swagger

It's already covered in this other PR:

https://github.com/openshift/assisted-service/pull/1752

src/inventory/routes.go
jordigilh: None, I was simply not aware of this library. I'll refactor the code and tests accordingly.

src/inventory/routes.go
```diff
@@ -0,0 +1,107 @@
...
+			Interface:   linkName,
+			Destination: r.Dst.IP.String(),
+			Gateway:     r.Gw.String(),
+			Family:      int32(familyIPv4),
+		})
+	}
+	return routes, nil
+}
+
+func getIPv6DefaultRoutes(h handler) ([]*models.Route, error) {
```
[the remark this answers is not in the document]
jordigilh: It would require an if/else to handle the specific use case of Ipv6 where it validates that the GW equals to IPv6zero (for the loopback device):
https://github.com/openshift/assisted-installer-agent/blob/e7883cf8a0d3399742fbeb690f149d14dd74e30f/src/inventory/routes.go#L91-L93
versus IPv4
https://github.com/openshift/assisted-installer-agent/blob/e7883cf8a0d3399742fbeb690f149d14dd74e30f/src/inventory/routes.go#L65-L67
I think it could cause more confusion to merge both functions because of this particular scenario.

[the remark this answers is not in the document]
jordigilh: I'll update the code with the `IsUnspecified()` method. Looks much better :smile:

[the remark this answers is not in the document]
jordigilh: I take it back: yes, we can merge both methods into one. Turns out that the GW for both IPv4 and IPv6 are zero, so we're good :smile:

src/inventory/routes.go
```diff
@@ -0,0 +1,83 @@
...
+		return routes //If ipv6 failed, we still return the ipv4 default route(s) since that could be sufficient.
+	}
+	routes = append(routes, ipv6Routes...)
+	return routes
+}
+
+func getIPDefaultRoutes(h handler) ([]*models.Route, error) {
+	rList, err := h.getRouteList()
+	if err != nil {
+		logrus.Errorf("Unable to retrieve the IPv6 routes: %s", err)
```
ori-amizur, line 62: Please fix the error message

jordigilh: done

PR #211, opened 2021-06-22
MGMT-6498: Fixes empty slice capturing the default route

jordigilh:
While testing the assisted-service [PR](https://github.com/openshift/assisted-service/pull/1785) I noticed that the slice returning from the agent for the default route was always empty. I tested the logic locally using my own routing information and I detected that the value returned by the go library for the destination field is a nil, and in the current filtering logic it is expecting an IP of zero value (0.0.0.0 for IPv4 and equivalent for IPv6). 

This PR addresses this discrepancy by making sure that the filter now only takes into account routing entries where the Dst field is nil, or if it's not nil, then it is unspecified (== IP zero) in order to be considered a default route.

### DIFF
--- a/src/inventory/routes_test.go
+++ b/src/inventory/routes_test.go
@@ -17,29 +17,11 @@ type netPair struct {
 }
 
 var (
-	gwFirst = netPair{
+	ipV4GW = netPair{
 		routes: []netlink.Route{
-			{LinkIndex: 0, Dst: &net.IPNet{IP: net.IPv4zero}, Gw: net.IPv4(10, 254, 0, 1)},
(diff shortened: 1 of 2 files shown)


src/inventory/routes_test.go
```diff
@@ -106,6 +106,13 @@ func (th testHandler) getFamily() int {
 	return th.family
 }
 
+func (th testHandler) getIPZero() *net.IP {
```
ori-amizur, line 109: Remove duplicate function

jordigilh: it's needed to implement the interface contract.

ori-amizur: So make them both call a function with family parameter that does the same thing

jordigilh: ok

ori-amizur:
LGTM

PR #216, opened 2021-07-14
MGMT-7113: [WIP] Added arm64 build to Makefile and Dockerfile

danielerez:
* Makefile: added build-arm64 target for bulding arm64 binaries,
  add a proper postfix and output to build/arm64 folder.
* Dockerfile: Invoked build-arm64 target and copied binaries to /usr/bin

### DIFF
--- a/Dockerfile.assisted_installer_agent
+++ b/Dockerfile.assisted_installer_agent
@@ -10,6 +10,7 @@ RUN go mod download
 COPY . .
 
 RUN make build
+RUN make build-arm64
 
 FROM quay.io/centos/centos:centos8
 RUN dnf install -y \
@@ -43,5 +44,7 @@ COPY --from=builder /go/src/github.com/openshift/assisted-installer-agent/build/
 COPY --from=builder /go/src/github.com/openshift/assisted-installer-agent/build/container_image_availability /usr/bin/container_image_availability
 COPY --from=builder /go/src/github.com/openshift/assisted-installer-agent/build/domain_resolution /usr/bin/domain_resolution
 COPY --from=builder /go/src/github.com/openshift/assisted-installer-agent/build/disk_speed_check /usr/bin/disk_speed_check
+
+COPY --from=builder /go/src/github.com/openshift/assisted-installer-agent/build/arm64/ /usr/bin/
 
 COPY scripts/installer/* /usr/local/bin/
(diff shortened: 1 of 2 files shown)


filanov:
cc: @YuviGold @osherdp

Makefile
```diff
@@ -27,13 +27,16 @@ ci-lint:
...
 
-.PHONY: build clean build-image push subsystem
+.PHONY: build build-arm64 clean build-image push subsystem
 build: build-agent build-connectivity_check build-inventory build-free_addresses build-logs_sender \
 	   build-dhcp_lease_allocate build-apivip_check build-next_step_runner build-ntp_synchronizer \
 	   build-container_image_availability build-domain_resolution build-disk_speed_check
 
 build-%: $(BIN) src/$* #lint
-	CGO_ENABLED=0 go build -o $(BIN)/$* src/$*/main/main.go
+	GOOS=$(GOOS) GOARCH=$(GOARCH) CGO_ENABLED=0 go build -o ${BIN}/$*${FILE_POSTFIX} src/$*/main/main.go
```
YuviGold, line 36: Should we set defaults for `GOOS`, `GOARCH` ? (for x86)

Dockerfile.assisted_installer_agent
```diff
@@ -44,4 +45,6 @@ COPY --from=builder /go/src/github.com/openshift/assisted-installer-agent/build/
 COPY --from=builder /go/src/github.com/openshift/assisted-installer-agent/build/domain_resolution /usr/bin/domain_resolution
 COPY --from=builder /go/src/github.com/openshift/assisted-installer-agent/build/disk_speed_check /usr/bin/disk_speed_check
 
+COPY --from=builder /go/src/github.com/openshift/assisted-installer-agent/build/arm64/ /usr/bin/
```
YuviGold, line 48: So the same image has 2 artifacts? I think we should have a different image for that, no?

Makefile
danielerez: Doesn't seem to be necessary, empty defaults just means using local os and arch (i.e. similar to the current behaviour).

YuviGold (approved):
clean solution! LGTM

danielerez:
Not relevant any more, using a specific image for each architecture instead.

PR #217, opened 2021-07-14
Bug 1980901: Parse GW nil as net zero

jordigilh:
When the GW is returned as `nil` from the agent, it is populated as garbage (`<nil>`) in the json, instead of containing a properly formatted IP. This causes the validation in the service to dump error log messages as it is unable to parse the GW value. 

This fix tackles the parsing part from the agent side, and sends an empty string value when it detects the value of the Gateway is nil.

There is a counterpart PR for the service to handle this use case:

https://github.com/openshift/assisted-service/pull/2223

I also enhanced the unit tests to check that the contents of the processed routes match as expected.

@ori-amizur can you take a look at this PR when you have some time?

### DIFF
--- a/src/inventory/interfaces.go
+++ b/src/inventory/interfaces.go
@@ -56,7 +56,7 @@ func ipWithCidrInCidr(ipWithCidrStr, cidrStr string) bool {
 func analyzeAddress(addr net.Addr) (isIpv4 bool, addrStr string, err error) {
 	ipNet, ok := addr.(*net.IPNet)
 	if !ok {
-		return false, "", fmt.Errorf("Could not cast to *net.IPNet")
+		return false, "", fmt.Errorf("could not cast to *net.IPNet")
 	}
 	mask, _ := ipNet.Mask.Size()
 	addrStr = fmt.Sprintf("%s/%d", ipNet.IP.String(), mask)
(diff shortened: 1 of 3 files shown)


src/inventory/interfaces.go
```diff
@@ -56,7 +56,7 @@ func ipWithCidrInCidr(ipWithCidrStr, cidrStr string) bool {
 func analyzeAddress(addr net.Addr) (isIpv4 bool, addrStr string, err error) {
 	ipNet, ok := addr.(*net.IPNet)
 	if !ok {
-		return false, "", fmt.Errorf("Could not cast to *net.IPNet")
+		return false, "", fmt.Errorf("could not cast to *net.IPNet")
```
ori-amizur, line 59: Is this change necessary?

jordigilh: Yes, it is. Errors should not be capitalized, as stated in the golang [coding review document](https://github.com/golang/go/wiki/CodeReviewComments#error-strings), which is being referred in the kuberentes [coding convention document](https://github.com/kubernetes/community/blob/master/contributors/guide/coding-conventions.md).
You can also find a warning message in your IDE by the go-staticcheck tool:
`error strings should not be capitalized (ST1005)`

jordigilh:
@ori-amizur can you approve if you have no additional comments? Thanks.

PR #225, opened 2021-08-19
OCPRHV-591: Add oVirt support to the agent side

bkhizgiy:
To add oVirt provider support some modifications need to be done on the agent side.
oVirt platform is detected according to the family parameter and not by product name 
which can be various, to detect the right platform and add it to the isVitual list a new 
function was added and it will change the host product type to match oVirt one, it will
ensure that the rest of the flow will remain the same as for the different types of hosts.

### DIFF
--- a/src/inventory/system_vendor.go
+++ b/src/inventory/system_vendor.go
@@ -17,6 +17,7 @@ func isVirtual(product string) bool {
 		"Virtual Machine",
 		"AHV",
 		"HVM domU",
+		"oVirt",
 	} {
 		if strings.Contains(product, vmTech) {
 			return true
@@ -24,6 +25,12 @@ func isVirtual(product string) bool {
 	}
 
 	return false
+}
+
+// For oVirt VMs the correct platform can be detected only by the Family value,
+// this function is used to check that and update the productName accordingly.
+func isOVirtPlatform(family string) bool {
+	return family == "oVirt" || family == "RHV"
 }
 
 func GetVendor(dependencies util.IDependencies) *models.SystemVendor {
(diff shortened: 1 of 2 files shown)


src/inventory/system_vendor.go
```diff
@@ -8,7 +8,11 @@ import (
 
 	"strings"
 )
-
+const (
+	ovirtProductName string = "oVirt"
+	rhvProductName string = "RHV"
+	)
```
avishayt, line 14: Why is this defined here but vmware consts defined in internal/common/platform.go?  Can we keep them consistent?

bkhizgiy: Sure will remove it from here and move it to the same place as VMware, I wasn't aware of the definition in the other file.

bkhizgiy:
@avishayt Can you please review the patch? It's a really small change and I think it's ready to be merged.
all staff related to the service side will be merged after the enhancement doc is approved.

src/inventory/system_vendor.go
```diff
@@ -8,7 +8,7 @@ import (
 
 	"strings"
 )
-
+
```
avishayt, line 11: Remove whitespace

src/inventory/system_vendor.go
```diff
@@ -26,6 +27,11 @@ func isVirtual(product string) bool {
 	return false
 }
 
+// For oVirt VMs the correct platform can be detected only by the Family value, this function is used to check that and update the productName accordingly.
```
avishayt, line 30: Break into 2 lines

src/inventory/system_vendor.go
bkhizgiy: done

src/inventory/system_vendor.go
bkhizgiy: done

rollandf:
What about adding a test here? src/inventory/system_vendor_test.go

bkhizgiy:
> What about adding a test here? src/inventory/system_vendor_test.go

Good Idea, I've added a new test for oVirt product detection and modified the virtual detection one.

avishayt (approved):

PR #227, opened 2021-08-23
MGMT-5400: Get physical_bytes from ghw if dmidecode was failed

eliorerz:
image

### DIFF
--- a/src/inventory/memory.go
+++ b/src/inventory/memory.go
@@ -1,12 +1,11 @@
 package inventory
 
 import (
+	"github.com/openshift/assisted-installer-agent/src/util"
+	"github.com/sirupsen/logrus"
 	"regexp"
 	"strconv"
 	"strings"
-
-	"github.com/openshift/assisted-installer-agent/src/util"
-	"github.com/sirupsen/logrus"
 
 	"github.com/openshift/assisted-service/models"
 )
@@ -40,6 +39,35 @@ func newMemory(dependencies util.IDependencies) *memory {
 }
 
 func (m *memory) getTotalPhysicalBytes() int64 {
+
+	if physicalBytes := m.getDmidecodeTotalPhysicalBytes(); physicalBytes != 0 {
+		return physicalBytes
+	}
+
+	// Try to use ghw to get total physical bytes, as dmidecode returns 0 on some platforms
+	if physicalBytes := m.getGhwTotalPhysicalBytes(); physicalBytes != 0 {
+		return physicalBytes
+	}
+
+	// Treat usable bytes as physical bytes, as ghw returns 0 on some platforms
+	if usableBytes := m.getTotalUsabeBytes(); usableBytes != 0 {
+		return usableBytes
+	}
+
+	return 0
+}
+
+func (m *memory) getGhwTotalPhysicalBytes() int64 {
+	mem, err := m.dependencies.Memory()
+	if err != nil {
+		logrus.Errorf("Error getting memory info: %v", err)
+		return 0
+	}
+
+	return mem.TotalPhysicalBytes
+}
+
+func (m *memory) getDmidecodeTotalPhysicalBytes() int64 {
 	o, e, exitCode := m.dependencies.Execute("dmidecode", "-t", "17")
 	if exitCode != 0 {
 		logrus.Errorf("Could not run dmidecode: %s", e)
--- a/src/inventory/memory_test.go
+++ b/src/inventory/memory_test.go
@@ -2,7 +2,7 @@ package inventory
 
 import (
 	"fmt"
-
+	"github.com/jaypipes/ghw"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/openshift/assisted-installer-agent/src/util"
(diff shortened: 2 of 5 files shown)


src/inventory/memory.go
```diff
@@ -40,6 +40,32 @@ func newMemory(dependencies util.IDependencies) *memory {
 }
 
 func (m *memory) getTotalPhysicalBytes() int64 {
+
+	if physicalBytes := m.getDmidecodeTotalPhysicalBytes(); physicalBytes != 0 {
+		return physicalBytes
+	}
+
+	if physicalBytes := getGhwTotalPhysicalBytes(); physicalBytes != 0 {
```
omertuc, line 48: `// Try to use ghw to get total physical bytes, as `dmidecode` returns 0 on some platforms`

src/inventory/memory.go
```diff
@@ -40,6 +40,32 @@ func newMemory(dependencies util.IDependencies) *memory {
...
+		return physicalBytes
+	}
+
+	if usableBytes := m.getTotalUsabeBytes(); usableBytes != 0 {
```
omertuc, line 52: `// Treat usable bytes as physical bytes, as `ghw` returns 0 on some platforms`

src/inventory/memory.go
```diff
@@ -40,6 +40,32 @@ func newMemory(dependencies util.IDependencies) *memory {
 }
 
 func (m *memory) getTotalPhysicalBytes() int64 {
```
omertuc, line 42: Return the method that was used to get the bytes
`func (m *memory) getTotalPhysicalBytes() (int64, string) {`

src/inventory/memory_test.go
```diff
@@ -95,7 +95,8 @@ var _ = Describe("Memory test", func() {
 		dependencies.On("Execute", "dmidecode", "-t", "17").Return("", "Just an error", -1).Once()
 		dependencies.On("ReadFile", "/proc/meminfo").Return(nil, fmt.Errorf("Another error")).Once()
 		ret := GetMemory(dependencies)
-		Expect(ret).To(Equal(&models.Memory{}))
+		Expect(ret).To(Not(Equal(&models.Memory{})))
+		Expect(ret).To(Equal(&models.Memory{PhysicalBytes: getGhwTotalPhysicalBytes()}))
 	})
```
omertuc, line 100: Add tests to make sure `getTotalPhysicalBytges` correctly does fallbacks

src/inventory/memory.go
eliorerz: That will be in a different PR. Need to add it to swagger first

omertuc: No reason not to have this function return the method now & test it, but ignore it - then in the second PR just populate the swagger

src/inventory/memory.go
```diff
@@ -40,6 +39,35 @@ func newMemory(dependencies util.IDependencies) *memory {
...
+
+	if physicalBytes := m.getDmidecodeTotalPhysicalBytes(); physicalBytes != 0 {
+		// Try to use ghw to get total physical bytes, as dmidecode returns 0 on some platforms
```
omertuc, line 44: The comments are not placed correctly

src/inventory/memory_test.go
```diff
@@ -92,11 +101,23 @@ var _ = Describe("Memory test", func() {
...
+		dependencies.On("ReadFile", "/proc/meminfo").Return(nil, fmt.Errorf("meminfo error")).Twice()
+		dependencies.On("Memory").Return(&mem1, nil).Once()
+
 		ret := GetMemory(dependencies)
 		Expect(ret).To(Equal(&models.Memory{}))
+		Expect(ret.PhysicalBytes).To(Equal(int64(0)))
+		Expect(ret.UsableBytes).To(Equal(int64(0)))
 	})
+	It("Physical fallback to useable", func() {
+		dependencies.On("Execute", "dmidecode", "-t", "17").Return("", "Just an error", -1).Once()
```
omertuc, line 114: What about testing fallback to ghw?

src/inventory/memory.go
eliorerz: @omertuc 
Adding `physical_bytes_method` to swagger.yaml will generate ENUM that can and will be the second return value from `getTotalPhysicalBytes` on the agent.
We need to add it first and then add it in the agent. Returning any other value will cause another PR for replacing it to the new ENUM.

omertuc: Fair enough

src/inventory/memory_test.go
eliorerz: @omertuc Done

src/inventory/memory_test.go
```diff
@@ -92,11 +108,30 @@ var _ = Describe("Memory test", func() {
...
+		dependencies.On("Execute", "dmidecode", "-t", "17").Return("", "dmidecode error", -1).Once()
+		dependencies.On("ReadFile", "/proc/meminfo").Return(nil, fmt.Errorf("meminfo error")).Twice()
+		dependencies.On("Memory").Return(&mem1, nil).Once()
+
 		ret := GetMemory(dependencies)
 		Expect(ret).To(Equal(&models.Memory{}))
+		Expect(ret.PhysicalBytes).To(Equal(int64(0)))
+		Expect(ret.UsableBytes).To(Equal(int64(0)))
+	})
+	It("Physical fallback to ghw", func() {
```
omertuc, line 120: "dmidecode fallback to ghw"

src/inventory/memory_test.go
```diff
@@ -92,11 +108,30 @@ var _ = Describe("Memory test", func() {
...
+		dependencies.On("Execute", "dmidecode", "-t", "17").Return("", "Just an error", -1).Once()
+		dependencies.On("ReadFile", "/proc/meminfo").Return([]byte(meminfoContentskB), nil).Once()
+		dependencies.On("Memory").Return(&mem2, nil).Once()
+		ret := GetMemory(dependencies)
+		Expect(ret.PhysicalBytes).To(Equal(mem2.TotalPhysicalBytes))
 	})
+	It("Physical fallback to useable", func() {
```
omertuc, line 127: "ghw fallback to usable"

PR #232, opened 2021-09-13
MGMT-7457: Adding TPM version to host inventory.

ybettan:
The TPM version is validated in the service to make sure the specific
TPM HW is supported by openshift.


ybettan:
/hold
Need to run test-infra first and see if the inventory is valid.

PR #235, opened 2021-09-17
 MGMT-7847: Don't check verify CIDR in the host

flaper87:
The CIDR validation, the majorityGroup check, and the L2 and L3 validations, are
already performed by assisted service when validating the cluster, host, and network
data. There should not be a need to replicate this validation in the agent itself.

UPDATE
==

This was manually tested by deploying a 3-masters spoke cluster and then a 1 worker (on a separate network). Tests were done by myself and @alknopfler 

image (11)

Signed-off-by: Flavio Percoco <flavio@redhat.com>

### DIFF
--- a/src/apivip_check/apivip_check.go
+++ b/src/apivip_check/apivip_check.go
@@ -3,11 +3,8 @@ package apivip_check
 import (
 	"encoding/json"
 	"io/ioutil"
-	"net"
 	"net/http"
-	"net/url"
 
-	"github.com/openshift/assisted-installer-agent/src/util"
 	"github.com/openshift/assisted-service/models"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -32,15 +29,6 @@ func CheckAPIConnectivity(checkAPIRequestStr string, log logrus.FieldLogger) (st
 		wrapped := errors.Wrap(err, "Failed to download worker.ign file")
 		log.WithError(err).Error(wrapped.Error())
 		return createResponse(false), wrapped.Error(), 0
-	}
-
-	if checkAPIRequest.VerifyCidr {
-		log.Warnf("The VerifyCidr (verify_cidr) is being deprecated. CIDR verification is done by Assisted Service directly")
-		if err := verifyCIDR(*checkAPIRequest.URL, log); err != nil {
-			wrapped := errors.Wrap(err, "CheckAPIConnectivity: failure verifying CIDR of API VIP")
-			log.WithError(err).Error(wrapped.Error())
-			return createResponse(false), wrapped.Error(), 0
-		}
 	}
 
 	return createResponse(true), "", 0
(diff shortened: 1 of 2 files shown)


flaper87:
@danielerez according to `git blame`, you added this code a while back. I'd love to get some feedback from you on why it was needed, and whether you think these same checks are indeed covered by assisted-service.

mkowalski:
+1 from me. Without knowing the rationale behind introducing this check directly in the agent it looks to me like we are duplicating code of the service, as there is the same check in there (i.e. going over agent's interfaces and matching them with the configured/calculated CIDR)

src/apivip_check/apivip_check.go
```diff
@@ -38,14 +35,6 @@ func CheckAPIConnectivity(checkAPIRequestStr string, log logrus.FieldLogger) (st
 		return createResponse(false), wrapped.Error(), 0
 	}
 
-	if checkAPIRequest.VerifyCidr {
-		if err := verifyCIDR(*checkAPIRequest.URL, log); err != nil {
```
omertuc, line 42: Replace this with an error log, rather than removing it altogether

romfreiman:
@danielerez  @yevgeny-shnaidman

danielerez:
> @danielerez according to `git blame`, you added this code a while back. I'd love to get some feedback from you on why it was needed, and whether you think these same checks are indeed covered by assisted-service.

So this check was added for adding an host in day2 flow, if support_l2 is enabled: https://github.com/openshift/assisted-service/blob/5c005bce5b74d4ca9000329bc466a4360e8980aa/internal/host/hostcommands/instruction_manager.go#L76
IINM, the validation in the service is still only for day1 flow, i.e. only when installing a new cluster.
Thus, I think this validation is probably still needed for the day2 flow, so if we want to remove it, need to ensure it's indeed covered by the service.

flaper87:
> Thus, I think this validation is probably still needed for the day2 flow, so if we want to remove it, need to ensure it's indeed covered by the service.

This check is really only used for day1. In day to the CIDR validation is disabled (which is what `SUPPORT_L2` is for). The day1 checks for CIDR are all done by the service, whether we use vpi-dhcp or not.

The agent check is not doing anything that is not covered by the service. If this was the only use-csae, I think we are good to remove it.

src/apivip_check/apivip_check.go
flaper87: The verification happens at the service level. Switching this to an error will make it fail for all deployments wen `SUPPORT_L2=True`

omertuc: I didn't mean literal error, just error log. If the service passing this VerifyCidr flag and the flag does nothing, I prefer we have some indication the service is doing "nonsensical" things and that we should fix the service to stop doing those nonsensical things

flaper87: The flag will be removed, there's a PR for it already: https://github.com/openshift/assisted-service/pull/2650

omertuc: I understand, but it's still a parameter [defined](https://github.com/openshift/assisted-service/blob/d8888171f88fd2ab69920e2088f742fd35fa92ee/swagger.yaml#L7592-L7594) in the swagger and it's still part of our service-agent API, and I prefer to have an error/warning log when that parameter is set to true rather than silently ignoring it, in case someone tries to use it in the future for whatever reason

flaper87: Oh, I now see where you are coming from. Considering this validation as an API, and the inputs it receives (as defined in `swagger`).

with that in mind, I think we may want to hold off with removing this functionality until we completely remove the endpoint from the service. What do you think?

omertuc:
So set this PR as draft for now and re-open it in the future?

romfreiman:
Did we test it?

PR #236, opened 2021-09-20
MGMT-7847: Add a deprecation warning for the VerifyCidr option

flaper87:
Signed-off-by: Flavio Percoco <flavio@redhat.com>

### DIFF
--- a/src/apivip_check/apivip_check.go
+++ b/src/apivip_check/apivip_check.go
@@ -39,6 +39,7 @@ func CheckAPIConnectivity(checkAPIRequestStr string, log logrus.FieldLogger) (st
 	}
 
 	if checkAPIRequest.VerifyCidr {
+		log.Warnf("The VerifyCidr (verify_cidr) is being deprecated. CIDR verification is done by Assisted Service directly")
 		if err := verifyCIDR(*checkAPIRequest.URL, log); err != nil {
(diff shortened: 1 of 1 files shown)


src/apivip_check/apivip_check.go
```diff
@@ -39,6 +39,7 @@ func CheckAPIConnectivity(checkAPIRequestStr string, log logrus.FieldLogger) (st
 	}
 
 	if checkAPIRequest.VerifyCidr {
+		log.Warnf("The VerifyCidr ()`verify_cidr`) is being deprecated. CIDR verification is done by Assisted Service directly")
```
omertuc, line 42: uneven ()

flaper87: :facepalm:

src/apivip_check/apivip_check.go
```diff
@@ -39,6 +39,7 @@ func CheckAPIConnectivity(checkAPIRequestStr string, log logrus.FieldLogger) (st
 	}
 
 	if checkAPIRequest.VerifyCidr {
+		log.Warnf("The VerifyCidr (`verify_cidr`) is being deprecated. CIDR verification is done by Assisted Service directly")
```
mkowalski, line 42: I'm going to LGTM anyway, but not sure if backtick (`\``) printed to the terminal and/or log file is not crashing something

flaper87: Let me remove it, we don't want things to crash

PR #242, opened 2021-10-07
NO-ISSUE: Removing \n suffix from TPM version.

ybettan:
It will make the agent send an invalid TPM version format to the service
and, therefore, to fail.


eranco74 (commented):
/lgtm

PR #245, opened 2021-10-12
OCPBUGSM-34971: container image availability command should not exit 0 if it failed on one of the images

tsorya:
Image availability command should exit failure in case images are not available. It is done to not cache failed answers and allow retry

### DIFF
--- a/src/container_image_availability/container_image_availability.go
+++ b/src/container_image_availability/container_image_availability.go
@@ -24,6 +24,7 @@ const (
 	templatePull     = "podman pull %s"
 	templateGetImage = "podman images --quiet %s"
 	templateTimeout  = "timeout %s %s"
+	failedToPullImageExitCode = 2
 )
 
 //go:generate mockery -name ImageAvailabilityDependencies -inpkg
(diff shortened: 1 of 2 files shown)


src/container_image_availability/container_image_availability.go
```diff
@@ -141,13 +142,17 @@ func Run(requestStr string, executer ImageAvailabilityDependencies, log logrus.F
 
 	finishOnTimeout := time.Now().Add(time.Duration(request.Timeout) * time.Second)
 	for _, image := range request.Images {
-		response.Images = append(response.Images, handleImageAvailability(executer, log, int64(time.Until(finishOnTimeout).Seconds()), image))
+		imageResponse := handleImageAvailability(executer, log, int64(time.Until(finishOnTimeout).Seconds()), image)
+		response.Images = append(response.Images, imageResponse)
+		if imageResponse.Result != models.ContainerImageAvailabilityResultSuccess {
+			exitCode = 2
```
osherdp, line 148: 2 as return code has a special meaning apart from "failure"?
https://tldp.org/LDP/abs/html/exitcodes.html

Maybe also save it as a constant to make it more obvious

tsorya: cool, it should be 1 in that case. Can be const too

osherdp (approved):

PR #254, opened 2021-11-07
MGMT-7941: Fix agent dry reboot condition in wrong code path

omertuc:
The agent binary only checked if a reboot happened when next_step_runner has
an error. This is not the case when reboots happen, next_step_runner exits
cleanly with a 0 exit code, so we should only check dry reboot in the agent
if next_step_runner had no error

### DIFF
--- a/src/agent/main/main.go
+++ b/src/agent/main/main.go
@@ -32,17 +32,17 @@ func main() {
 				reRegistrerDelay = defaultRetryDelay
 			}
 
-			if config.GlobalDryRunConfig.DryRunEnabled {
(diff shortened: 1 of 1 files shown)


src/commands/start_agent.go
```diff
@@ -7,7 +7,6 @@ import (
 )
 
 func StartStepRunner(command string, args ...string) error {
-
```
flaper87, line 10: it may be worth reverting this change as it's not relevant to this PR

flaper87 (changes requested):
Just a tiny change request. The rest seems fine

flaper87 (approved):
:pizza:

PR #256, opened 2021-11-18
MGMT-7941: Allow forcing the IP address and hostname of the agent in dry mode

omertuc:
Dry run mode was only tried so far with single-node. This commit makes
the IP and hostname configurable, allowing the swarm to launch
multi-node clusters without having hosts clash.

It also disables the connectivity check in dry run mode because with
different fake IP addresses that would obviously not work

### DIFF
--- a/src/config/dry_run_config.go
+++ b/src/config/dry_run_config.go
@@ -12,7 +12,9 @@ import (
 type DryRunConfig struct {
 	DryRunEnabled        bool   `envconfig:"DRY_ENABLE"`
 	ForcedHostID         string `envconfig:"DRY_HOST_ID"`
-	ForcedMacAddress     string `envconfig:"DRY_MAC_ADDRESS"`
+	ForcedHostIPv4       string `envconfig:"DRY_FORCED_HOST_IPV4"`
+	ForcedMacAddress     string `envconfig:"DRY_FORCED_MAC_ADDRESS"`
+	ForcedHostname       string `envconfig:"DRY_FORCED_HOSTNAME"`
 	FakeRebootMarkerPath string `envconfig:"DRY_FAKE_REBOOT_MARKER_PATH"`
 }
 
@@ -21,7 +23,9 @@ var GlobalDryRunConfig DryRunConfig
 var DefaultDryRunConfig DryRunConfig = DryRunConfig{
 	DryRunEnabled:        false,
 	ForcedHostID:         "",
+	ForcedHostIPv4:       "",
 	ForcedMacAddress:     "",
+	ForcedHostname:       "",
 	FakeRebootMarkerPath: "",
 }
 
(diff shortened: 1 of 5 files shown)


src/config/dry_run_config.go
```diff
@@ -12,7 +12,9 @@ import (
 type DryRunConfig struct {
 	DryRunEnabled        bool   `envconfig:"DRY_ENABLE"`
 	ForcedHostID         string `envconfig:"DRY_HOST_ID"`
+	ForcedHostIPv4       string `envconfig:"DRY_FORCED_HOST_IPV4"`
 	ForcedMacAddress     string `envconfig:"DRY_MAC_ADDRESS"`
+	ForcedHostname       string `envconfig:"DRY_FORCED_HOSTNAME"`
 	FakeRebootMarkerPath string `envconfig:"DRY_FAKE_REBOOT_MARKER_PATH"`
```
osherdp, line 18: I would give it the same name structure, e.g. either ``DRY_<CONFIGURATION>`` or ``DRY_FORCED_<CONFIGURATION>``
Unless maybe I miss something about the differences between some of those values

src/inventory/dry_inventory.go
```diff
@@ -0,0 +1,40 @@
...
+	if config.GlobalDryRunConfig.DryRunEnabled {
+		targetInterface, err := findRelevantInterface(inventory)
+		if err != nil {
+			return
+		}
+
+		// Override the first mac address to the user requested one
+		inventory.Interfaces[targetInterface].MacAddress = config.GlobalDryRunConfig.ForcedMacAddress
+
+		if err == nil {
```
osherdp, line 20: at this stage ``err`` is definitely ``nil``, no?

src/inventory/dry_inventory.go
```diff
@@ -0,0 +1,40 @@
+package inventory
+
+import (
+	"fmt"
+
+	"github.com/openshift/assisted-installer-agent/src/config"
+	"github.com/openshift/assisted-service/models"
+)
+
+func maybeApplyDryRunConfig(inventory *models.Inventory) {
```
osherdp, line 10: name is a bit weird
can we return to the previous way that it was guarded with the ``if``?

osherdp (changes requested):

src/config/dry_run_config.go
omertuc: Are you talking about `FakeRebootMarkerPath` ? It's read as "((Fake-reboot)-marker)-path" as in "the path of the fake-reboot marker", it's not forced or anything.

omertuc: The description on line 44 makes it clearer "A path whose existence indicates a fake reboot happened"

omertuc: Also it's not part of this PR, and that same variable name is also used by the installer/controller so changing it would require another PR there

src/inventory/dry_inventory.go
omertuc: Yeah leftover junk

src/config/dry_run_config.go
osherdp: no, I mean the difference between ``DRY_MAC_ADDRESS`` and ``DRY_FORCED_HOSTNAME``
that both have the structure ``Forced<name>`` in their field name

omertuc: Ah yeah I'll rename `DRY_MAC_ADDRESS` to `DRY_FORCED_MAC_ADDRESS`

osherdp (approved):

omertuc:
forced push again import order

PR #259, opened 2021-12-01
OCPBUGSM-37106: List only physical disks on inventory.

eliorerz:
List only physical disks on inventory. Ignore LVM and other volumes.

### DIFF
--- a/src/inventory/disks.go
+++ b/src/inventory/disks.go
@@ -179,6 +179,13 @@ func checkEligibility(disk *ghw.Disk) (notEligibleReasons []string, isInstallati
 	return notEligibleReasons, isInstallationMedia
 }
 
+func IsPhysicalDisk(disk *block.Disk) bool {
+	return !(strings.HasPrefix(disk.Name, "dm-") || // Kernel device-mapper block devices, technology used by LVM/Docker
+		strings.HasPrefix(disk.Name, "loop") || // Loop devices (see `man loop`)
+		strings.HasPrefix(disk.Name, "zram") || // Default name usually assigned to "swap on ZRAM" block devices
+		strings.HasPrefix(disk.Name, "md")) // Linux multiple-device-driver block devices
+}
+
 func (d *disks) getDisks() []*models.Disk {
 	ret := make([]*models.Disk, 0)
 	var blockInfo *ghw.BlockInfo
@@ -198,6 +205,11 @@ func (d *disks) getDisks() []*models.Disk {
 	for diskIndex, disk := range blockInfo.Disks {
 		var eligibility models.DiskInstallationEligibility
 		var isInstallationMedia bool
+
+		// Ignore non physical disks
+		if !IsPhysicalDisk(disk) {
+			continue
+		}
 
 		eligibility.NotEligibleReasons, isInstallationMedia = checkEligibility(disk)
 		eligibility.Eligible = len(eligibility.NotEligibleReasons) == 0
--- a/src/inventory/disks_test.go
+++ b/src/inventory/disks_test.go
@@ -151,36 +151,26 @@ func createExpectedSDBModelDisk() *models.Disk {
 SDA disk is real disk data from a bare metal machine.
 */
 func createSDADisk() *ghw.Disk {
-	return &ghw.Disk{
-		Name:      "sda",
-		SizeBytes: 999653638144,
-		DriveType: ghw.DRIVE_TYPE_HDD,
-		// run $cat  /sys/block/diskName/dev and then cat /run/udev/data/b{previous command output}
-		BusPath:                "pci-0000:02:00.0-scsi-0:2:0:0",
-		Vendor:                 "DELL",
-		Model:                  "PERC_H330_Mini",
-		SerialNumber:           "6141877064533b0020adf3bb03167694",
-		WWN:                    "0x6141877064533b0020adf3bb03167694",
-		IsRemovable:            false,
-		NUMANodeID:             0,
-		PhysicalBlockSizeBytes: 512,
-		StorageController:      ghw.STORAGE_CONTROLLER_SCSI,
-	}
+	return createDisk("sda", 0, "6141877064533b0020adf3bb03167694", "0x6141877064533b0020adf3bb03167694")
 }
 
 /**
 SDB disk is real disk data from a bare metal machine.
 */
 func createSDBDisk() *ghw.Disk {
+	return createDisk("sdb", 1, "6141877064533b0020adf3bc0325d664", "0x6141877064533b0020adf3bc0325d664")
+}
+
+func createDisk(name string, index int, serialNumber string, wwn string) *ghw.Disk {
 	return &ghw.Disk{
-		Name:                   "sdb",
+		Name:                   name,
 		SizeBytes:              999653638144,
 		DriveType:              ghw.DRIVE_TYPE_HDD,
-		BusPath:                "pci-0000:02:00.0-scsi-0:2:1:0",
+		BusPath:                fmt.Sprintf("pci-0000:02:00.0-scsi-0:2:%d:0", index),
 		Vendor:                 "DELL",
 		Model:                  "PERC_H330_Mini",
-		SerialNumber:           "6141877064533b0020adf3bc0325d664",
-		WWN:                    "0x6141877064533b0020adf3bc0325d664",
+		SerialNumber:           serialNumber,
+		WWN:                    wwn,
 		IsRemovable:            false,
 		NUMANodeID:             0,
 		PhysicalBlockSizeBytes: 512,
@@ -323,6 +313,10 @@ func mockAllForSuccess(dependencies *util.MockIDependencies, disks ...*ghw.Disk)
 		name := disk.Name
 		busPath := disk.BusPath
 
+		if !IsPhysicalDisk(disk) {
+			continue
+		}
+
 		mockGetPathFromDev(dependencies, name, "")
 		mockGetByPath(dependencies, busPath, "")
 		mockGetHctl(dependencies, name, "")
@@ -382,6 +376,19 @@ var _ = Describe("Disks test", func() {
 		mockFetchDisks(dependencies, nil)
 		ret := GetDisks(dependencies)
 		Expect(ret).To(Equal([]*models.Disk{}))
+	})
+
+	It("Invalid disks", func() {
+		dmDisk := createDisk("dm-2", 3, "6141877064533b0020adf3bc0325d664", "0x6141877064533b0020adf3bc0325d664")
+		zramDisk := createDisk("zram0", 4, "6141877064533b0020adf3bc0325d665", "0x6141877064533b0020adf3bc0325d665")
+		mdDisk := createDisk("md-1", 5, "6141877064533b0020adf3bc0325d667", "0x6141877064533b0020adf3bc0325d667")
+		loopDisk := createDisk("loop5", 6, "6141877064533b0020adf3bc0325d668", "0x6141877064533b0020adf3bc0325d668")
+
+		mockReadDir(dependencies, "/dev/disk/by-id", "fetching the by-id disk failed")
+		mockAllForSuccess(dependencies, createSDADisk(), createSDBDisk(), dmDisk, zramDisk, mdDisk, loopDisk)
+
+		ret := GetDisks(dependencies)
+		Expect(ret).Should(HaveLen(2))
 	})
 
 	Describe("Single disk", func() {


src/inventory/disks.go
```diff
@@ -179,6 +179,29 @@ func checkEligibility(disk *ghw.Disk) (notEligibleReasons []string, isInstallati
 	return notEligibleReasons, isInstallationMedia
 }
 
+func (d *disks) getPhysicalDisks() map[string]string {
+	stdout, stderr, exitCode := d.dependencies.Execute("lsblk", "-o", "TYPE,KNAME")
```
osherdp, line 183: Why won't we use information gathered by ghw here instead?
It already chroot to the location /host/ which has the more relevant data about the host (instead of risking getting data from the container instead)
Also less parsing is needed

src/inventory/disks.go
```diff
@@ -179,6 +179,29 @@ func checkEligibility(disk *ghw.Disk) (notEligibleReasons []string, isInstallati
...
+	if exitCode != 0 {
+		logrus.Warnf("Could not get physical disks information: %s", stderr)
+		return nil
+	}
+
+	lines := strings.Split(stdout, "\n")
+	physicalDisks := map[string]string{}
+	for lineno, line := range lines {
+		if lineno > 0 {
```
tsorya, line 192: you can use "lsblk -n -o TYPE,KNAME" and then there will be no need for  "if lineno > 0"

src/inventory/disks.go
```diff
@@ -194,11 +217,16 @@ func (d *disks) getDisks() []*models.Disk {
 	}
 
 	diskPath2diskWWN := d.getDisksWWNs()
+	physicalDisks := d.getPhysicalDisks()
```
osherdp, line 220: Just my phone or the indentation here looks weird?

src/inventory/disks.go
```diff
@@ -179,6 +179,29 @@ func checkEligibility(disk *ghw.Disk) (notEligibleReasons []string, isInstallati
...
+	}
+
+	lines := strings.Split(stdout, "\n")
+	physicalDisks := map[string]string{}
+	for lineno, line := range lines {
+		if lineno > 0 {
+			data := strings.Split(line, " ")
+			volumeType := data[0]
+			volumeName := data[len(data)-1]
+			if volumeType == "disk" {
```
tsorya, line 196: i am not sure how removable disks will appear here, maybe we should check lvm and ignore by it?

src/inventory/disks.go
```diff
@@ -179,6 +179,29 @@ func checkEligibility(disk *ghw.Disk) (notEligibleReasons []string, isInstallati
 	return notEligibleReasons, isInstallationMedia
 }
 
+func (d *disks) getPhysicalDisks() map[string]string {
+	stdout, stderr, exitCode := d.dependencies.Execute("lsblk", "-o", "TYPE,KNAME")
+	if exitCode != 0 {
+		logrus.Warnf("Could not get physical disks information: %s", stderr)
+		return nil
```
osherdp, line 186: nil allows for the "is key in map" check below?
Worth also adding a UT for this case

osherdp: Or rather not allow failures

osherdp (changes requested):
Just those couple of comments, and add UTs for some cases

src/inventory/disks.go
omertuc: Just your phone. GitHub is only usable on mobile when the browser is in Desktop mode

src/inventory/disks.go
omertuc: I agree with Osher, if it's possible to use ghw here we better do that.

But if we're doing lsblk, let's use `lsblk --json --output TYPE,KNAME` for more structured parsing

src/inventory/disks.go
```diff
@@ -179,6 +179,29 @@ func checkEligibility(disk *ghw.Disk) (notEligibleReasons []string, isInstallati
...
+
+	lines := strings.Split(stdout, "\n")
+	physicalDisks := map[string]string{}
+	for lineno, line := range lines {
+		if lineno > 0 {
+			data := strings.Split(line, " ")
+			volumeType := data[0]
+			volumeName := data[len(data)-1]
+			if volumeType == "disk" {
+				physicalDisks[volumeName] = volumeType
```
omertuc, line 197: physicalDisks is being used as a Set structure so please just give it a "true/false" boolean value (which would always be true) rather than a "volumeType" value

src/inventory/disks.go
omertuc: >maybe we should check lvm and ignore by it?
This will not suffice. I device-mapper devices appearing with the "crypt" type by lsblk

omertuc: Looking at lsblk source code maybe even checking for "disk" won't be good enough, for example it looks like RAID arrays (which should technically be fine for us to install on) may appear as `raid` instead of `disk`:

https://github.com/util-linux/util-linux/blob/4939964c9d1977522de1a3de03ebb3011ca866de/lib/blkdev.c#L355-L356

omertuc: From my local testing with ramdisks, swap, LVM, encrypted partitions I find this method quite reliable:

```
if strings.HasPrefix(disk.Name, "dm-") ||
strings.HasPrefix(disk.Name, "loop") ||
strings.HasPrefix(disk.Name, "md") {
	// Ignore this disk!
}
```

It's pretty much identical to [what](https://github.com/util-linux/util-linux/blob/4939964c9d1977522de1a3de03ebb3011ca866de/misc-utils/lsblk.c#L413-L455) `lsblk` is doing behind the scenes when determining the `type`, except without the extra problematic transformations it does with `[blkdev_scsi_type_to_name](https://github.com/util-linux/util-linux/blob/4939964c9d1977522de1a3de03ebb3011ca866de/lib/blkdev.c#L332)` in the end  

I noticed that it doesn't detect the default swap blocked device used by Fedora with [SwapOnZRAM](https://fedoraproject.org/wiki/Changes/SwapOnZRAM)

Even `lsblk` shows that device as "disk" even though it's definitely a virtual block device, and the `HasPrefix` comparisons above won't detect it either. The issue is that such block devices can be created with any name - you can try it yourself with `mknod -m 660 /dev/blabla b 1 0`

One particular `lsblk` column that stands out in being especially good at detecting "real" disks is the `subsystems` column - it seems that any block device is not backed up by real physical connections will have the value `block` in `subsystems`

omertuc: It seems like `ghw`'s `BusPath` being equal to `unknown` is a pretty good indication that it's not a true physical disk, this `BusPath` `ghw` field, taken from udev's `ID_PATH` is pretty similar to the `subsystems` column calculated by `lsblk`. I think we should just look at it and filter any disks that don't have it.

src/inventory/disks.go
omertuc: >Although I don't think ghw has the information we're looking for

Or maybe it does, see [this](https://github.com/openshift/assisted-installer-agent/pull/259/files#r760629845) comment

eliorerz: ghw doesn't give any information about LVK volumes

src/inventory/disks.go
eliorerz: There are other types we want to ignore. Currently we supports only physical volumes so I thought it will be better not to ignore every unwanted type.
Also, removable disk type is `disk` and will not be filtered here

src/inventory/disks.go
eliorerz: I think it's just your phone :)

src/inventory/disks.go
eliorerz: > It appears ghw's BusPath being equal to unknown is a pretty good indication that it's not a true physical disk

Ghw unknown fields are not fully indicates that this volume is physical or not (at least not explicitly).

> From my local testing with ramdisks, swap, LVM and encrypted partitions I find this method quite reliable

Same here, the prefix of the device name is not good enough for deciding if it physical of not

omertuc: >Ghw unknown fields are not fully indicates that this volume is physical or not (at least not explicitly).

Do you have an example where a physical disk has a ghw BusPath value that is `"unknown"`?

omertuc: >Same here, the prefix of the device name is not good enough for deciding if it physical of not

It's literally what `lsblk` is doing as I've [linked](https://github.com/util-linux/util-linux/blob/4939964c9d1977522de1a3de03ebb3011ca866de/misc-utils/lsblk.c#L413-L455) in the message above. Except `lsblk` might call RAID disks "raid" rather than "disk" which will cause you to wrongfully filter them out, as mentioned

src/inventory/disks.go
omertuc: If the `BusPath` is good indication that a device is virtual (`"unknown"` bus types indicates it's not backed up by physical connections) then `ghw` might not give us information about LVK volumes specifically - but it gives us something even more general and useful which is ignoring all non physical disks

omertuc: Also how do you overcome the `/host` problem @osherdp mentioned?

src/inventory/disks.go
eliorerz: > Do you have an example where a physical disk has a ghw BusPath value that is "unknown"?

No, But do you have a list of all the volumes and ghw BusPath is not "unknown"? :) 
Although you might be right, I think for proving that you need to check all the options for all physical disks and configurations. It's much easier to get the physical disks list then assume that ghw BusPath set to "unknown" is what we wanted (unless it was specified that "unknown"  specify unphysical disk) .

>Same here, the prefix of the device name is not good enough for deciding if it physical of not
It's literally what lsblk is doing as I've linked in the message above. Except lsblk might call RAID disks "raid" rather than "disk" which will cause you to wrongfully filter them out, as mentioned

That a valid case.  We might want to specify the specific physical disk types that we currently supporting instead of filtering by BusPath.

omertuc: > No, But do you have a list of all the volumes and ghw BusPath is not "unknown"? :)
No I do not. I tried looking on my local machine and some old inventories I had saved and it seemed pretty reliable.

It also makes sense - the bus path tells you how the disk is physically connected, a disk that exists only virtually isn't physically connected in any way. 

Choosing to look at `lsblk` `type` output is also "guesswork" where we have no guarantees it will work 100% of the time, except I've provided examples (SwapOnZRAM) where the `lsblk` `type` method falls short and causes you to report fake disks. 

> That a valid case. We might want to specify the specific physical disk types that we currently supporting instead of filtering by BusPath.

Or just use the prefix comparisons that I provided which are identical to what lsblk is doing before it starts to potentially split the 'disk' type into other more specific types. If what `lsblk` is doing to calculate the `type` is so simple, why even use `lsblk`? It complicates things, requires parsing, requires us to compare to our already-existing list of disks which we got with a `chroot` that is necessary.

While we could instead just do 3 HasPrefix calls.

In summary there are 3 solutions here:
1) lsblk
2) hasprefix
3) buspath

I don't see how (1) along with:

> We might want to specify the specific physical disk types that we currently supporting 

is in any way better than (2).

and regardless of (2) being simpler and more robust than (1), I think we should just go with (3) because I couldn't find an example where it's wrong while I could find an example where (1) and (2) are wrong

eliorerz: How's `hasPrefix` is good? Are we familiar with all the prefixes? You need to maintain list of prefixes? Are you sure we know them all? You cannot conclude about the all from few test cases.

Regarding 3 (bustype), I agree that this a good approach, but I'm still feeling that this might be more open to bugs than lsblk. We can combine 1 and 3 together to remove some unwanted disks (such as zram)

eliorerz: If lsblk is a big no no from your perspective I believe we can use bustype as an indication for physical (although I'm not completely comfortable with it)

omertuc: > How's hasPrefix is good? 

Because it's exactly what `lsblk` is doing anyway. `lsblk` is simply comparing those 3 prefixes before saying the type is `disk`, then it goes on to split `disk` into even more types which is useless and complicated things further from our perspective. You can go look at the source code, I've linked it multiple times already.

> Are we familiar with all the prefixes? You need to maintain list of prefixes? Are you sure we know them all? 

When you "use lsblk" you're just delegating those responsibilities to `lsblk`. Are the `lsblk` maintainers familiar with all the prefixes? Do they maintain a list of prefixes? Maybe, maybe not. Maybe it's a good idea delegating that responsibility to `lsblk`. But that delegation has a cost - it's putting too much complexity between us and what we want to achieve (know if disk is virtual or not) into a program with no guaranteed stable interface:
How do we know they won't completely change how they determine the `type`, how do we know they don't introduce even more granular types a year from now? Who decided that `type == "disk"` is even the right thing to do in order to determine if a disk is virtual or not? I've even shown it's **not** the right thing in case on SwapOnZRAM 

Comparing these 3 prefixes is simply saying "Let's ignore device-mapper block devices, loop devices, and multiple-devices block devices", it's much more clear and stable then delegating to `lsblk` which have no guaranteed stability, already returns types that are too granular (raid vs disk), and might even become more granular in the future outside of our control, and is a black box until you open up the source code and try to follow what it's actually doing to determine the `type`.

> You cannot conclude about the all from few test cases.

You also cannot conclude that looking at `lsblk` `type` is the right thing to do from a few test cases 

> that this might be more open to bugs than lsblk

I don't understand how BusPath is more open to bugs than `lsblk` `type` in any way.

> If lsblk is a big no no from your perspective I believe we can use bustype as an indication for physical (although I'm not completely comfortable with it)

`lsblk` is a big no until you explain / address the [chroot problem](https://github.com/openshift/assisted-installer-agent/pull/259#discussion_r760970629). Maybe it's an issue, maybe it's not, but it needs to be addressed and explained why it's not an issue.

Once that's addressed it's possible to consider it, but given all I've said about it, I don't understand what benefits it has that the `HasPrefix` method doesn't have and I don't understand what downsides other methods have that the `lsblk` `type` method doesn't have.

eliorerz: >lsblk is simply comparing those 3 prefixes before saying the type is disk 
....
>When you "use lsblk" you're just delegating those responsibilities to lsblk. Are the lsblk maintainers familiar with all the prefixes? Do they maintain a list of prefixes? Maybe, maybe not. Maybe it's a good idea delegating that responsibility to lsblk.


As you mentioned, lsblk is comparing the prefixes to decide whether it's a physical disk or not so I believe they are familiar with those prefixes. Also, I believe that lsblk are more common used that assisted-service (for now) and more likely that for every change in the prefixes the fix will apply before we even noticed that.

src/inventory/disks.go
eliorerz: lsblk is working the same inside the container and on the host itself

src/inventory/disks.go
omertuc: >  Also, I believe that lsblk are more common used that assisted-service (for now) and more likely that for every change in the prefixes the fix will apply before we even noticed that.

I would completely agree with you here and with that approach *if* `lsblk` had a boolean "is_real_disk" column.

But the `type` column doesn't have that meaning, it's just some column that happens to *currently* fit the distinction you're looking for under *some* circumstances when its value equals `disk`. This coincidental behavior is not stable and I don't think relying on it is a good idea. The `lsblk` maintainers give no guarantees here and are completely free to change this behavior in ways that can easily break our flow

src/inventory/disks.go
omertuc: I assume it works because of this mount https://github.com/openshift/assisted-service/blob/e125b59fd271dbeace653e02631a03d8d57ba98f/internal/host/hostcommands/inventory_cmd.go#L37 and some other default mounts 

But we've recently made this `chroot` configurable due to some new projects using the agent code, using `lsblk` without going through ghw would mean that this configuration is ignored, which is not ideal (although I don't think it'll break any current real world use cases)

src/inventory/disks.go
eliorerz: OK, So in that case, what is your suggestion? Use `BusPath` or `HasPrefix` instead ?

omertuc: I say let's do `HasPrefix` for now because it's very unlikely to get rid of disks that are actually real.

But we should open a ticket to look into the reliability of `BusPath`, because if we can't find counter-examples for where it doesn't work, it's a really simple and nice solution that makes sense and doesn't require maintaining a disallow-list of "fake disk technologies" (loop, dm, md).

eliorerz: OK, I'll use `HasPrefix` to filter the unphysical disks.
Note that I'm also adding zram to the filtered disks so it will not be listed in the final disks list.
And on other subject, I noticed that ghw is not always identifying if device is removable or nor (I reproduced both cases in my personal laptop), we might need to open ticket for that as well.

src/inventory/disks.go
```diff
@@ -179,6 +179,13 @@ func checkEligibility(disk *ghw.Disk) (notEligibleReasons []string, isInstallati
 	return notEligibleReasons, isInstallationMedia
 }
 
+func IsPhysicalDisk(disk *block.Disk) bool {
+	return !(strings.HasPrefix(disk.Name, "dm-") ||
```
omertuc, line 183: Let's add a small description about each technology? Just to make it easier to google what each one is about

```go
// Kernel device-mapper block devices, technology used LVM/Docker and others to create virtual disks
// Loop devices (see `man loop`)
// Default name usually assigned to "swap on ZRAM" block devices
// Linux multiple-device-driver block devices
```

PR #265, opened 2022-01-09
MGMT-8847: Add bf serial value to unknown serials in agent

tsorya:
In BF cards there is know serial of mother board and we get "Unspecified Base Board Serial Number" ,
we should take systemuuid in that case

### DIFF
--- a/src/scanners/machine_uuid_scanner.go
+++ b/src/scanners/machine_uuid_scanner.go
@@ -6,6 +6,8 @@ import (
 	"os"
 	"strings"
 
+	"github.com/thoas/go-funk"
+
 	"github.com/go-openapi/strfmt"
 	"github.com/jaypipes/ghw"
 	"github.com/jaypipes/ghw/pkg/util"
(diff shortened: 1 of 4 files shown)


src/scanners/machine_uuid_scanner.go
```diff
@@ -74,19 +78,22 @@ func (ir *idReader) readSystemUUID() *strfmt.UUID {
...
 	}
-	return md5GenerateUUID(value)
+	return md5GenerateUUID(basedboard.SerialNumber)
 }
 
 func ReadId(d SerialDiscovery) *strfmt.UUID {
 	ir := &idReader{serialDiscovery: d}
 	ret := ir.readMotherboardSerial()
 	if ret == nil {
+		log.Warn("No valid serial for motherboard moving to systemUUID")
```
avishayt, line 96:
```suggestion
		log.Warn("No valid motherboard serial, using system UUID instead")
```

PR #272, opened 2022-01-20
MGMT-8481: Allow fallback for generate uuid from mac for kaloom deployment

tsorya:
[MGMT-8481](https://issues.redhat.com/browse/MGMT-9003): Allow fallback for generate uuid from mac for kaloom deployment
Currently kaloom are not able to deploy multinode cluster as their machines generates same uuid as serial uuid and mother board uuid has "Default string" value
We want to create quick fix till we fully move to using macs


flaper87:
@tsorya mind using `MGMT-8481` instead? It's the original issue tracking this change.

flaper87 (commented):
code looks good, let's change the jira tracking number in both the PR and the commit :pray: 

/cc @slaviered 
/assign @flaper87

flaper87 (approved):
:spaghetti:

PR #277, opened 2022-01-25
MGMT-9040: include only Luks section of ignition in response

danielerez:
CheckAPIConnectivity should return only the Luks section of the ignition. That's the only information needed by the service, so it's redundant to send the entire ignition config which tends to be quite large.


src/apivip_check/apivip_check.go
```diff
@@ -38,7 +40,25 @@ func CheckAPIConnectivity(checkAPIRequestStr string, log logrus.FieldLogger) (st
...
-	return createResponse(true, ignition), "", 0
+	config, _, err := v3_2.ParseCompatibleVersion([]byte(ignition))
+	if err != nil {
+		wrapped := errors.Wrap(err, "Invalid ignition format")
+		log.WithError(err).Error(wrapped.Error())
+		return createResponse(false, ""), wrapped.Error(), 0
+	}
+
+	filteredConfig := types.Config{}
+	filteredConfig.Storage.Luks = config.Storage.Luks
```
omertuc, line 51: Would be nice to include the information in the commit message here:
```
CheckAPIConnectivity should return only the Luks section of the ignition.
This is the only information needed by the service, so it's redundant
to send the entire ignition config which tends to be quite large.
```

danielerez: Yeah good idea, added.

PR #278, opened 2022-01-30
MGMT-8752: handle ocp router 503 errors

slaviered:
The OCP router sends a 503 response with a standard HTML formatted message when the service is down.

However, the go client generated by swagger expects a JSON-formatted API message and fails to analyze the response body. The result is an unclear error message in the agent, installer, and controller when the service is down.

Based on the recommended course of action this PR replaces the consumer of text/html responses to handle all sort of embedded responses (Both text, openshift and assisted-service generated errors) so all possible 503 (and other general responses) will be covered. Since BE errors are not printed well with the generic Error() function in case of embedded reason, it also put forward a marshaling code for all sort of errors, so in all cases the logs will be informative.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.16
 require (
 	github.com/coreos/go-systemd/v22 v22.3.2 // indirect
 	github.com/coreos/ignition/v2 v2.13.0
+	github.com/go-openapi/runtime v0.19.24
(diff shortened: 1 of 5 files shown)


avishayt (approved):

filanov:
No tests?

PR #283, opened 2022-02-01
MGMT-9043: Set multipath device as eligible

avishayt:
Properly report iSCSI and DeviceMapper devices, and allow installation
on DeviceMapper devices.

### DIFF
--- a/src/inventory/disks.go
+++ b/src/inventory/disks.go
@@ -148,17 +148,36 @@ func unknownToEmpty(value string) string {
 	return value
 }
 
+func isDeviceMapper(disk *ghw.Disk) bool {
+	return strings.HasPrefix(disk.Name, "dm-")
+}
+
+func (d *disks) isMultipath(disk *ghw.Disk) bool {
+	if !isDeviceMapper(disk) {
+		return false
+	}
+	path := filepath.Join("/sys", "block", disk.Name, "dm", "uuid")
+	b, err := d.dependencies.ReadFile(path)
+	if err != nil {
+		return false
+	}
+	if strings.HasPrefix(string(b), "mpath-") {
+		return true
+	}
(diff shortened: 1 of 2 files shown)


src/inventory/disks.go
```diff
@@ -148,6 +148,10 @@ func unknownToEmpty(value string) string {
 	return value
 }
 
+func isDeviceMapper(disk *ghw.Disk) bool {
+	return strings.HasPrefix(disk.Name, "dm-0")
```
omertuc, line 152: Shouldn't this be `dm-*` rather than `dm-0`?

src/inventory/disks.go
```diff
@@ -180,8 +184,7 @@ func checkEligibility(disk *ghw.Disk) (notEligibleReasons []string, isInstallati
 }
 
 func IsPhysicalDisk(disk *block.Disk) bool {
-	return !(strings.HasPrefix(disk.Name, "dm-") || // Kernel device-mapper block devices, technology used by LVM/Docker
```
omertuc, line 183: The reason we had this check was to get rid of LVM devices from appearing in the list of disks. 

Instead of removing this exclusion completely, can you instead only consider a `dm-` disk physical if it's multipath? 

(Not sure if iSCSI also shows up as device-mapper, if so, then same thing applies to that to)

src/inventory/disks.go
avishayt: Yes, good call

src/inventory/disks.go
avishayt: Good point, but we now have an epic to support LVM installations:
https://issues.redhat.com/browse/MGMT-8262
Maybe the issue last time is that devices weren't clearly marked.

PR #287, opened 2022-02-02
Install smartmontools from stream8 repos

osherdp:
Seems like with the move to stream8 base image, we can also RPM package for ``smartmontools`` from the official stream 8 repositories
Version installed is ``7.1-1`` which should be okay based on the comment
/cc @omertuc

### DIFF
--- a/Dockerfile.assisted_installer_agent
+++ b/Dockerfile.assisted_installer_agent
@@ -18,19 +18,16 @@ RUN dnf install -y \
 		podman \
 		# inventory
 		dmidecode ipmitool biosdevname file fio \
-        # Since centos:8 is using an older version of smartmontools (smartctl needed for the inventory command), we download the fedora 32 RPM package and
-        # install it instead of using the centos repos. (We need version 7.1+ for the `--json=c` flag to work)
-        # Note that doing this is NOT needed in the downstream Dockerfile, the base image used there is openshift/ose-base:ubi8 and it has the 7.1 version of smartmontools in the repos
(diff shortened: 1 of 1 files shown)


Dockerfile.assisted_installer_agent
```diff
@@ -30,7 +27,7 @@ RUN dnf install -y \
 		tar openssh-clients\
 		# ntp_synchronizer
 		chrony && \
-		dnf update -y systemd && dnf clean all
+			dnf update -y systemd && dnf clean all
```
omertuc, line 30: accidental diff?

osherdp: no, I just like indentation differences between parameters and next commands
just a styling preference that I can revert if you'd like

omertuc: I think the more common way to differentiate is to put the `&&` at the beginning like:
```dockerfile
RUN dnf install -y \
		findutils iputils \
		podman \
		# inventory
		dmidecode ipmitool biosdevname file fio \
		smartmontools \
		# free_addresses
		nmap \
		# dhcp_lease_allocate
		dhclient \
		# logs_sender
		tar openssh-clients\
		# ntp_synchronizer
		chrony \
		&& dnf update -y systemd && dnf clean all
```

omertuc: your call, lmk if you keep it the same or I should lgtm

osherdp: done

omertuc: lol you still kept the indentation

osherdp: ahh, should have read better
done

osherdp:
doesn't really affect downstream build, as RHEL already has the relevant package

PR #288, opened 2022-02-06
NO-ISSUE: Dockerfile: don't install docs and weak dependencies

vrutkovs:
This shaves off ~50Mb from final container size:
```
$ podman images | grep quay.io/vrutkovs/assisted-installer-agent                                               
quay.io/vrutkovs/assisted-installer-agent                slim-agent              869abe541647  16 seconds ago      948 MB
quay.io/vrutkovs/assisted-installer-agent                master                  558e0ee6bd54  22 minutes ago      1.04 GB
```

### DIFF
--- a/Dockerfile.assisted_installer_agent
+++ b/Dockerfile.assisted_installer_agent
@@ -16,7 +16,7 @@ FROM quay.io/centos/centos:stream8
 ARG TARGETPLATFORM
 RUN if [[ "$TARGETPLATFORM" == "linux/amd64" || -z "$TARGETPLATFORM" ]] ; then  dnf install -y biosdevname ; fi
 
-RUN dnf install -y \
+RUN dnf install --setopt=install_weak_deps=False --setopt=tsdocs=False -y \
 		findutils iputils \
 		podman \
 		# inventory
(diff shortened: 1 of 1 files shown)


omertuc:
Can you summarize which weak dependencies are being dropped by doing this? It would be nice to eyeball this list to make sure that it has no potential effects on the behavior of the agent

vrutkovs:
```
Dependencies resolved.
================================================================================
 Package                Architecture Version              Repository       Size
================================================================================
Upgrading:
 systemd                x86_64       239-56.el8           baseos          3.6 M
 systemd-libs           x86_64       239-56.el8           baseos          1.1 M
 systemd-pam            x86_64       239-56.el8           baseos          480 k
Installing dependencies:
 xkeyboard-config       noarch       2.28-1.el8           appstream       782 k
Installing weak dependencies:
 diffutils              x86_64       3.6-6.el8            baseos          358 k
 libxkbcommon           x86_64       0.9.1-1.el8          appstream       116 k
```
is what I see now, but these may show up during a routine install/update later on

vrutkovs:
>It would be nice to eyeball this list to make sure that it has no potential effects on the behavior of the agent

weak dependencies are "you might find these packages useful when using package X", these are enhancing experience, not strictly required like dependecies

omertuc:
> > It would be nice to eyeball this list to make sure that it has no potential effects on the behavior of the agent
> 
> weak dependencies are "you might find these packages useful when using package X", these are enhancing experience, not strictly required like dependecies

Yeah I understand that, I'm just afraid that some packages may change their default behavior in some way when those dependencies are missing. We're relying heavily on the exact behavior of a lot of utilities in this project so I want to go through the list to make sure it can't bite us

osherdp:
@vrutkovs it only complains about the commit message apparently

vrutkovs:
>Should not have downstream-change-needed label

Not sure what that label means - do we need to update ART image definition too?

omertuc:
> > Should not have downstream-change-needed label
> 
> Not sure what that label means - do we need to update ART image definition too?



> > Should not have downstream-change-needed label
> 
> Not sure what that label means - do we need to update ART image definition too?

That label was originally referring to the downstream gerrit dockerfiles that are used to build the actual agent image served on cloud.redhat.com

osherdp:
In this case I think no change is needed in the D/S Dockerfile

PR #293, opened 2022-02-21
NO-ISSUE: remove Jenkins publishing step

osherdp:
Now that we have no need in bundles+indexes created by Jenkins on ocpmetal, we can completely ditch all the "publishing" parts.

This should leave the subsystem running (which we currently have no alternative to) but remove everything else that is already duplicated by other jobs/automation/etc.

/cc @eliorerz @filanov @lranjbar


eliorerz (approved):
LGTM

PR #295, opened 2022-02-28
WIP: Moving to new service-agent protocol

tsorya:


src/commands/step_processor.go
```diff
@@ -106,14 +112,30 @@ stderr:
 	return s.createStepReply(stepType, stepID, stdout, stderr, exitCode)
 }
 
+func (s *stepSession) handleSingleStepV2(stepType models.StepType, stepID string, command string, args []string, handler HandlerType) models.StepReply {
+	s.Logger().Infof("Creating execution step for %s %s", stepType, stepID)
+	var err error
+
+	// TODO: remove command == "" after agent changes for new protocol will be pushed, added to allow pushing agent before service
```
eranco74, line 119: Add the Jira task to the comment

src/commands/step_processor.go
```diff
@@ -106,14 +112,30 @@ stderr:
...
+	if action, ok := stepActionMap[stepType]; ok && command == "" {
+		command, args, err = action(args)
+		s.Logger().WithError(err).Infof("Failed to create command for step %s", stepType)
```
eranco74, line 122: This log seems misplaced, perhaps you meant to place it inside the `if err != nil {` scope

src/commands/step_processor.go
```diff
@@ -106,14 +112,30 @@ stderr:
 	return s.createStepReply(stepType, stepID, stdout, stderr, exitCode)
 }
 
+func (s *stepSession) handleSingleStepV2(stepType models.StepType, stepID string, command string, args []string, handler HandlerType) models.StepReply {
+	s.Logger().Infof("Creating execution step for %s %s", stepType, stepID)
+	var err error
+
+	// TODO: remove command == "" after agent changes for new protocol will be pushed, added to allow pushing agent before service
+	if action, ok := stepActionMap[stepType]; ok && command == "" {
```
eranco74, line 120: If the stepType exists in the map why do you care about the command?

src/commands/steps.go
```diff
@@ -0,0 +1,229 @@
+package commands
```
eranco74, line 1: Document what are the expected args for each function.
It's unclear what params each function should get when they all get (args []string)
Also, consider defining an interface for the steps args and create the relevant command args struct for each command where you call `action(args)` (it can save you the trouble of validating the args in each function)

src/commands/steps.go
```diff
@@ -0,0 +1,229 @@
...
+		"id=`podman ps --quiet --filter \"name=disk_performance\"` ; " +
+			"test ! -z \"$id\" || " +
+			fmt.Sprintf("timeout %s ", args[1]) +
+			"podman run --privileged --rm --quiet -v /dev:/dev:rw -v /var/log:/var/log -v /run/systemd/journal/socket:/run/systemd/journal/socket " +
+			"--name disk_performance " +
+			config.GlobalAgentConfig.AgentVersion + " disk_speed_check '" +
+			args[0] + "'",
+	}
+
+	return "bash", arguments, nil
```
eranco74, line 228: Other commands use `sh` why do you need `bash` here?

src/commands/steps.go
```diff
@@ -0,0 +1,229 @@
...
+		log.WithError(err).Errorf("Failed to parse timeout value to float: %s", args[1])
+		return "", nil, err
+	}
+
+	arguments := []string{
+		"-c",
+		"id=`podman ps --quiet --filter \"name=disk_performance\"` ; " +
+			"test ! -z \"$id\" || " +
+			fmt.Sprintf("timeout %s ", args[1]) +
+			"podman run --privileged --rm --quiet -v /dev:/dev:rw -v /var/log:/var/log -v /run/systemd/journal/socket:/run/systemd/journal/socket " +
```
eranco74, line 222: This is doing the same thing as `checkAlreadyRunning` in the freeAddressesCmd.
Perhaps we need a better and consisted way to check that

src/commands/steps.go
```diff
@@ -0,0 +1,229 @@
...
+
+	podmanRunCmd = append(podmanRunCmd, args...)
+	cmdString := shellescape.QuoteCommand(podmanRunCmd)
+
+	// Sometimes the address scanning takes longer than the interval we wait between invocations.
+	// To avoid flooding the log with "container already exists" errors, we silently fail by manually
+	// checking if it exists and only running if it doesn't
+	checkAlreadyRunningCmd := fmt.Sprintf("podman ps --format '{{.Names}}' | grep -q '^%s$'", containerName)
+
+	return "sh", []string{"-c", fmt.Sprintf("%s || %s", checkAlreadyRunningCmd, cmdString)}, nil
```
eranco74, line 122: So we actually run 2 commands, this isn't optimal, unsure how to avoid it with the current implementation

src/commands/step_processor.go
tsorya: i commented with the explanation , to be able to push agent without service, then we can push service

src/commands/step_processor.go
tsorya: i know, will be fixed

src/commands/steps.go
tsorya: i thought about it but we can do it as optimization. I don't see any reason to create interface as i don't see any reason for it

src/commands/steps.go
tsorya: i don't change anything in the commands, this is not a task. We can change commands later

src/commands/steps.go
tsorya: as i said previously, i don't try to optimize commands , i want to move them as they are now. We can open task for optimizing them

src/commands/steps.go
tsorya: as i said previously, i don't try to optimize commands , i want to move them as they are now. We can open task for optimizing them

src/commands/steps.go
eranco74: OK, open a jira for it

src/commands/action/action.go
```diff
@@ -0,0 +1,55 @@
...
+type ActionInterface interface {
+	Run() (string, []string)
+	Validate() error
+}
+
+type Action struct {
+	ActionInterface
+}
+
+func validateCommon(name string, expectedArgsLength int, args []string, jsonAsStringToValidate string, modelToValidate interface{}) error {
```
eranco74, line 22: jsonAsStringToValidate seems odd, and I think you can avoid it (seems that when there is a modelToVAlidate it's always args[0]).

eranco74, line 22: Consider returning a pointer to the initalized modelToValidate.
This way you can use the model for constructing the command later.

src/commands/action/api_vip_connectivity_check.go
```diff
@@ -0,0 +1,30 @@
...
+func (a apiVipConnectivityCheck) Run() (string, []string) {
+	podmanRunCmd := []string{
+		"run", "--privileged", "--net=host", "--rm", "--quiet",
+		"-v", "/var/log:/var/log",
+		"-v", "/run/systemd/journal/socket:/run/systemd/journal/socket",
+		config.GlobalAgentConfig.AgentVersion,
+		"apivip_check",
+	}
+
+	return podman, append(podmanRunCmd, a.args...)
```
eranco74, line 29: if you create an instance of models.APIVipConnectivityRequest from the args, here you could do:
```suggestion
	return podman, append(podmanRunCmd, apiVipConnectivityRequest.URL)
```
Which is less generic but clearer.

eranco74 (commented):
Overall looks good, just few nits

src/commands/action/action.go
tsorya: it depends and it is right for now, wanted to make is more generic as i can but can change it

src/commands/action/api_vip_connectivity_check.go
tsorya: we can't do it, as the currently it expects dump json and will be unmarshalled in api_vip command 
Till we don't change the way all our cmds works we will need to dump each time and there is no point for it as args have needed value

src/commands/actions/action.go
```diff
@@ -0,0 +1,60 @@
...
+	Validate() error
+}
+
+type Action struct {
+	ActionInterface
+}
+
+func validateCommon(name string, expectedArgsLength int, args []string, modelToValidate runtime.Validatable) error {
+	if len(args) != expectedArgsLength {
+		return fmt.Errorf("%s cmd accepts only 1 params in args, given args %v", name, args)
```
eranco74, line 25:
```suggestion
		return fmt.Errorf("%s cmd accepts %d params in args, given args %v", name, expectedArgsLength, args)
```

src/commands/actions/action.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	if modelToValidate != nil {
+		err := json.Unmarshal([]byte(args[0]), &modelToValidate)
+		if err != nil {
+			log.WithError(err).Errorf("Failed to unmarshal %s: json.Unmarshal, %s", name, args[0])
+			return err
+		}
+		err = modelToValidate.Validate(nil)
+		if err != nil {
+			log.WithError(err).Errorf("Failed to on data validation of %s: data, %s", name, args[0])
```
eranco74, line 35:
```suggestion
			log.WithError(err).Errorf("Failed to validate %s: data, %s", name, args[0])
```

romfreiman:
Does it mean that it's actually working?

tsorya:
@romfreiman currently it runs with old service and CI doesn't test new protocol. I test installation manually

eranco74:
> Does it mean that it's actually working?

No, The new flow isn't triggered because the service is still sending the commands in the old format.

PR #297, opened 2022-03-01
MGMT-9344: Return non-zero exit when subsystem test fails

adriengentil:
As part of moving the subsystem test to prow, we need to know when the
step that runs the test fails.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,13 +61,15 @@ endif
 unit-test:
 	$(MAKE) _test TEST_SCENARIO=unit TIMEOUT=30m TEST="$(or $(TEST),$(shell go list ./... | grep -v subsystem))" || (docker kill postgres && /bin/false)
 
-subsystem: build-image
-	-docker image rm subsystem_agent
-	$(DOCKER_COMPOSE) up --build -d dhcpd wiremock
-	-$(MAKE) _test TEST_SCENARIO=subsystem TIMEOUT=30m TEST="$(or $(TEST),./subsystem/...)"
-	$(DOCKER_COMPOSE) logs dhcpd > dhcpd.log
-	$(DOCKER_COMPOSE) logs wiremock > wiremock.log
-	$(DOCKER_COMPOSE) down
+subsystem: build-image 
+	docker image rm subsystem_agent; \
+	$(DOCKER_COMPOSE) up --build -d dhcpd wiremock; \
+	$(MAKE) _test TEST_SCENARIO=subsystem TIMEOUT=30m TEST="$(or $(TEST),./subsystem/...)"; \
+	rc=$$?; \
+	$(DOCKER_COMPOSE) logs dhcpd > dhcpd.log; \
+	$(DOCKER_COMPOSE) logs wiremock > wiremock.log; \
+	$(DOCKER_COMPOSE) down; \
+	exit $$rc;
 
 generate:
 	find "${ROOT_DIR}" -name 'mock_*.go' -type f -delete


Makefile
```diff
@@ -61,14 +61,20 @@ endif
...
-	-$(MAKE) _test TEST_SCENARIO=subsystem TIMEOUT=30m TEST="$(or $(TEST),./subsystem/...)"
+
+_post_subsystem:
 	$(DOCKER_COMPOSE) logs dhcpd > dhcpd.log
 	$(DOCKER_COMPOSE) logs wiremock > wiremock.log
 	$(DOCKER_COMPOSE) down
 
+subsystem: build-image
+	$(MAKE) _pre_subsystem
+	$(MAKE) _test TEST_SCENARIO=subsystem TIMEOUT=30m TEST="$(or $(TEST),./subsystem/...)" || ($(MAKE) _post_subsystem && /bin/false)
```
osherdp, line 75: it means we only cleanup on success?

Makefile
```diff
@@ -61,14 +61,20 @@ endif
 unit-test:
 	$(MAKE) _test TEST_SCENARIO=unit TIMEOUT=30m TEST="$(or $(TEST),$(shell go list ./... | grep -v subsystem))" || (docker kill postgres && /bin/false)
 
-subsystem: build-image
+_pre_subsystem:
```
omertuc, line 64: target should be PHONY

Makefile
```diff
@@ -61,14 +61,20 @@ endif
...
 	-docker image rm subsystem_agent
 	$(DOCKER_COMPOSE) up --build -d dhcpd wiremock
-	-$(MAKE) _test TEST_SCENARIO=subsystem TIMEOUT=30m TEST="$(or $(TEST),./subsystem/...)"
+
+_post_subsystem:
 	$(DOCKER_COMPOSE) logs dhcpd > dhcpd.log
 	$(DOCKER_COMPOSE) logs wiremock > wiremock.log
 	$(DOCKER_COMPOSE) down
 
+subsystem: build-image
```
osherdp, line 73: in your release PR this will allow you to use "public" makefile targets (and not ones like ``_<name>``)?

Makefile
```diff
@@ -61,14 +61,20 @@ endif
 unit-test:
 	$(MAKE) _test TEST_SCENARIO=unit TIMEOUT=30m TEST="$(or $(TEST),$(shell go list ./... | grep -v subsystem))" || (docker kill postgres && /bin/false)
 
-subsystem: build-image
+_pre_subsystem:
 	-docker image rm subsystem_agent
 	$(DOCKER_COMPOSE) up --build -d dhcpd wiremock
-	-$(MAKE) _test TEST_SCENARIO=subsystem TIMEOUT=30m TEST="$(or $(TEST),./subsystem/...)"
+
+_post_subsystem:
```
omertuc, line 68: same here

Makefile
osherdp: not only cleanup, but also gathering

osherdp (changes requested):

Makefile
```diff
@@ -61,14 +61,20 @@ endif
...
 	$(DOCKER_COMPOSE) logs dhcpd > dhcpd.log
 	$(DOCKER_COMPOSE) logs wiremock > wiremock.log
 	$(DOCKER_COMPOSE) down
 
+subsystem: build-image
+	$(MAKE) _pre_subsystem
```
omertuc, line 74:
```suggestion
subsystem: build-image _pre_subsystem
```

omertuc:
How is splitting up this target into pre, do, post achieving the goal of the PR?

osherdp:
> How is splitting up this target into pre, do, post achieving the goal of the PR?

I think because it used to be:
```
-$(MAKE) _test TEST_SCENARIO=subsystem TIMEOUT=30m TEST="$(or $(TEST),./subsystem/...)"
```

so the minus sign at start of the line made it to ignore return code

Makefile
adriengentil: the trick is in the `|| ($(MAKE) _post_subsystem && /bin/false)`.
If `_test` target fails, the command after the `||` will be executed and since it return a non-zero (`/bin/false`), make will stop the process.

We use the same logic to run the unit tests: https://github.com/openshift/assisted-installer-agent/blob/master/Makefile#L44

@omertuc I think it answers your question

osherdp: sorry, I meant "we only cleanup on failure"
not sure it's that critical though, just an observation

Makefile
adriengentil: I'm not sure to understand your question, but I don't change the interface so will we still launch the subsystem tests using `make subsystem`

Makefile
osherdp: ah no, all good
please ignore

omertuc: I get what you're doing but I feel it's a bit hard to follow. rather than short-circuiting / `/bin/false` tricks, maybe we can do something like this pseudocode:

```
// Run the tests, record the result
result = run_tests()

// Cleanup, ignore cleanup failures
cleanup || true

// Return the test result
exit $result
```

adriengentil: Each statement in a makefile is executed in a sub shell which makes things a bit awkward, but I just discovered `.ONESHELL` which could makes things simpler.

https://www.gnu.org/software/make/manual/html_node/One-Shell.html

osherdp: make sure to add a comment when using ``.ONESHELL``, because I think it's not a known feature

adriengentil:
@omertuc @osherdp I updated my PR, all the statements are now executed in the same shell (thanks to the backslashes) in the `subsystem` target. I think it's cleaner than before, what do you think?

(I mentioned to use `.ONESCRIPT` in a post above, but it applies to the whole makefile, and as @osherdp pointed out, it can be confusing because it's not a well known feature, so I think it's a good idea to use it)

osherdp (approved):
yes, it's cleaner

PR #300, opened 2022-03-02
NO-ISSUE: Remove small timeout subsystem test

adriengentil:
This test expects that the pull of a valid image will timeout. It turns
out that it is flaky because the pull of the image may happen before the
timeout is reached.

I guess this issue was hidden because the tests in Jenkins always
succeeded before the merge of
https://github.com/openshift/assisted-installer-agent/pull/297.

This PR removes the test.

Example of failures:
* in Jenkins: http://assisted-jenkins.usersys.redhat.com/blue/organizations/jenkins/assisted-installer-agent/detail/PR-299/1/pipeline
* in Prow: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_release/26575/rehearse-26575-pull-ci-openshift-assisted-installer-agent-master-subsystem-test/1498708068857286656/artifacts/subsystem-test/assisted-agent-test/build-log.txt


adriengentil:
> You mean the pull may happen so fast that it finished in under 2 seconds?

yes, it seems so

omertuc:
Shouldn't we fix this by using a bigger image, or by making the timeout even smaller?

omertuc:
Could the pull be too fast because the image is already pulled (i.e., cached)? Could we somehow avoid this cache instead of deleting the test?

omertuc:
Actually seems like the test already tried to rule out cache:
https://github.com/openshift/assisted-installer-agent/blob/bf66b232529b5023f29eca554f9ed6c51b2a6044/subsystem/container_image_availability_test.go#L190-L193

adriengentil:
> Shouldn't we fix this by using a bigger image, or by making the timeout even smaller?

it could, but the test is still tied by external dependencies, we still hope that it behave in that way. Maybe it can be seen from a different angle, but I don;t have much knowledge on that.

omertuc:
I'm running this test locally modifies with a 60 seconds timeout and the test is still successful, which makes me think that if the test was failing for the reasons you think it does, then I should have seen the exact opposite

adriengentil:
Well , the issue does not consistently fails, it fails sometimes, and I faced it a couple of times with a machine in equinix.

omertuc:
I didn't say it consistently fails, I said that if the job is flaky because of the timeout you mentioned, then increasing the timeout from 2 to 60 locally would have caused it to consistently fail locally, but that doesn't happen - the opposite happens, it consistently succeeds, which makes me think your hypothesis is wrong

adriengentil:
Sorry :man_facepalming, I read too fast!
I get the same result, this test puzzled me

omertuc:
>this test puzzles me

It's likely that it's completely broken, don't assume its current behavior is the intended behavior

We have no practical way of knowing which tests are green due to actually doing their job and testing what we want them to test or they're green because they don't test anything, I think in this case it's the latter

adriengentil:
I did the test on RHEL with pullTimeoutInSeconds=60s, and it behaves as expected (it checks that the image is absent until it times out).

If like me you were running it on Fedora, should get this error in your system logs:
```
Mar 03 10:39:04 cpe-172-100-0-1.twcny.res.rr.com container_image[71155]: time="03-03-2022 09:39:04" level=warning msg="Pulling image quay.io/coreos/etcd:latest wasn't available" file="container_image_availability.go:127" error="podman pull exited with non-zero exit code 1: \n podman: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by podman)\npodman: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by podman)\n"
```

I guess it's because `podman` executable is actually the host's `docker` binary in the container running the agent, and it expects a higher version of the glibc to be available:
```
Executing step: <ad4e9391-10a4-4f91-ae29-c12780f54cbe>, command: <docker>, args: <[run --privileged --rm -v /var/log:/var/log -v /run/systemd/journal/socket:/run/systemd/journal/socket -v /usr/bin/docker:/usr/bin/podman -v /var/run/docker.sock:/var/run/docker.sock quay.io/edge-infrastructure/assisted-installer-agent:latest /usr/bin/container_image_availability --request {\"images\":[\"quay.io/aptible/hello-world\",\"quay.io/coreos/etcd:latest\"],\"timeout\":60}]>" file="step_processor.go:97
```

adriengentil:
FYI, there is a unit test that checks the behavior in case of timeout: https://github.com/openshift/assisted-installer-agent/blob/master/src/container_image_availability/container_image_availability_test.go#L170-L179

osherdp (approved):

osherdp:
/override ci/prow/e2e-metal-assisted-olm-suite
not relevant at all to this change

/unhold

PR #314, opened 2022-03-21
AGENT-34: Write HostID to file at registration

zaneb:
In order to work out whether the agent is running on the same host as
assisted-service, write the host ID to the file
/var/run/assisted-agent-host-id during registration.

### DIFF
--- a/src/commands/service_api.go
+++ b/src/commands/service_api.go
@@ -1,6 +1,9 @@
 package commands
 
 import (
+	"os"
+	"path/filepath"
+
 	"github.com/go-openapi/strfmt"
 	"github.com/openshift/assisted-installer-agent/src/config"
 	"github.com/openshift/assisted-installer-agent/src/scanners"
@@ -8,6 +11,11 @@ import (
 	agent_utils "github.com/openshift/assisted-installer-agent/src/util"
 	"github.com/openshift/assisted-service/client/installer"
 	"github.com/openshift/assisted-service/models"
+)
+
+const (
+	hostIDDir  = "/var/run/assisted-agent"
+	hostIDFile = "host-id"
 )
 
 type serviceAPI interface {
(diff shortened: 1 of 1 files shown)


src/commands/service_api.go
```diff
@@ -26,6 +28,10 @@ func (v *v2ServiceAPI) RegisterHost(s *session.InventorySession) (*models.HostRe
 		hostID = strfmt.UUID(config.GlobalDryRunConfig.ForcedHostID)
 	}
 
+	if err := os.WriteFile("/var/run/assisted-agent-host-id", []byte(hostID), 0644); err != nil {
```
gamli75, line 31: Please add a const value for the path & file name.
When and where we will use this file? If it's in the service - can we have this const value in one place and use it in both  agent & the service?

omertuc, line 31: Can you avoid doing this on dry-run mode? Dry run is used for running a lot of fake agents on a single machine and they would all race to opening/closing this file

romfreiman:
Can you elaborate how will it help? what changes will be introduced in the assisted service - I could find it in jira

avishayt:
> Can you elaborate how will it help? what changes will be introduced in the assisted service - I could find it in jira

+1

filanov:
> > Can you elaborate how will it help? what changes will be introduced in the assisted service - I could find it in jira
> 
> +1

+1

eranco74:
Why do we need this file? can't the service get the host-id the same way the agent does?

zaneb:
> Can you elaborate how will it help? what changes will be introduced in the assisted service - I could find it in jira

I intend to propose checking it directly in the assisted-service host registration code, and setting the host as a bootstrap if it matches: https://github.com/zaneb/assisted-service/commit/ff0fcf92587ba4785a633899b1c217be9ea33e71
(I haven't created a PR yet, because I need to refactor some other stuff in fleeting before I can test it.)

However, even if a different approach turned out to be preferred (e.g. using the ReST API to designate the host as a bootstrap), we would still need to know the Host ID of node0. So I think this is a necessary first step in any case.

> Why do we need this file? can't the service get the host-id the same way the agent does?

We'd have to use the exact same logic. Currently assisted-service does not depend on the agent (and if it tried there would be a circular dependency). So the code would have to be moved around, and there'd be a risk of version skew if it ever changed.
Since we're aiming to communicated between two processes running on the same host, using the filesystem seems like a convenient way to do it.

filanov:
We can expose the package that generate the ID, the ID should always be the same for a specific machine

flaper87:
[Commented on the service side PR](https://github.com/openshift/assisted-service/pull/3589#issuecomment-1088494768) 

/hold

zaneb:
/close
replaced by replaced by https://github.com/openshift/assisted-service/pull/3627

PR #320, opened 2022-03-31
MGMT-9807: Enable next step runner to run tools directly in addition to the running them in container

ori-amizur:
- Create runner interface that encapsulates running operation
- Create runner implemntation which is compatible with current
  implementation and make it the current implemntation for running tools
  in next step runner

/cc @omertuc 
/cc @filanov

### DIFF
--- a/src/commands/runner.go
+++ b/src/commands/runner.go
@@ -1,0 +1,99 @@
+package commands
+
+import (
+	"github.com/openshift/assisted-installer-agent/src/commands/actions"
+	"github.com/openshift/assisted-installer-agent/src/util"
+	"github.com/openshift/assisted-service/models"
+)
+
+type NextStepRunnerFactory interface {
+	Create(stepType models.StepType, command string, args []string) (Runner, error)
+}
+
+// Runner is the means to allow pluggable running mechanism to agent.
+// The runner factory should be initiated once, and should be passed forward in the execution floe by dependency
+// injection.
+type Runner interface {
+	Run() (stdout, stderr string, exitCode int)
+	Command() string
+	Args() []string
+}
+
+func NewActionRunner(action actions.ActionInterface) Runner {
+	return &actionRunner{
+		action:  action,
+		handler: util.Execute,
+	}
+}
+
+func NewPrivilegedActionRunner(action actions.ActionInterface) Runner {
+	return &actionRunner{
+		action:  action,
+		handler: util.ExecutePrivileged,
+	}
+}
+
+type nextStepRunnerFactory struct{}
+
+func NewNextStepRunnerFactory() NextStepRunnerFactory {
+	return &nextStepRunnerFactory{}
+}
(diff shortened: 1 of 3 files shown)


src/commands/step_processor.go
```diff
@@ -91,64 +92,46 @@ func (s *stepSession) createStepReply(stepType models.StepType, stepID string, o
...
-		return s.createStepReply(stepType, stepID, "", errStr, -1)
-	}
-
-	s.Logger().Infof("Executing step: <%s>, command: <%s>, args: <%v>", stepID, command, args)
-	stdout, stderr, exitCode := handler(command, args...)
+func (s *stepSession) handleSingleStep(stepType models.StepType, stepID string, runner Runner) models.StepReply {
+	stdout, stderr, exitCode := runner.Run()
 	if exitCode != 0 {
-		s.Logger().Errorf(`Step execution failed (exit code %v): <%s>, command: <%s>, args: <%v>. Output:
+		s.Logger().Errorf(`Step execution failed (exit code %v): <%s>, . Output:
```
omertuc, line 98: Please update the corresponding pattern in our triage signature automation https://github.com/openshift-assisted/assisted-installer-deployment/blob/a0085457e4e3b7d518a29bd45ba056093d0a5794/tools/add_triage_signature.py#L960-L964

omertuc: Also add a comment linking to it so it won't be missed in the future

src/commands/runner.go
```diff
@@ -0,0 +1,81 @@
+package commands
+
+import (
+	"github.com/openshift/assisted-installer-agent/src/commands/actions"
+	"github.com/openshift/assisted-installer-agent/src/util"
+	"github.com/openshift/assisted-service/models"
+)
+
+/*
```
omertuc, line 9: Documentation is misplaced (should be above `Runner`) and also should use regular comments, see rather than multiline comments. See https://go.dev/blog/godoc

src/commands/runner.go
```diff
@@ -0,0 +1,81 @@
...
+ * Runner is the means to allow pluggable running mechanism to agent.
+ * The runner factory should be initiated once, and should be passed forward in the execution floe by dependecy
```
omertuc, line 11: nit s/floe/flow s/dependecy/dependency

omertuc:
So the plan is in the future to add a new implementation of this interface to actually achieve "Enable next step runner to run tools directly in addition to the running them in container".

Right now I don't see how this PR achieves that goal, only how it lays the groundwork for it.

Is my understanding correct?

src/commands/step_processor.go
ori-amizur: Will open PR for triage signature after this PR is merged

ori-amizur: Not needed. Returned to the same format as it was previously

PR #323, opened 2022-04-03
MGMT-9747: Implement params validation for install cmd

tsorya:
Adding validations for install command


src/commands/actions/install_cmd.go
```diff
@@ -23,14 +33,52 @@ var podmanBaseCmd = [...]string{
...
+		err = validations.ValidateInstallerArgs(installAgs)
+		if err != nil {
+			return err
+		}
+	}
+
+	if a.installParams.OpenshiftVersion != "" {
+		_, err := version.NewVersion(a.installParams.OpenshiftVersion)
+		if err != nil {
+
```
avishayt, line 75:
```suggestion
```

src/commands/actions/install_cmd.go
```diff
@@ -132,4 +180,80 @@ func getProxyArguments(proxy *models.Proxy) []string {
...
+}
+
+func validateMustGatherImages(mustGatherImage string) error {
+	var imageMap map[string]string
+	err := json.Unmarshal([]byte(mustGatherImage), &imageMap)
+	if err != nil {
+		// must gather image can be a string and not json
+		imageMap = map[string]string{"ocp": mustGatherImage}
+	}
+	r, errCompile := regexp.Compile(`^(([a-zA-Z0-9\-\.]+)(:[0-9]+)?\/)?[a-z0-9\._\-\/@]+[?::a-zA-Z0-9_\-.]+$`)
```
avishayt, line 221: I guess this is used elsewhere for other image validation?  If so, make it common.

PR #328, opened 2022-04-05
MGMT-9453: Implement params validation per each step

tsorya:
[MGMT-9453](https://issues.redhat.com/browse/MGMT-9453): Implement params validation per each step
    Moving part of step actions to run commands without need to run os.exec
    that allows not to validate every params as in that case bad commands
    could not be inject.

[MGMT-9453](https://issues.redhat.com/browse/MGMT-9453): moving connectivity check from command to it's folder
    Running connectivity check from action without os.exec
[MGMT-9747](https://issues.redhat.com/browse/MGMT-9747): Implement params validation for install cmd
[MGMT-9453](https://issues.redhat.com/browse/MGMT-9453): updating to new service swagger


src/commands/actions/connectivity_check_cmd.go
```diff
@@ -15,16 +15,22 @@ func (a *connectivityCheck) Validate() error {
...
 func (a *connectivityCheck) CreateCmd() (string, []string) {
-	commandArgs := []string{
-		"run", "--privileged", "--net=host", "--rm", "--quiet",
-		"-v", "/var/log:/var/log",
-		"-v", "/run/systemd/journal/socket:/run/systemd/journal/socket",
-		config.GlobalAgentConfig.AgentVersion,
-		"connectivity_check",
-	}
-	return podman, append(commandArgs, a.args...)
+	return "", nil
```
ori-amizur, line 23: Maybe remove CreateCmd and use Command and Args instead ?

src/commands/actions/install_cmd.go
```diff
@@ -23,19 +35,56 @@ var podmanBaseCmd = [...]string{
 type install struct {
 	args          []string
 	installParams models.InstallCmdRequest
+	filesystem    afero.Fs
```
ori-amizur, line 38: Seems that os standard package provide all this functionality.  Why new package ?

src/commands/actions/install_cmd.go
```diff
@@ -132,4 +181,93 @@ func getProxyArguments(proxy *models.Proxy) []string {
...
+		// TODO: adding check for supported operators
+		if !funk.Contains([]string{"cnv", "lso", "ocs", "odf", "ocp"}, op) {
+			return fmt.Errorf("operator name %s validation failed", op)
+		}
+	}
+	return nil
+}
+
+func (a *install) validateDisks() error {
+	disksToValidate := append(a.installParams.DisksToFormat, swag.StringValue(a.installParams.BootDevice))
```
ori-amizur, line 240: We discover disks in the inventory.  We do not return all disks (we do some filtering).  Maybe we can reuse ?

src/commands/actions/install_cmd.go
tsorya: needed it to be able to mock filesystem

src/commands/actions/install_cmd.go
tsorya: The issue is that we can't run get disks from current next_step_runner as mounts are not right and we don't get full data.
We had a discussion with @avishayt about it and decided not to change mounts for now at least.
Another option was to have some variable that will be shared between inventory and install but became and option only after moving to new runner code. I think that it can stay these way for now.

PR #329, opened 2022-04-07
Add ocp release support

andfasano:
This patch adds the required Dockerfile for building the assisted-installer-agent image to be shipped in the OCP release payload.
It also adds a vendor folder to allow building the image in disconnected environments


omertuc:
Ah but it's using non-public images `registry.ci.openshift.org/ocp/4.11:base`. Can't those be set through `openshift/release` configs instead?

andfasano:
> Don't see a reason to keep around the older Dockerfile, seems like unnecessary duplication. can you try to `mv Dockerfile.ocp Dockerfile.assisted_installer_agent` and see whether the CI is still happy?
> 
Yes, didn't want to touch the previous Dockerfiles, as I wasn't sure if they were still needed for some other specific reasons (I saw also a skipper file). I think it could makes sense to merge them if possible, but I'd like to keep the .ocp extension, as usually (but not always) highlights the fact that's used for building an image to be shipped in the release, and eventually it could be addressed in a separate PR

> EDIT: Ah but it's using non-public images `registry.ci.openshift.org/ocp/4.11:base`. Can't those be overridden through `openshift/release` configs instead?

Not sure about this point, the CI has access to such images

omertuc:
Can you add a vendor-diff Makefile lint target like in similar installer/controller [PR](https://github.com/openshift/assisted-installer/pull/436)?

Also follow [this](https://github.com/openshift/assisted-installer/pull/436#issuecomment-1094263435) comment (that has yet to be resolved by the PR author) to make sure that the target actually runs as part of `ci-lint`

andfasano:
> Can you add a vendor-diff Makefile lint target like in similar installer/controller [PR](https://github.com/openshift/assisted-installer/pull/436)?
> 
> Also follow [this](https://github.com/openshift/assisted-installer/pull/436#issuecomment-1094263435) comment (that has yet to be resolved by the PR author) to make sure that the target actually runs as part of `ci-lint`

Sure, done.

omertuc:
Weird failure https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_assisted-installer-agent/329/pull-ci-openshift-assisted-installer-agent-master-subsystem-test/1514187356641431552

/test subsystem-test

omertuc:
Happened again :( 

/test subsystem-test

PR #330, opened 2022-04-10
MGMT-9854: Eliminate use of global variables, and use context variables instead

ori-amizur:
In order to launch multiple agents and multiple tools representing
different hosts in the same process for the swarm tool , use of global variables has to be
eliminated, and use function parameters instead.

/cc @omertuc


omertuc:
That's great. You'll want to do a similar thing in the installer/controller, they also have similar global config

PR #335, opened 2022-04-22
MGMT-10080: Agent subsystem crashes on panic when trying to register host with no DMI information

ori-amizur:
When DMI information is missing, the agent attempts to use any interface
which is physical/vlan/bond.  When non exists, the agent crashes because
it assumes there is at least one such interface.
This change will mount DMI information directory to the agent container
in subsystem, in order to simulate that DMI information exists

/cc @adriengentil


src/scanners/machine_uuid_scanner.go
```diff
@@ -101,6 +102,25 @@ func (ir *idReader) readMotherboardSerial() *strfmt.UUID {
 	return md5GenerateUUID(basedboard.SerialNumber)
 }
 
+func getTemporaryUUID() *strfmt.UUID {
+	interfaces, err := net.Interfaces()
```
omertuc, line 106: Add a comment explaining that unlike inventory.GetInterfaces, this list of interfaces is completely unfiltered

src/scanners/machine_uuid_scanner.go
```diff
@@ -101,6 +102,25 @@ func (ir *idReader) readMotherboardSerial() *strfmt.UUID {
...
+	if err != nil {
+		log.WithError(err).Warnf("Could not get network interfaces")
+		return nil
+	}
+	sort.Slice(interfaces, func(i, j int) bool {
+		return interfaces[i].Name < interfaces[j].Name
```
omertuc, line 112: Shouldn't the sort be according to HardwareAddr like it is for the parent function?

omertuc: If not, can you please add the reason as for why we're sorting by name rather than HardwareAddr to a comment above this line?

src/scanners/machine_uuid_scanner.go
```diff
@@ -101,6 +102,25 @@ func (ir *idReader) readMotherboardSerial() *strfmt.UUID {
 	return md5GenerateUUID(basedboard.SerialNumber)
 }
 
+func getTemporaryUUID() *strfmt.UUID {
```
omertuc, line 105: Some documentation to this function would be nice:
1. What makes this UUID "Temporary"? What does that mean exactly?
2. What are the risks of using it? Could the UUID change between reboots? What would be the consequences of that?
3. Any other information that is important to understand about this fallback

src/scanners/machine_uuid_scanner.go
ori-amizur: Doesn't really matter. Just to verify that the same order is kept.

src/scanners/machine_uuid_scanner.go
ori-amizur: Changed solution

src/scanners/machine_uuid_scanner.go
ori-amizur: Changed solution

src/scanners/machine_uuid_scanner.go
ori-amizur: Changed solution

PR #339, opened 2022-04-25
MGMT-10097: Move subsystem tests to new agent protocol

tsorya:
[MGMT-10097](https://issues.redhat.com//browse/MGMT-10097): Move subsystem tests to new agent protocol
As now commands are stored in agent we should not change them for tests
we added 2 scripts that will override nsenter and podman commands
We need it as in susbsystem we can't run nsenter and we can't run podman
Nsenter and podman scripts will drop nsenter and podman part from the
command and will allow running command itself without need to change it
in the code


src/commands/actions/image_availability_cmd.go
```diff
@@ -1,15 +1,16 @@
...
-	"github.com/alessio/shellescape"
 	"github.com/openshift/assisted-installer-agent/src/config"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/openshift/assisted-installer-agent/src/container_image_availability"
 	"github.com/openshift/assisted-service/models"
+	"golang.org/x/sync/semaphore"
 )
 
+var sem = semaphore.NewWeighted(1)
```
ori-amizur, line 12: Shouldn't we have semaphore per image?

src/commands/actions/image_availability_cmd_test.go
```diff
@@ -20,19 +18,8 @@ var _ = Describe("image availability", func() {
 	})
 
 	It("image availability", func() {
-		action, err := New(&config.AgentConfig{}, models.StepTypeContainerImageAvailability, []string{param})
+		_, err := New(&config.AgentConfig{}, models.StepTypeContainerImageAvailability, []string{param})
```
ori-amizur, line 21: What are we testing here ?

src/commands/step_processor.go
```diff
@@ -237,7 +235,7 @@ func (s *stepSession) processSingleSession() (int64, string) {
 			s.Logger().WithError(err).Errorf("User is forbidden to perform the operation, going to sleep forever")
 			return -1, ""
 		default:
-			s.Logger().Warnf("Could not query next steps: %s", getErrorMessage(err))
+			s.Logger().WithError(err).Warnf("Could not query next steps: %s", getErrorMessage(err))
```
ori-amizur, line 238: Doesn't getErrorMessage bring the error already ?

src/config/agent_config.go
```diff
@@ -53,10 +53,16 @@ func ProcessArgs() *AgentConfig {
 		log.Warnf("Agent Authentication Token not set")
 	}
 
+	log.Errorf("8888888888888888888888888888888 %s", ret.AgentVersion)
```
ori-amizur, line 56: ??

src/config/agent_config.go
```diff
@@ -53,10 +53,16 @@ func ProcessArgs() *AgentConfig {
...
+
 	// When given <image_url>:<tag> format, DiscoveryAgentVersion should point to the image tag.
 	// In a case of multiple delimiters, grab the rightmost slice.
 	// Otherwise, we leave the agent-version str intact.
 	agentVersionTag := strings.Split(ret.AgentVersion, agentVersionTagDelimiter)
 	ret.DiscoveryAgentVersion = agentVersionTag[len(agentVersionTag)-1]
+
+	log.Errorf("99999999999999999999999999999999999 %s", ret.AgentVersion)
```
ori-amizur, line 64: ??

subsystem/podman
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python3
```
ori-amizur, line 1: Can you just call it with different name here ? When the file is copied to the image, call it podman
Just not to be confused with the real podman

src/commands/actions/image_availability_cmd.go
tsorya: nope, this leaves the logic we currently have. Run only signle instance of image availability command

src/commands/step_processor.go
tsorya: this is good question, it is more like why we have getErrorMessage at all if we have withError functionality ?

src/config/agent_config.go
tsorya: mistake

PR #342, opened 2022-04-27
NO-ISSUE: in ci pulling image can be too quick and we can't validate 2 running image availability commands

tsorya:
NO-ISSUE: in ci pulling image can be too quick and we can't validate 2 running image availability commands


omertuc:
Maybe we can pick a bigger image instead of deleting the test?

tsorya:
@omertuc there is unitest that checks it, not a big deal

omertuc:
> @omertuc there is unitest that checks it, not a big deal

ack

PR #350, opened 2022-05-06
MGMT-9379: Validation should allow HA mode to be empty in parameters.

paul-maidment:
It used to be the case that a day 2 worker could only be added to an HA
cluster. We are making this change to allow the addition of a day 2
worker to a non HA cluster.

The condition [1] was blocking this.

That validation ensures that Day 1 SNO clusters will only accept Master
or Bootsrap roles for the host being installed.

This change is to allow omission of the high availability parameter in
the cases where the host role is "worker"

1: If the installer parameters received from the service do not contain
   a High Availability Mode parameter, do not pass this to the install
command.

2: Ensure that we bump the assisted-service models library so that
   validation will permit the ommission of the high availability mode
parameter.

[1] https://github.com/openshift/assisted-installer/blob/d67d2843d0e757ef5643e6c873b49e61ef83a0ea/src/config/config.go#L93-L99


src/commands/actions/install_cmd_test.go
```diff
@@ -100,13 +109,72 @@ var _ = Describe("installer test", func() {
...
 			"--agent-image quay.io/edge-infrastructure/assisted-installer-agent:latest " +
+			"--high-availability-mode Full " +
 			"--mco-image quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:3c30f115dc95c3fef94ea5185f386aa1af8a4b5f07ce8f41a17007d54004e1c4 " +
 			"--must-gather-image '{\"cnv\":\"registry.redhat.io/container-native-virtualization/cnv-must-gather-rhel8:v2.6.5\"," +
 			"\"lso\":\"registry.redhat.io/openshift4/ose-local-storage-mustgather-rhel8\",\"ocp\":\"quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:3c30f115dc95c3fef94ea5185f386aa1af8a4b5f07ce8f41a17007d54004e1c4\"," +
 			"\"ocs\":\"registry.redhat.io/ocs4/ocs-must-gather-rhel8\"}' --openshift-version 4.9.24 --insecure --check-cluster-version --installer-args '[\"--append-karg\",\"ip=ens3:dhcp\"]'"))
 	})
+
+	It("ha mode is nil, parameter should be omitted", func() {
+		args = getInstallCommandRequest()
```
omertuc, line 122: This line looks like it can be safely omitted, as it's already done in line 83

src/commands/actions/install_cmd_test.go
```diff
@@ -100,13 +109,72 @@ var _ = Describe("installer test", func() {
...
+		validationError := action.Validate()
+		Expect(validationError).NotTo(HaveOccurred())
+
+		args := action.Args()
+		argsAsString := strings.Join(args, " ")
+		Expect(argsAsString).NotTo(ContainSubstring("--high-availability-mode"))
+	})
+
+	It("ha mode is empty string, parameter should be omitted", func() {
+		args = getInstallCommandRequest()
```
omertuc, line 136: This line looks like it can be safely omitted, as it's already done in line 83

src/commands/actions/install_cmd_test.go
```diff
@@ -100,13 +109,72 @@ var _ = Describe("installer test", func() {
...
+		validationError := action.Validate()
+		Expect(validationError).NotTo(HaveOccurred())
+
+		args := action.Args()
+		argsAsString := strings.Join(args, " ")
+		Expect(argsAsString).To(ContainSubstring("--high-availability-mode Full"))
+	})
+
+	It("ha mode is None, command line parameter should indicate this", func() {
+		args = getInstallCommandRequest()
```
omertuc, line 165: This line looks like it can be safely omitted, as it's already done in line 83

src/commands/actions/install_cmd_test.go
```diff
@@ -70,9 +60,28 @@ var _ = Describe("installer test", func() {
 			OpenshiftVersion: "4.9.24",
 			Role:             &role,
 		}
+	}
+
+	getParams := func(installCommandRequest models.InstallCmdRequest) string {
```
omertuc, line 65: Can we find a better name for params / args / this function? The distinction is not entirely clear

src/commands/actions/install_cmd.go
```diff
@@ -93,11 +93,14 @@ func (a *install) getFullInstallerCommand() string {
 		"--host-id", string(*a.installParams.HostID),
 		"--boot-device", swag.StringValue(a.installParams.BootDevice),
 		"--url", a.agentConfig.TargetURL,
-		"--high-availability-mode", swag.StringValue(a.installParams.HighAvailabilityMode),
 		"--controller-image", swag.StringValue(a.installParams.ControllerImage),
 		"--agent-image", a.agentConfig.AgentVersion,
 	}
 
+	if a.installParams.HighAvailabilityMode != nil && *a.installParams.HighAvailabilityMode != "" {
```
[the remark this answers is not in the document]
paul-maidment: The idea here is that if HighAvailability mode is unspecified either through being absent (set to nil) or through being empty, then the installer command line will not generate a flag for --high-availability mode.

This is part of a larger set of changes to omit this parameter when running the installer for hosts that have a role of "worker".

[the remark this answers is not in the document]
paul-maidment: I'm not sure entirely what your question around this is, could you clarify?

PR #352, opened 2022-05-11
NO-ISSUE: remove redundant ok-to-test label 

osherdp:
We no longer need the ok-to-test label, as dependabot is a
trusted-app.

/cc @eliorerz @filanov

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,7 +12,6 @@ updates:
     labels:
       - "approved"
       - "lgtm"
-      - "ok-to-test"
       - "go"
       - "dependencies"
     commit-message:


omertuc:
Why do we consider dependabot as "automatic"? it's up to us to /lgtm it, so it's not really automatic. It's nice to have its PRs remind us that our Dockerfiles are outdated

osherdp:
> Why do we consider dependabot as "automatic"? it's up to us to /lgtm it, so it's not really automatic. It's nice to have its PRs remind us that our Dockerfiles are outdated

For most cases (where it didn't need to update PR for a just newer dependency version) it will lgtm automatically and no member of us is needed for the complete merge of those PRs. So it's really is automatic in 90% of the times

In addition, the proposed changes are misleading. You don't only need to update the Dockerfiles, but also ``go.mod`` (for the same repo, even not mentioning what's need to be done in openshift/release)

Lastly, it would be nice getting a reminder indeed but it does not do that as well (https://github.com/openshift/assisted-installer-agent/pull/345 has been opened 9 days ago and didn't trigger any discussion)

omertuc:
>For most cases (where it didn't need to update PR for a just newer dependency version) it will lgtm automatically and no member of us is needed for the complete merge of those PRs. So it's really is automatic in 90% of the times

That can be fixed

>In addition, the proposed changes are misleading. You don't only need to update the Dockerfiles, but also go.mod (for the same repo, even not mentioning what's need to be done in openshift/release)

You don't *need* to update the go.mod file, the go.mod file version number is only a "minimum version", you can compile a 1.17 go mod file with a 1.18 toolchain completely fine. If anything you don't even want to update the go.mod file, as that would force everybody to update their local toolchains to the new minimum version you set, which would actually be annoying

> Lastly, it would be nice getting a reminder indeed but it does not do that as well (https://github.com/openshift/assisted-installer-agent/pull/345 has been opened 9 days ago and didn't trigger any discussion)

This sounds like a people process problem, ideally we need people in the team to own those dependabot PRs and act on them

osherdp:
> I think that if we fix the first quote in this comment we can leave the dependabot PRs for Dockerfiles, they'll be harmless reminders

what exactly should get fixed?
Not so sure if I understand

omertuc:
>what exactly should get fixed?

This:

>it will lgtm automatically and no member of us is needed for the complete merge of those PRs

We shouldn't lgtm automatically

osherdp:
You think so?
I actually prefer processes that can be automated and require less work on our side. A human reviewer will probably not do anything different than the "automatic" one for 99% of the cases, meaning it will only add more work and burden on most cases with not much added benefit

Of course if we've had lots of cases where we upgraded a golang dependency and it introduced a bug that was not caught on CI then I would have been more cautious. But I don't know any case like that

osherdp:
For the docker changes the "downstream change needed" label guards those changes, so even automatically lgtming shouldn't be a problem

omertuc:
>You think so?
I actually prefer processes that can be automated and require less work on our side. A human reviewer will probably not do anything different than the "automatic" one for 99% of the cases, meaning it will only add more work and burden on most cases with not much added benefit
>
>Of course if we've had lots of cases where we upgraded a golang dependency and it introduced a bug that was not caught on CI then I would have been more cautious. But I don't know any case like that

That's a fair approach. I personally believe we shouldn't merge anything without human intervention / review. Just like we don't trust ourselves and we force ourselves to review each other's PRs, we shouldn't trust dependabot changes just as well.

> For the docker changes the "downstream change needed" label guards those changes, so even automatically lgtming shouldn't be a problem

Alright so that means there's no good reason for #352 and similar PRs, if it's not automatically merged, it just serves as a nice notification

osherdp:
Yeah, I can leave those PRs as an extra reminder for us (it should ultimately notify about other things, not just the golang images)
I'll change the PR to just remove the ``ok-to-test`` label as it's redundant

PR #360, opened 2022-05-25
MGMT-9175: Pass all interfaces to the service

CrystalChun:
[MGMT-9175](https://issues.redhat.com/browse/MGMT-9175)
Based on epic [MGMT-10087](https://issues.redhat.com/browse/MGMT-10087)
This change removes filtering interfaces and sends
all interfaces to the assisted-service along with the type
of interface it is.

/cc @carbonin 
/cc @ori-amizur 

/hold Do not merge until https://github.com/openshift/assisted-service/pull/3855 for [MGMT-10473](https://issues.redhat.com/browse/MGMT-10473) goes in


carbonin:
Looks good, but I think you still need to pull in the new assisted service code now that the other PR is merged.

src/util/network_interface.go
```diff
@@ -47,6 +48,21 @@ func (n *NetworkInterface) Addrs() ([]net.Addr, error) {
...
 }
 
+func (n *NetworkInterface) Type() string {
+	evaledPath, err := n.dependencies.EvalSymlinks(fmt.Sprintf("/sys/class/net/%s", n.netInterface.Name))
+	if err != nil {
+		logrus.WithError(err).Warnf("Could not determine if interface %s is physical", n.netInterface.Name)
+		return err.Error()
+	} else if !strings.Contains(evaledPath, "/virtual/") {
+		return "physical"
+	}
```
carbonin, line 58: Couldn't this all be:
```
if n.IsPhysical {
  return "physical"
}
```

src/util/network_interface.go
```diff
@@ -47,6 +48,21 @@ func (n *NetworkInterface) Addrs() ([]net.Addr, error) {
...
+	link, err := n.dependencies.LinkByName(n.netInterface.Name)
+	if err != nil {
+		return err.Error()
```
carbonin, line 61: If you can return an error, then this function should return an error type.

Doing it this way would lead to us putting the error string as the interface type in inventory, but that's definitely not what we want, right?

src/util/network_interface.go
```diff
@@ -47,6 +48,21 @@ func (n *NetworkInterface) Addrs() ([]net.Addr, error) {
...
+	}
+	return link.Type()
```
carbonin, line 63: Why not return this always? What are the options that get grouped under "physical"?

CrystalChun: The value of `link.Type()` for both physical and virtual interfaces is "device" unless it's a bridge, vlan, etc.

src/util/network_interface.go
CrystalChun: Good point, thank you!

src/util/network_interface.go
CrystalChun: That's right! Thanks for catching that. Modified it!

src/inventory/interfaces.go
```diff
@@ -122,6 +120,12 @@ func (i *interfaces) getInterfaces() []*models.Interface {
 				rec.IPV6Addresses = append(rec.IPV6Addresses, addrStr)
 			}
 		}
+		interfaceType, err := in.Type()
+		if err != nil {
+			logrus.WithError(err).Warnf("Retrieiving interface type for %s", in.Name())
+			continue
+		}
+		rec.Type = interfaceType
```
carbonin, line 128: `err` should already be defined in this scope so you don't need this extra assignment.

Line 123 can just be `rec.Type, err = in.Type()`

carbonin (commented):
From a previous comment:

> The value of link.Type() for both physical and virtual interfaces is "device" unless it's a bridge, vlan, etc.

What does this mean for the range of values we could get back? Can/should we define an enumeration of some kind of the values we expect for this? Maybe this will be a part of the service code that checks the difference between what we consider "physical" and whatever else?

src/inventory/interfaces.go
CrystalChun: Oh! Thank you, modified it

CrystalChun:
> What does this mean for the range of values we could get back? Can/should we define an enumeration of some kind of the values we expect for this? Maybe this will be a part of the service code that checks the difference between what we consider "physical" and whatever else?

The values we could get back are all within this library https://github.com/openshift/assisted-installer-agent/blob/ce3cdbfd27dd2939dd52d1e46ab4242f56357175/vendor/github.com/vishvananda/netlink/link.go
There seems to be a lot of different types - is it worth it to enumerate them all? 

I would think the service only really cares about whether or not to filter actually, so we could describe the originally filtered ones (physical, vlan, bonds) as one value (default) and the others as another value

carbonin:
> There seems to be a lot of different types - is it worth it to enumerate them all?

Probably not for now because we haven't had to care so far. I just wanted to make sure we had a way to understand what values might be there and what they mean.

> I would think the service only really cares about whether or not to filter actually, so we could describe the originally filtered ones (physical, vlan, bonds) as one value (default) and the others as another value

I think we're trying as best we can to leave the decision up to the service. The idea is that we don't want to have to change the agent if we change how we want the filtering to work so if we're grouping different things together in the agent based on what we think should be filtered then it's not the service making that decision anymore.

> Sorry I didn't think this through earlier - would there even be a use-case where having the actual type of interface be useful as opposed to just a boolean saying whether or not it was filtered before?

No it's fine. I think we did consider this. It's the same reason that we try to avoid booleans in most APIs. It's possible (likely even) that any API that starts as a bool could eventually need to be expanded. So right now, no, there's no use case for actually needing the type, but there might be in the future so we should start the API as a string even if we're not fully utilizing the value.

Looking at the code again this is a good first pass that keeps everything working pretty much as it was before.

src/inventory/interfaces.go
```diff
@@ -88,10 +88,8 @@ func (i *interfaces) getInterfaces() []*models.Interface {
 		logrus.WithError(err).Warnf("Retrieving interfaces")
 		return ret
 	}
+
 	for _, in := range ins {
-		if !(in.IsPhysical() || in.IsBonding() || in.IsVlan()) {
```
carbonin, line 92: Do we use `IsBonding` or `IsVlan` anywhere else? Can they be removed if not?

CrystalChun: Oh yes we use it in the [connectivity_check](https://github.com/openshift/assisted-installer-agent/blob/master/src/connectivity_check/connectivity_check.go#L30) to avoid running the connectivity check on virtual interfaces.

carbonin: hmmm is that something that will need to change as a part of this effort?
Probably not in this PR though.

CrystalChun: Maybe eventually, but as a first iteration I think we agreed to only check these interfaces instead of all of them since the connectivity_check is the most demanding ([notes from kick-off meeting](https://docs.google.com/document/d/1LtwhA0jkR3jlatQiEtyE-_dbrn47XaMDnSki0rpTPnc/edit?disco=AAAAY1wZ5AU)).

carbonin:
> Gotcha! Then would it be a good idea to add an enumeration for this type on the Interface model? That way both the agent and service share the same enumeration

I think for this pass just a constant in the service for `physical` would be enough. We can add something else if/when we want to do more complicated logic I think.

carbonin (approved):

PR #370, opened 2022-06-06
MGMT-10705: Return multipath device types

avishayt:


src/inventory/disks.go
```diff
@@ -189,6 +197,19 @@ func (d *disks) isMultipath(disk *ghw.Disk) bool {
 	return false
 }
 
+func (d *disks) getHolders(diskName string) string {
+	dir := fmt.Sprintf("/sys/block/%s/holders", diskName)
+	files, err := d.dependencies.ReadDir(dir)
+	if err != nil || len(files) == 0 {
+		return ""
```
tsorya, line 204: Should not we log err? though not sure why it should happened but better to have some log in any case

avishayt: Fixed

PR #372, opened 2022-06-07
ART-4057: Work around parsing issue on ART's tooling

thiagoalessio:
ART's tooling makes use of [bashlex](https://github.com/idank/bashlex) to parse `RUN` commands found in Dockerfiles.
But for whatever reason it is failing to parse the following instruction:
```
RUN if [[ "$(arch)" == "x86_64" ]]; then dnf install -y biosdevname && dnf clean all; fi
```

```
bashlex.errors.ParsingError: unexpected token '"$(arch)"' (position 6)
```

At some point the root cause must be addressed, but given the time constraints, I'd like to suggest the following change for now, just to make our parser happy for the time being and proceed with the onboarding of this image into OCP 4.11.


celebdor:
/retitle  ART-4057: Work around parsing issue on ART's tooling

osherdp:
@thiagoalessio can you fix commit message?
for example ``ART-4057: Work around parsing issue on ART's tooling``

osherdp (approved):

PR #374, opened 2022-06-08
ART-4057: Only install `dmidecode` on x86_64 and aarch64

thiagoalessio:
dmidecode package is not available for ppc64le and s390x arches.


osherdp:
@ori-amizur can you take a look at this?
what will it mean to not have the ``dmidecode`` binary?

omertuc:
We seem to have fallback, so it should be fine:

https://github.com/openshift/assisted-installer-agent/blob/bf66b232529b5023f29eca554f9ed6c51b2a6044/src/inventory/memory.go#L48-L51

omertuc:
>So why installing it in the first place?

I think it has something to do with the operating system reporting a little less RAM than you have in practice because some of it is reserved for the OS. So we make a best effort attempt to read the size via DMI directly rather than going through the OS, and if that fails, we fallback to ghw which I believe reads from the OS

PR #375, opened 2022-06-09
MGMT-10748: Use DriveType definitions from swagger

avishayt:
Rather than returning strings for disk DriveType, use the enum defined
in the swagger.  Also return LVM logical volumes, but still do not allow
installation on them.


src/inventory/disks.go
```diff
@@ -192,12 +192,20 @@ func (d *disks) isMultipath(disk *ghw.Disk) bool {
...
 	}
 	return false
 }
 
+func (d *disks) isMultipath(disk *ghw.Disk) bool {
+	return d.dmUUIDHasPrefix(disk, "mpath-")
+}
+
+func (d *disks) isLVM(disk *ghw.Disk) bool {
+	return d.dmUUIDHasPrefix(disk, "LVM-")
```
tsorya, line 206: i think we should move to lower or upper case but to be consistent, what do you think?

PR #380, opened 2022-06-16
NO_ISSUE: fix mock generation.

michaellevy101:
go generate command failed when the package uses the generated code
The make generate target deletes the old mocks before regenerating them
mocks.go file used util.MockInterface type so after the deletion the compilation failed which fails the mock generation for this package
Using mock.Mock embedded object instead

### DIFF
--- a/src/inventory/interfaces_test.go
+++ b/src/inventory/interfaces_test.go
@@ -39,7 +39,7 @@ var _ = Describe("Interfaces", func() {
 	})
 
 	It("Single result", func() {
-		interfaceMock := util.NewFilledInterfaceMock(1500, "eth0", "f8:75:a4:a4:00:fe", net.FlagBroadcast|net.FlagUp, []string{"10.0.0.18/24", "fe80::d832:8def:dd51:3527/128", "de90::d832:8def:dd51:3527/128"}, true, false, false, 1000, "physical")
+		interfaceMock := newMockInterface(1500, "eth0", "f8:75:a4:a4:00:fe", net.FlagBroadcast|net.FlagUp, []string{"10.0.0.18/24", "fe80::d832:8def:dd51:3527/128", "de90::d832:8def:dd51:3527/128"}, 1000, "physical")
 		dependencies.On("Interfaces").Return([]util.Interface{interfaceMock}, nil).Once()
 		dependencies.On("Execute", "biosdevname", "-i", "eth0").Return("em2", "", 0).Once()
 		dependencies.On("ReadFile", "/sys/class/net/eth0/carrier").Return([]byte("1\n"), nil).Once()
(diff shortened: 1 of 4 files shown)


michaellevy101:
@omertuc

src/inventory/interfaces_test.go
```diff
@@ -195,3 +195,9 @@ var _ = Describe("Interfaces", func() {
 		}
 	})
 })
+
+func newInterface(mtu int, name string, macAddr string, flags net.Flags, addrs []string, speedMbps int64, interfaceType string) *util.MockInterface {
```
osherdp, line 199: ``newInterface`` seems like the real deal, maybe:
```suggestion
func newMockInterface(mtu int, name string, macAddr string, flags net.Flags, addrs []string, speedMbps int64, interfaceType string) *util.MockInterface {
```

src/util/test_utils.go
```diff
@@ -0,0 +1,33 @@
...
+	for _, a := range addrs {
+		ret = append(ret, str2Addr(a))
+	}
+	return ret
+}
+
+func str2Addr(addrStr string) net.Addr {
+	ip, ipnet, err := net.ParseCIDR(addrStr)
+	if err != nil {
+		return &net.IPNet{}
```
osherdp, line 30: seems weird to return an empty value when tester entered malformed values
it's harder to debug those cases

src/util/test_utils.go
```diff
@@ -0,0 +1,33 @@
...
+
+func toAddresses(addrs []string) []net.Addr {
+	ret := make([]net.Addr, 0)
+	for _, a := range addrs {
+		ret = append(ret, str2Addr(a))
+	}
+	return ret
+}
+
+func str2Addr(addrStr string) net.Addr {
```
osherdp, line 27:
```suggestion
func parseAddress(addrStr string) net.Addr {
```

src/util/test_utils.go
```diff
@@ -0,0 +1,33 @@
...
+	mock.On("MTU").Return(mtu)
+	hwAddr, _ := net.ParseMAC(macAddr)
+	mock.On("HardwareAddr").Return(hwAddr)
+	mock.On("Flags").Return(flags)
+	mock.On("Addrs").Return(toAddresses(addrs), nil).Once()
+	mock.On("SpeedMbps").Return(speedMbps)
+	mock.On("Type").Return(interfaceType, nil).Once()
+}
+
+func toAddresses(addrs []string) []net.Addr {
```
osherdp, line 19:
```suggestion
func parseAddresses(addrs []string) []net.Addr {
```

src/inventory/interfaces_test.go
```diff
@@ -73,11 +73,11 @@ var _ = Describe("Interfaces", func() {
...
-			util.NewFilledInterfaceMock(1400, "eth2.10", "f8:75:a4:a4:00:fc", net.FlagBroadcast|net.FlagUp, []string{"10.0.0.25/24", "192.168.6.14/20", "fe80::d832:8def:dd51:3520/125", "de90::d832:8def:dd51:3520/125"}, false, false, true, -1, "vlan"),
+			newInterface(1500, "eth0", "f8:75:a4:a4:00:fe", net.FlagBroadcast|net.FlagUp, []string{"10.0.0.18/24", "192.168.6.7/20", "fe80::d832:8def:dd51:3527/128", "de90::d832:8def:dd51:3527/128"}, 100, "physical"),
+			newInterface(1400, "eth1", "f8:75:a4:a4:00:ff", net.FlagBroadcast|net.FlagLoopback, []string{"10.0.0.19/24", "192.168.6.8/20", "fe80::d832:8def:dd51:3528/127", "de90::d832:8def:dd51:3528/127"}, 10, "physical"),
+			newInterface(1400, "eth2", "f8:75:a4:a4:00:ff", net.FlagBroadcast|net.FlagLoopback, []string{"10.0.0.20/24", "192.168.6.9/20", "fe80::d832:8def:dd51:3529/126", "de90::d832:8def:dd51:3529/126"}, 5, ""),
+			newInterface(1400, "bond0", "f8:75:a4:a4:00:fd", net.FlagBroadcast|net.FlagUp, []string{"10.0.0.21/24", "192.168.6.10/20", "fe80::d832:8def:dd51:3529/125", "de90::d832:8def:dd51:3529/125"}, -1, "bond"),
```
osherdp, line 79: seems like the intention was to have it as a bond address (with the ``isBonding=true`` parameter), same goes with the physical address 
it's worth returning those parameters so that we'll have the same "coverage" as before

src/util/mocks.go
```diff
@@ -1,42 +1,10 @@
...
-}
-
-func str2Addr(addrStr string) net.Addr {
-	ip, ipnet, err := net.ParseCIDR(addrStr)
-	if err != nil {
-		return &net.IPNet{}
-	}
-	return &net.IPNet{IP: ip, Mask: ipnet.Mask}
-}
+}
```
osherdp, line 10: can you add a blank line at the end?

src/scanners/machine_uuid_scanner_test.go
```diff
@@ -118,3 +118,9 @@ func TestSubsystem(t *testing.T) {
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "Scanner unit tests")
 }
+
+func newInterface(mtu int, name string, macAddr string, flags net.Flags, addrs []string, speedMbps int64, interfaceType string) *agentutils.MockInterface {
+	interfaceMock := agentutils.MockInterface{}
```
osherdp, line 123: Is it possible to have one ``newInterface`` implementation? (in ``test_utils.go`` probably)

osherdp (changes requested):

osherdp:
Seems like we have 2 other problems with code generation in this repo:
1. we're replacing code in ``vendor`` dir as well
2. we have no CI job to validate code generation

but those can be handled regardless of this PR

src/util/test_utils.go
michaellevy101: It was copied from another file. Not sure if I understand the logic correctly.
It is only a technical PR to prevent the mock generation from failing

src/util/test_utils.go
michaellevy101: Please read my previous comment

michaellevy101: please read my previous comment. I will change the function name

src/inventory/interfaces_test.go
michaellevy101: The purpose of this PR is not logic, but to fix the mock generation.
Anyhow, if I understand correctly, we replace the flags with a string value for interfaceType.

src/scanners/machine_uuid_scanner_test.go
michaellevy101: unfortunately i don't know. you can't import a test file from another....

michaellevy101: maybe adding a separated test package but it isn't so clear

osherdp: wait, not sure I understand
you import ``FillInterfaceMock`` from ``test_utils.go``. why not import ``newMockInterface`` from the same ``test_utils.go``?

src/inventory/interfaces_test.go
osherdp: I was referring to the missing ``isVlan`` usage, which should be considered as logic-change (that we don't want to have in this PR)
But now I'm seeing it was unused even before, so it should probably be okay to remove

osherdp (changes requested):

michaellevy101:
https://github.com/openshift/assisted-installer-agent/pull/380#discussion_r901064264
@osherdp During regenerating, we delete all generated resources, so any files using a generated type will not compile.
A compilation error prevents the mockery tool from generating the resources. 
Specifically. The mocks.go file uses MockInterface, so mockery is unable to regenerate it because mocks.go has a compilation problem after deleting MockInterface

osherdp:
> [#380 (comment)](https://github.com/openshift/assisted-installer-agent/pull/380#discussion_r901064264) @osherdp During regenerating, we delete all generated resources, so any files using a generated type will not compile. A compilation error prevents the mockery tool from generating the resources. Specifically. The mocks.go file uses MockInterface, so mockery is unable to regenerate it because mocks.go has a compilation problem after deleting MockInterface This problem can be solved by using MockInterface type by the test and having mocks.go deal with the mock.Mock objet that is embedded inside the generate type. "go build" command ignores test files

ack, it's now more clear
thanks!

osherdp (approved):

PR #386, opened 2022-06-26
MGMT-10936: Cleanup agent code for commands that are not run as with podman run anymore

tsorya:
[MGMT-10936](https://issues.redhat.com//browse/MGMT-10936): Cleanup agent code for commands that are not run as with podman run anymore

### DIFF
--- a/Dockerfile.ocp
+++ b/Dockerfile.ocp
@@ -10,17 +10,11 @@ FROM registry.ci.openshift.org/ocp/4.11:base
 LABEL io.openshift.release.operator=true
 
 COPY --from=builder /go/src/github.com/openshift/assisted-installer-agent/build/agent /usr/bin/agent
-COPY --from=builder /go/src/github.com/openshift/assisted-installer-agent/build/connectivity_check /usr/bin/connectivity_check
 COPY --from=builder /go/src/github.com/openshift/assisted-installer-agent/build/inventory /usr/bin/inventory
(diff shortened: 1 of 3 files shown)


Dockerfile.ocp
```diff
@@ -10,17 +10,11 @@ FROM registry.ci.openshift.org/ocp/4.11:base
 LABEL io.openshift.release.operator=true
 
 COPY --from=builder /go/src/github.com/openshift/assisted-installer-agent/build/agent /usr/bin/agent
-COPY --from=builder /go/src/github.com/openshift/assisted-installer-agent/build/connectivity_check /usr/bin/connectivity_check
```
osherdp, line 13: if it hasn't been done already, it's worth to do those changes in D/S Dockerfiles prior to merging this PR

test.py
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env python
```
osherdp, line 1: what's this?
same question for the ``sys`` file

osherdp (changes requested):

test.py
tsorya: oops

Dockerfile.ocp
tsorya: not sure regarding prior to merging. They actually independent and we can merge this and then downstream. In any case first we should be sure we want to it and that everything works

osherdp (approved):

gamli75:
the DS changes were merged https://code.engineering.redhat.com/gerrit/c/assisted-installer-projects/+/417254

osherdp (approved):

PR #388, opened 2022-07-01
MGMT-10973: Copy coredns & keepalived static pod manifests in day-2 connectivity-check ignition

omertuc:
# Goal

During day-2 installations, we want the service to optionally perform
DNS validations when the worker attempts to join none-platform clusters.

# Issue

When the cluster is imported via our `.../v2/clusters/import` endpoint,
we have no way to know whether that cluster is a none-platform cluster
or cluster with managed networking (e.g. baremetal), so we have no way
to know whether we should perform the DNS validation or not. We wouldn't
like to have that validation on all the time, because it's unnecessarily
annoying for clusters with managed networking.

# Background

As part of our existing API connectivity-check, the service asks the
agent to download the worker.ign file from the to-be-joined-cluster's
machine-config-server, then send that back to the service.

# Solution

The contents of that file include information that will allow the
service to make an educated guess about whether the ignition originated
from a cluster with managed networking or not.

However, that ignition file is incomplete as currently we only include
necessary information in it (because that file is very big). The parts
of that file that we care in this case are currently missing, so we have
to modify the agent to include those files.

This commit does that. A follow-up commit will be done on the service to
actually check for the presence of these files in this ignition to make
the decision of whether the cluster is managed or not, and in turn turn
on the necessary validations if not.

# Ignition Files

The ignition files we currently include are:

```
/etc/kubernetes/manifests/coredns.yaml
/etc/kubernetes/manifests/keepalived.yaml
```

This is a hack - since we have no official way to know whether a worker
ignition file originated from a cluster with managed networking or not,
we instead rely on the presence of coredns and keepalived pod manifests
to indicate that. We only expect those to be present in clusters with
managed networking. To be a bit more robust, the service will consider
the presence of any one of them to mean that the cluster has managed
networking. This gives us better forwards compatibility if one of them
gets renamed / replaced with other technologies in future OCP versions.

Another way in which this is hacky is that users could manually create
static pods with the same name as part of their machine-configs, in
which case we would have a false-positive detection. But that is
admittedly very unlikely.

Hopefully we can negotiate with the relevant OCP teams to have a more
official, stable way to have this detection - like a magic empty file
placed somewhere in the ignition that we can check for the presence of.
Once we have such file, we can slowly deprecate this detection mechanism
and fully move to the new one by inspecting that file instead.

### DIFF
--- a/src/apivip_check/apivip_check.go
+++ b/src/apivip_check/apivip_check.go
@@ -10,13 +10,80 @@ import (
 	"net/http"
 
 	"github.com/coreos/ignition/v2/config/v3_2"
-	"github.com/coreos/ignition/v2/config/v3_2/types"
+	ignition_types "github.com/coreos/ignition/v2/config/v3_2/types"
 	"github.com/openshift/assisted-service/models"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
 const ignitionVersion string = "3.2.0"
+
+func getIgnitionFile(ignition *ignition_types.Config, path string) *ignition_types.File {
+	for _, file := range ignition.Storage.Files {
+		if file.Path == path {
+			return &file
+		}
+	}
+
+	return nil
+}
+
+func copyIgnitionFile(originalConfig *ignition_types.Config, filteredConfig *ignition_types.Config, path string) {
+	if file := getIgnitionFile(originalConfig, path); file != nil {
+		filteredConfig.Storage.Files = append(filteredConfig.Storage.Files, *file)
+	}
+}
+
+// copyIgnitionDiskEncryptionInformation copies disk-encryption related
+// information. See filterIgnition for more info.
+func copyIgnitionDiskEncryptionInformation(originalConfig *ignition_types.Config, filteredConfig *ignition_types.Config) {
+	filteredConfig.Storage.Luks = originalConfig.Storage.Luks
+}
+
+// copyIgnitionManagedNetworkIndications copies (see filterIgnition for more
+// info) ignition storage file entries that can be used by the service in order
+// to detect whether this worker ignition file originated from a cluster that
+// has managed networking or not. This helps the service, for example, to know
+// whether it should perform DNS validations for the day-2 hosts trying to join
+// that cluster or not (as managed network clusters can be joined without any
+// DNS configuration, they do not require such validations).
+func copyIgnitionManagedNetworkIndications(originalConfig *ignition_types.Config, filteredConfig *ignition_types.Config) {
+	// This is a hack - since we have no official way to know whether a worker
+	// ignition file originated from a cluster with managed networking or not,
+	// we instead rely on the presence of coredns and keepalived pod manifests
+	// to indicate that. We only expect those to be present in clusters with
+	// managed networking. To be a bit more robust, we consider the presence of
+	// any one of them to mean that the cluster has managed networking. This
+	// gives us better forwards compatibility if one of them gets renamed /
+	// replaced with other technologies in future OCP versions.
+	//
+	// Another way in which this is hacky is that users could manually create
+	// static pods with the same name as part of their machine-configs, in
+	// which case we would have a false-positive detection. But that is
+	// admittedly very unlikely.
+	//
+	// Hopefully we can negotiate with the relevant OCP teams to have a more
+	// official, stable way to have this detection - like a magic empty file
+	// placed somewhere in the ignition that we can check for the presence of.
+	// Once we have such file, we can slowly deprecate this detection mechanism
+	// and fully move to the new one by inspecting that file instead.
+	copyIgnitionFile(originalConfig, filteredConfig, "/etc/kubernetes/manifests/coredns.yaml")
+	copyIgnitionFile(originalConfig, filteredConfig, "/etc/kubernetes/manifests/keepalived.yaml")
+}
+
+// filterIgnition removes unnecessary sections of the ignition config, leaving
+// only those needed by the service. We do this because the ignition config
+// tends to be quite large.
+func filterIgnition(originalConfig *ignition_types.Config) *ignition_types.Config {
+	filteredConfig := ignition_types.Config{}
+
+	copyIgnitionDiskEncryptionInformation(originalConfig, &filteredConfig)
+	copyIgnitionManagedNetworkIndications(originalConfig, &filteredConfig)
+
+	// NOTE: add here any additional objects from the config (when needed by the service)
+
+	return &filteredConfig
+}
 
 func CheckAPIConnectivity(checkAPIRequestStr string, log logrus.FieldLogger) (stdout string, stderr string, exitCode int) {
 	var checkAPIRequest models.APIVipConnectivityRequest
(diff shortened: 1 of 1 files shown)


avishayt:
> We wouldn't like to have that validation on all the time, because it's unnecessarily annoying for clusters with managed networking.

Why is it annoying?

romfreiman:
Trying to summarize the flow:
1. Try to access api fqdn.... (being generated by guessing from what we get from OCM) to verify 1st dns access (by pulling the ign?)
2. If api fqdn is not resolvable, suggest the user to input an ip
3. repeat 1 by ip provided in 2
4. Once ign is downloaded, verify keepalived/coredns.
5. If 4 exists, CMN, continue
6. If 4 does not exist, UMN, run full dns validtaion (api, api-int, apps)

Did I miss anything?

tsorya:
IMO agent should not do any decisions

eranco74:
I'm a bit confused, why do we care if the ignition file originated from a cluster with managed networking or not?
we need DNS to the cluster API (and api-int) to work. and we have a validation that checks it, what am I missing?

omertuc:
>We wouldn't like to have that validation on all the time, because it's unnecessarily annoying for clusters with managed networking.

DNS validations in day-2 mean we force the user to set up DNS configuration for the cluster domains of the cluster they want to join, something that they didn't have to care about in day-1 and have no reason to care about in day-2 when the day-1 cluster has managed networking 

All they have to do today, in managed networking clusters, is give us the IP address in a UI prompt and that is enough because it allows the worker to download the worker ignition, and in that ignition there's an instance of coredns that takes care of setting up that DNS configuration internally within the node.

But if the cluster-to-be-joined doesn't have managed networking (e.g. none-platform, sno) then coredns won't be there to help the node finish joining of the cluster. The worker can download the ignition thanks to the IP we got in the UI prompt, but after that it will have no DNS configuration to allow it to continue the process.

To summarize, enabling DNS validations in day-2 across the board will make our current user experience worse and require users to perform more steps when they don't have to, in some situations.

We can actually do better here, we can push our own instance of coredns with a corefile containing all the required DNS entries so the user has a seamless experience regardless of whether the cluster-to-be-joined has managed networking or not, but this will be a much more intrusive and complicated feature than simply enabling validations for now to make sure the user manually sets up DNS correctly

omertuc:
>IMO agent should not do any decisions

Where does it make any decisions? All I did was put back some files that we previously removed from the ignition because it was too big

omertuc:
>I'm a bit confused, why do we care if the ignition file originated from a cluster with managed networking or not?
we need DNS to the cluster API (and api-int) to work. and we have a validation that checks it, what am I missing?

See https://github.com/openshift/assisted-installer-agent/pull/388#issuecomment-1173040910

omertuc:
> Trying to summarize the flow:
> 
> 1. Try to access api fqdn.... (being generated by guessing from what we get from OCM) to verify 1st dns access (by pulling the ign?)
> 2. If api fqdn is not resolvable, suggest the user to input an ip
> 3. repeat 1 by ip provided in 2
> 4. Once ign is downloaded, verify keepalived/coredns.
> 5. If 4 exists, CMN, continue
> 6. If 4 does not exist, UMN, run full dns validtaion (api, api-int, apps)
> 
> Did I miss anything?

1. Right, but we don't really verify "DNS" directly here, we just try to pull the ignition, we call it API connectivity check
2. If, for any reason, we failed to get the ignition, then we ask the user to input an IP. As mentioned in, (1), it's not specifically about DNS, but it's usually because of DNS, yes
3. Correct
4. Correct
5. Correct
6. Correct

src/apivip_check/apivip_check.go
```diff
@@ -10,14 +10,111 @@ import (
...
+	// static pods with the same name as part of their machine-configs, in
+	// which case we would have a false-positive detection. But that is
+	// admittedly very unlikely.
+	//
+	// Hopefully we can negotiate with the relevant OCP teams to have a more
+	// official, stable way to have this detection - like a magic empty file
+	// placed somewhere in the ignition that we can check for the presence of.
+	// Once we have such file, we can slowly deprecate this detection mechanism
+	// and fully move to the new one by inspecting that file instead.
+	copyIgnitionFile(originalConfig, filteredConfig, "/etc/kubernetes/manifests/coredns.yaml")
```
tsorya, line 98: i assume it is enough to check coredns

eranco74:
> * Right, but we don't really verify "DNS" directly here, we just try to pull the ignition, we call it API connectivity check
I think we changed it to ignition downloadable:
https://github.com/openshift/assisted-service/pull/3318

src/apivip_check/apivip_check.go
```diff
@@ -10,14 +10,111 @@ import (
...
+// used by the service to know whether this filteredConfig was filtered by a
+// recent version of the agent or not.
+//
+// Since the *absence* of the coredns and keepalived files copied by
+// copyIgnitionManagedNetworkIndications indicates that the cluster is running
+// a non-managed-network cluster, the service has no way to know whether the
+// absence is due to that ignition file being processed by an old agent or due
+// to an actual absence. To overcome this ambiguity, we add this dummy file to
+// the ignition which can help the service tell those two situations apart.
+func agentVersionSkewHack(filteredConfig *ignition_types.Config) {
```
tsorya, line 53: I am not sure we should add this logic, especially cause we are already working on agent upgrade

src/apivip_check/apivip_check.go
```diff
@@ -10,14 +10,111 @@ import (
...
+}
+
+// filterIgnition removes unnecessary sections of the ignition config, leaving
+// only those needed by the service. We do this because the ignition config
+// tends to be quite large.
+func filterIgnition(originalConfig *ignition_types.Config) *ignition_types.Config {
+	filteredConfig := ignition_types.Config{}
+
+	copyIgnitionDiskEncryptionInformation(originalConfig, &filteredConfig)
+	copyIgnitionManagedNetworkIndications(originalConfig, &filteredConfig)
```
tsorya, line 111: can't we just check if specific file exist in ignition? and return true/false? or we will have logic based on file data?

src/apivip_check/apivip_check.go
omertuc: Yes it's enough, but I explained why I think we should check both:

>This is a hack - since we have no **official** way to know whether a worker
ignition file originated from a cluster with managed networking or not,
we instead rely on the presence of coredns and keepalived pod manifests
to indicate that. We only expect those to be present in clusters with
managed networking. To be a bit more robust, the service will consider
the presence of **any one of them** to mean that the cluster has managed
networking. **This gives us better forwards compatibility if one of them
gets renamed / replaced with other technologies in future OCP versions**.

src/apivip_check/apivip_check.go
omertuc: But there is no agent upgrade yet, and we want to include this PR in a hotfix, so I prefer to do it right. If we don't do backwards compatibility hack, we're gonna break ongoing installations and suddenly ask them for DNS validation when they don't really need it

src/apivip_check/apivip_check.go
omertuc: >or we will have logic based on file data?

I have no plans to do anything based on the contents of the file currently.

>and return true/false?

But where will you return this true/false? how will the agent communicate that to the service? Currently the agent sends the worker ign as a string. 

Also that would mean that the agent suddenly understands what these files mean, I don't want to have logic in the agent. All the agent should do is filter the worker.ign *less* then it currently does today. Ideally the agent would send the entire worker.ign, we just don't do that because it's huge

src/apivip_check/apivip_check.go
omertuc: Removed version skew hack because I figured out a way to solve it in the service (the service will only run the new decision logic on fresh clusters)

eranco74 (commented):
/lgtm

PR #392, opened 2022-07-04
MGMT-10841: increased file size for fio check

danielerez:
DiskSpeedCheck:
* Increased file size to 1GB to resemble the discovery ISO.
* Increased block size to 1MB to imitate large file writing.


romfreiman:
Do we understand how much time it will add to installations? What will happen on slow disks? What will be user experience? They gonna think it stuck?

avishayt:
/hold

Slower HDDs can do 80MB/s, so 1GB in less than 13 seconds.  On my laptop NVMe the new one finishes in 7.278s.
I just ran the original settings and surprisingly it takes 18.315s on my laptop.  It's writing less data, but with a much smaller block size.  
The current settings are based on this:
https://github.com/openshift-scale/images/blob/master/etcd-perf/run.sh and
https://www.ibm.com/cloud/blog/using-fio-to-tell-whether-your-storage-is-fast-enough-for-etcd

Before we proceed, let's do a test on vmware.  We can manually run fio and see the output.  Let's figure out if we're seeing 0 because of a bug.

tsorya:
Let's start checking on test infra . Let's see that it is not 0 seconds

danielerez:
> Do we understand how much time it will add to installations? What will happen on slow disks? What will be user experience? They gonna think it stuck?

So the check itself indeed shouldn't take longer.
Tested on a few env:
* scale lab machine
--size=22m -bs=2300:
time: `real 0m1.342s
user	0m3.553s
sys	0m0.644s`
lat_ns.percentile -> "99.000000": 32128, (0ms)
--size=1g -bs=1m:
time: `real 0m1.831s
user	0m3.274s
sys	0m1.056s`
lat_ns.percentile -> "99.000000": 528384, (0ms)
* kvm machine
--size=22m -bs=2300:
time: `real 0m29.688s
user	0m0.508s
sys	0m3.820s`
lat_ns.percentile -> "99.000000": 5603328, (5ms)
--size=1g -bs=1m:
time: `real 0m7.900s
user	0m0.446s
sys	0m2.308s`
lat_ns.percentile -> "99.000000": 11075584, (11ms)

So although 1gb test can be shorter, the sync latency 99 percentile is higher (that's the value we check in code).
I'll try to test it manually on vmware. Also with test-infra indeed...

avishayt:
99th percentile will of course be much higher when syncing 1MB instead of 2300 bytes.  We would need to change the threshold.
I have a few suspicions that I'd like to rule out before we proceed:
1. In the cases where we saw a value of 0, did we check the agent logs to see that everything was OK?
2. Are there cases where we get a result but it isn't propagated properly to the service, and we're using 0 as a default value?
3. Maybe there is an error running fio on vmware and we're defaulting to 0?
4. Maybe vmware doesn't respect the fdatasync operation and is writing the data to cache (still hard to believe it would be 0ms)?

danielerez:
> 99th percentile will of course be much higher when syncing 1MB instead of 2300 bytes. We would need to change the threshold. I have a few suspicions that I'd like to rule out before we proceed:
> 
> 1. In the cases where we saw a value of 0, did we check the agent logs to see that everything was OK?

Invoked fio on a machine with fast storage and indeed got low latency ("99.000000": 32128,). So makes sense 0ms was return, haven't seen anything else suspicious in the logs.

> 2. Are there cases where we get a result but it isn't propagated properly to the service, and we're using 0 as a default value?

When latency is very low (<1ms> we indeed don't send a value, but that's fine as it treated 0ms.

> 3. Maybe there is an error running fio on vmware and we're defaulting to 0?

Tried fio on a vmware machine, latency is <1ms ("99.000000": 144384,). So we return it as 0ms, which is ok.

> 4. Maybe vmware doesn't respect the fdatasync operation and is writing the data to cache (still hard to believe it would be 0ms)?

Invoked fio against some file in fs both on baremetal and vmware machine, same results - probably due to fast storage (nvme, etc).

Any way, seems that the disk speed validation is fine. Maybe there was an issue in storage later on due to some overload? 
It would be to validate of course as we can't predict it:/

Do we want to move forward with this change anyway, as it could be a better indication for storage latency?

avishayt:
Thanks for investigating.  It sounds like you tested fio on a filesystem on vsphere - can you please do one more check and try it directly on a device like we do when installing?

Assuming that works, maybe a better direction would be to run fio with multiple (5?) threads.  That way we can keep the same thresholds, but simulate other VMs installing at the same time and creating a noisy neighbor issue?

danielerez:
> Thanks for investigating. It sounds like you tested fio on a filesystem on vsphere - can you please do one more check and try it directly on a device like we do when installing?

Yeah, similar results when testing directly on a device (<1ms)

> 
> Assuming that works, maybe a better direction would be to run fio with multiple (5?) threads. That way we can keep the same thresholds, but simulate other VMs installing at the same time and creating a noisy neighbor issue?

yeah, could be. the problem is that this validation gets only a single file/device to check on. so invoking multiple checks concurrently might be an issue. I'll check if it affects fio drastically.

danielerez:
/cc @ori-amizur 
/cc @filanov 
/cc @oourfali 

Can you please have a look?

src/disk_speed_check/disk_speed_check.go
```diff
@@ -38,36 +45,67 @@ func (p *DiskSpeedCheck) FioPerfCheck(diskSpeedCheckRequestStr string, log logru
 		return "", err.Error(), -1
 	}
 
-	diskPerf, err := p.getDiskPerf(*diskPerfCheckRequest.Path)
+	// Invoke multiple FIO checks concurrently to simulate overload on disk
+	responseCh := make(chan fioCheckResponse, numOfFioJobs)
+	p.invokeMultipleDiskPerf(*diskPerfCheckRequest.Path, responseCh, numOfFioJobs)
```
filanov, line 50: Consider change the name, because of the wait inside the function it's not really invoking, more of executing.

src/disk_speed_check/disk_speed_check.go
```diff
@@ -38,36 +45,67 @@ func (p *DiskSpeedCheck) FioPerfCheck(diskSpeedCheckRequestStr string, log logru
...
 }
 
+func (p *DiskSpeedCheck) invokeMultipleDiskPerf(path string, responseCh chan fioCheckResponse, numOfJobs int) {
+	defer close(responseCh)
+	wg := sync.WaitGroup{}
+	for i := 1; i <= numOfJobs; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			// Invoke FIO on specified path
```
filanov, line 82:
```suggestion
			responseCh <- p.getDiskPerf(path)
```

filanov (commented):
Please add a test case of one job failed and one was successful

src/disk_speed_check/disk_speed_check.go
danielerez: Sure, renamed.

src/disk_speed_check/disk_speed_check.go
```diff
@@ -38,36 +45,67 @@ func (p *DiskSpeedCheck) FioPerfCheck(diskSpeedCheckRequestStr string, log logru
...
-	diskPerf, err := p.getDiskPerf(*diskPerfCheckRequest.Path)
+	// Invoke multiple FIO checks concurrently to simulate overload on disk
+	responseCh := make(chan fioCheckResponse, numOfFioJobs)
+	p.invokeMultipleDiskPerf(*diskPerfCheckRequest.Path, responseCh, numOfFioJobs)
+	var maxLatency int64
+	var err error
+	for res := range responseCh {
+		if res.err != nil {
+			log.WithError(res.err).Warnf("Failed to get disk's I/O performance: %s", *diskPerfCheckRequest.Path)
+			err = res.err
```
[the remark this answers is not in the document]
danielerez: RIght, changed the condition to return an error just when all requests fails.

src/disk_speed_check/disk_speed_check.go
danielerez: Done

danielerez:
> Please add a test case of one job failed and one was successful

Sure, added.

avishayt:
@filanov Good to go?  We'll need to do some intensive testing on this before it reaches prod to make sure we won't be blocking too many installations.

PR #393, opened 2022-07-04
MGMT-11037: Update models for the upgrade-agent feature

jhernand:
This patch updates the dependency on the `models` package from the
`assisted-service` project to version `d347dc4d2f2f`, so that it will be
possible to use the `upgrade-agent` step type and the corresponding
request and response types.
    
Related: https://github.com/openshift/assisted-service/blob/master/docs/enhancements/agent-upgrade.md  
Related: https://github.com/openshift/assisted-service/pull/4028  
Related: https://issues.redhat.com/browse/MGMT-11037


jhernand:
@avishayt I see that you were the last to update the dependencies on the `assisted-service` project. This is the update necessary for the upgrade agent feature. Please take a look.

jhernand:
@avishayt I think this needs an "approved" label to move forward. Can you set that?

PR #400, opened 2022-07-11
MGMT-11037: Add upgrade-agent step

jhernand:
This patch implements the `upgrade_agent` step. It downloads the new
agent image indicated by the service, and if that succeeds then it
exits the next step runner, so that the main process will start it again
with the new image.

Related: https://github.com/openshift/assisted-service/blob/master/docs/enhancements/agent-upgrade.md  
Related: https://issues.redhat.com/browse/MGMT-11037


src/upgrade_agent/upgrade_agent.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// Prepare the response:
+	response := models.UpgradeAgentResponse{
+		AgentImage: request.AgentImage,
+	}
+
+	// Download the image:
+	log.Info("Downloading image")
+	stdout, stderr, exitCode = dependencies.ExecutePrivileged(
+		"podman", "pull", request.AgentImage,
```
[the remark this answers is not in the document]
omertuc: This can be solved by:

- Increasing the retry time for this particular step (it's configured per step). But that would be hacky.

or

- Making sure no two instances of this step are allowed to run in parallel (IIRC usually we achieve this by running steps in a container and giving the container a name so new invocations of the step naturally fail with a harmless "container already exists" error)

src/commands/step_processor.go
```diff
@@ -143,6 +145,10 @@ func (s *stepSession) handleSteps(steps *models.Steps) {
 			}
 
 			s.sendStepReply(reply)
+
+			if s.requiresRestart(reply) {
+				os.Exit(0)
```
omertuc, line 150: This is too violent, simply returning from the function should be enough https://github.com/openshift/assisted-installer-agent/blob/4aecb19f0df5393be0d85937a3ee5b09bfc269b4/src/next_step_runner/main/main.go#L39-L40

jhernand: It is certainly violent, but this code runs inside a separate goroutine, one per each step, and the code that creates those goroutines doesn't wait for them to finish or coordinate in any way. So returning from the function would do nothing. The alternative to this is coordinating the main loop with this goroutine, so that this code can tell the main loop to gracefully stop. I don't see much benefit in so doing, and we aren't currently doing it for other use cases. For example, when the "after step action" is "exit" (not sure if that is ever used) the main loop will just exit without waiting for steps to complete. Also I think we don't have any code to handle signals gracefully either. My line of thought was that if those are acceptable behaviors then this should also be acceptable. If you still think that this isn't acceptable I will implement that coordination.

src/upgrade_agent/upgrade_agent.go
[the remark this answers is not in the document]
omertuc: Ack, so bullet 2 then

src/commands/step_processor.go
omertuc: I don't like `os.Exit(0)` because the agent can be [used](https://github.com/openshift-assisted/assisted-swarm/blob/94528a4076326b248b57ab7dc8afa68d4249ca83/go.mod#L16) as a library, we'd have to evaluate what would be the effect of doing that on the swarm

src/upgrade_agent/upgrade_agent.go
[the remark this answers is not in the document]
jhernand: @omertuc I pushed a new version of the patch including the use of a semaphore to prevent multiple simultaneous executions of the pull image command. Please check.

src/commands/step_processor.go
jhernand: @omertuc Regarding the `os.Exit` issue, I think that we could improve the shutdown sequence creating a `shutdown.Sequence` object that can be configured with cleanup steps and a custom exit function. That would allow it to be used in this use case, creating it like this:

```go
// Create the shutdown sequence in the main program, and pass it to other
// objects that need it:
sequence, err := shutdown.NewSequence().
        Logger(logger).
        Signal(syscall.SIGKILL).
        Signal(syscall.SIGTERM).
        Build()
if err != nil {
         ...
}

// Process the steps:
go commands.ProcessSteps(..., shutdown)

// At the end of the main program:
sequence.Start(0)

// Inside the step processor, where I am currently calling os.Exit:
if s.requiresRestart(reply) {
        s.shutdown.Start(0)
}
```

In other contexts like the swarm or in unit tests the shutdown sequence can be configured to do nothing passing explicitly a function that doesn't call `os.Exit`:

```go
sequence, err := shutdown.NewSequence().
        Logger(logger).
        Exit(func (int) {}).
        Build()
if err != nil {
        ...
}
```

This would also implement an ordered shutdown when a signal is received.

See the details in the second commit of #402.

jhernand:
Holding until after release 2.6 is done later this week.

src/commands/actions/upgrade_agent_cmd.go
```diff
@@ -0,0 +1,28 @@
...
+	log "github.com/sirupsen/logrus"
+)
+
+type upgradeAgent struct {
+	args []string
+}
+
+func (u *upgradeAgent) Validate() error {
+	var request models.UpgradeAgentRequest
+	return ValidateCommon("upgrade agent", 1, u.args, &request)
```
tsorya, line 15: maybe we should have more strict validation here? For example that we get image from the same registry ?
@avishayt ?

src/commands/step_processor.go
tsorya: Why not to check if agent should be restarted per all steps? i mean after it finished all steps it will be returned from go routing and you can restart easy without a need to exit(0) or other complicated handling

tsorya: Another question.
We want to fully restart agent? I think we should handle next_step_runner exit and not fail the agent service.
Agent service removes it's image and will re-pull it.

tsorya: Now i see that we handled exit of next_step_runner, my fault.

tsorya: Regarding exit, we are running in the loop with context , we can provide this context to go routine and set done when we want to exit in that case we will stop on next iteration.

jhernand: Done.

src/commands/actions/upgrade_agent_cmd.go
jhernand: @tsorya can you elaborate? Do you mean checking that the image should come from the same image registry than the current one? Why would we want to do that validation? Isn't the service and the communication between the agent and the service trustful enough to ensure that the image sent by the service is good to run?

@avishayt any suggestion?

src/upgrade_agent/upgrade_agent.go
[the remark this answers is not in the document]
jhernand: @omertuc kind reminder to take a look at this when you have some time.

src/upgrade_agent/upgrade_agent.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func (d *RealDependencies) ExecutePrivileged(command string, args ...string) (stdout,
+	stderr string, exitcode int) {
+	return util.ExecutePrivileged(command, args...)
+}
+
+// pullSem is used to prevent multiple simultaneous executions of the command that downloads
+// the image.
+var pullSem = semaphore.NewWeighted(1)
```
omertuc, line 32: Sempahores are intended for signaling between threads rather than mutual exclusion, so I'm not sure it's a great fit here.

A mutex would be better but sadly looks like https://pkg.go.dev/sync#Mutex.TryLock is 1.18 only so in our use case where we want the thread to simply die rather than wait we have to use a semaphore until we move to 1.18

omertuc:
PTAL
https://github.com/openshift/assisted-installer-agent/pull/406
https://github.com/openshift/assisted-installer/pull/507

PR #407, opened 2022-07-21
MGMT-11262: Include `/etc/kubernetes/kubeconfig` in day-2 ignition

omertuc:
When the ignition downloaded by the day-2 worker discovery indicates
that kubelet is connecting to the day-1 cluster using an IP address
rather than a domain name, we can safely disable DNS validations for
that worker in day-2.

This is the case in Hypershift.

The service should also be modified to detect this situation and disable
DNS validations when it's detected.

The agent should be modified to include /etc/kubernetes/kubeconfig in
the day-2 ignition, because right now it's omitted for size concerns


eranco74 (commented):
/lgtm

PR #415, opened 2022-08-03
MGMT-10738: Add download boot artifacts and reboot commands to agent

CrystalChun:
[MGMT-10738](https://issues.redhat.com/browse/MGMT-10738)
Adds two new commands for the agent:
1. Download boot artifacts
2. Reboot for reclaim

Download boot artifacts downloads the initrd and kernel to the host
machine then writes a boot loader entry config file that points
to these artifacts and to the rootfs URL.

Reboot for reclaim reboots the host machine into the discovery
image (what's written in the boot loader entry config file).

/cc @carbonin


src/download_boot_artifacts/download_boot_artifacts.go
```diff
@@ -0,0 +1,139 @@
...
+		log.Errorf("Error creating boot loader config file on host: %s", stderr)
+		return stdout, stderr, exitcode
+	}
+	return stdout, stderr, exitcode
+}
+
+func Download(downloader DownloadBootArtifactsDependencies, basePath, kernelUrl, initrdUrl string) (string, string, int) {
+	kernelPath := path.Join(basePath, kernelFile)
+	initrdPath := path.Join(basePath, initrdFile)
+	stdout, stderr, exitcode := downloader.ExecutePrivileged("curl", kernelUrl, "-o", kernelPath)
```
omertuc, line 92: Avoid short form in code, `--output` vs `-o`

Also use `--location` in case any of those links every do a redirect in the future (we've been bitten by it before)

go.mod
```diff
@@ -34,7 +34,12 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 )
 
-require github.com/itchyny/timefmt-go v0.1.3 // indirect
+require (
+	github.com/emicklei/go-restful v2.15.0+incompatible // indirect
+	github.com/google/gnostic v0.5.7-v3refs // indirect
+	github.com/itchyny/timefmt-go v0.1.3 // indirect
+	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
+)
```
carbonin, line 42: Where did these new dependencies come from?
Is this just because we updated assisted-service?

CrystalChun: Yes updating assisted-service updated these as well.

src/download_boot_artifacts/main/main.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func DownloadBootArtifactsFlags() *models.DownloadBootArtifactsRequest {
+	var (
+		initrdUrl string
+		kernelUrl string
+		rootfsUrl string
+	)
+	flag.StringVar(&initrdUrl, "initrd-url", "", "URL to download from, includes the host and port of the image service")
+	flag.StringVar(&kernelUrl, "kernel-url", "x86_64", "Architecture of host")
```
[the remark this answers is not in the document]
CrystalChun: Thanks for catching! Modifying them

src/download_boot_artifacts/download_boot_artifacts.go
CrystalChun: Gotcha, thank you!

src/commands/actions/reboot_for_reclaim.go
```diff
@@ -0,0 +1,36 @@
...
+	return stdout, stderr, exitCode
+}
+
+func (a *rebootForReclaim) Command() string {
+	return "reboot"
+}
+
+func (a *rebootForReclaim) Args() []string {
+	return []string{}
+}
```
carbonin, line 36: Should these match what we're actually running?

I don't really know what they're supposed to do.

src/commands/actions/reboot_for_reclaim_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+var _ = Describe("reboot_for_reclaim", func() {
+	It("happy flow", func() {
+		action, err := New(&config.AgentConfig{}, models.StepTypeRebootForReclaim, []string{})
+		Expect(err).NotTo(HaveOccurred())
+
+		args := action.Args()
+		command := action.Command()
+		Expect(command).To(Equal("reboot"))
+		Expect(args).To(BeEmpty())
```
carbonin, line 18: This feels like it isn't testing much.

As I put in my other comment, I'm not sure what these do as they don't match what we're actually running.
Also they're just hardcoded values so I don't see much point in testing this.

src/download_boot_artifacts/download_boot_artifacts.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+	return stdout, stderr, exitcode
+}
+
+func CreateBootLoaderConfig(downloader DownloadBootArtifactsDependencies, rootfsUrl, artifactsPath string) (string, string, int) {
+	kernelPath := path.Join(artifactsPath, kernelFile)
+	initrdPath := path.Join(artifactsPath, initrdFile)
+	bootLoaderConfig := fmt.Sprintf("title Assisted Installer Discovery\nversion 999\n"+
+		"options random.trust_cpu=on ignition.firstboot ignition.platform.id=metal "+
+		"'coreos.live.rootfs_url=%s'\nlinux %s\ninitrd %s",
```
carbonin, line 109: This will probably be easier to read if you put it in a const or something.
Similar to what we did for [ipxe](https://github.com/openshift/assisted-service/blob/d0cfbf7a9d9a65a68efb19eadff508ed8da20818/internal/bminventory/inventory_v2_handlers.go#L635-L639)

carbonin (commented):
First question I have is why everything has to go through nsenter (i.e. `ExecutePrivileged`).

In this case we have a volume mounted into our pod, I don't see why we couldn't use regular go-native utilities to do things like write files.

Did you run into issues with that approach?
Ideally I'd like to use as much native go as we can. Even things like [`mount` seem available in the `syscall` package](https://pkg.go.dev/syscall@go1.19#Mount). I see you're using that for chroot already so I'm not sure why we wouldn't be able to use it in other cases as well.

omertuc:
> I don't see why we couldn't use regular go-native utilities to do things like write file

Should probably also use `go`'s native `http` and `io.Copy` to download the files instead of using `curl`

carbonin:
Yeah, everywhere we can I'd rather native libs over execing.

src/download_boot_artifacts/download_boot_artifacts.go
CrystalChun: Good idea, modified! How does it look?

src/commands/actions/reboot_for_reclaim.go
CrystalChun: They're part of the action interface that's inherited. Left a comment that they're unused

src/download_boot_artifacts/download_boot_artifacts.go
```diff
@@ -0,0 +1,120 @@
...
+	return nil
+}
+
+func Download(filePath, url string) error {
+	res, err := http.Get(url)
+	if err != nil {
+		log.WithError(err).Errorf("Error getting %s", url)
+		return err
+	}
+
```
omertuc, line 71: Ideally immediately `defer res.Body.Close()` at this point, rather than manually closing it at a later point

src/download_boot_artifacts/download_boot_artifacts.go
```diff
@@ -0,0 +1,120 @@
...
+options random.trust_cpu=on ignition.firstboot ignition.platform.id=metal 'coreos.live.rootfs_url=%s'
+linux %s
+initrd %s`
+
+func Run(infraEnvId, downloaderRequestStr string) error {
+	var req models.DownloadBootArtifactsRequest
+	if err := json.Unmarshal([]byte(downloaderRequestStr), &req); err != nil {
+		log.WithError(err).Errorf("DownloadBootArtifacts: json.Unmarshal")
+		return err
+	}
```
omertuc, line 36: I think it's best that if instead of logging at every error site, we wrap the error using `fmt.Errorf` and return that, then log higher up the stack.

e.g.:
```suggestion
	if err := json.Unmarshal([]byte(downloaderRequestStr), &req); err != nil {
		return fmt.Errorf("failed to unmarshal request: %w", err)
	}
```

This way both the error that gets logged and the error that gets returned contain the full context information

src/download_boot_artifacts/download_boot_artifacts.go
```diff
@@ -0,0 +1,120 @@
...
+	log.Infof("Successfully downloaded boot artifacts and created bootloader config.")
+	return nil
+}
+
+func Download(filePath, url string) error {
+	res, err := http.Get(url)
+	if err != nil {
+		log.WithError(err).Errorf("Error getting %s", url)
+		return err
+	}
```
omertuc, line 70: [Same here](https://github.com/openshift/assisted-installer-agent/pull/415/files#r941227306)

src/download_boot_artifacts/download_boot_artifacts.go
CrystalChun: Good idea! Modified to wrap error messages

src/download_boot_artifacts/download_boot_artifacts.go
CrystalChun: Gotcha, thanks! Modified to do this

src/download_boot_artifacts/download_boot_artifacts.go
```diff
@@ -0,0 +1,106 @@
...
+initrd %s`
+
+func Run(infraEnvId, downloaderRequestStr string) error {
+	var req models.DownloadBootArtifactsRequest
+	if err := json.Unmarshal([]byte(downloaderRequestStr), &req); err != nil {
+		return fmt.Errorf("failed unmarshalling download boot artifacts request, %w", err)
+	}
+
+	if err := syscall.Mount("/", "/host/boot", "", syscall.MS_REMOUNT, ""); err != nil {
+		return fmt.Errorf("failed remounting /host/boot folder as rw, %w", err)
```
omertuc, line 38: nit: I believe the idiomatic separator before `%w` is `:` and not `,`

src/download_boot_artifacts/download_boot_artifacts.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func Download(filePath, url string) error {
+	res, err := http.Get(url)
+	if err != nil {
+		return fmt.Errorf("failed getting %s, %w", url, err)
+	}
+	defer res.Body.Close()
+	body, err := io.ReadAll(res.Body)
+	if res.StatusCode != 200 {
```
omertuc, line 67: check status >= 200 and < 300 - all 2xx statuses are somewhat positive

src/download_boot_artifacts/download_boot_artifacts.go
CrystalChun: Thanks for letting me know! Modified!

src/download_boot_artifacts/download_boot_artifacts.go
CrystalChun: Gotcha, modified it, thanks!

CrystalChun:
@carbonin @omertuc Ready for re-review! I separated the vendor updates into a separate commit based on Nick's suggestion

Makefile
```diff
@@ -33,7 +33,7 @@ lint: ci-lint
 .PHONY: build clean build-image push subsystem
 build: build-agent build-connectivity_check build-inventory build-free_addresses build-logs_sender \
 	   build-dhcp_lease_allocate build-apivip_check build-next_step_runner build-ntp_synchronizer \
-	   build-container_image_availability build-domain_resolution build-disk_speed_check
+	   build-container_image_availability build-domain_resolution build-disk_speed_check build-download_boot_artifacts
```
carbonin, line 36: I don't think this is needed.

We run our code directly as a part of `next_step_runner` so I don't think we need an extra executable.

src/download_boot_artifacts/main/main.go
```diff
@@ -0,0 +1,48 @@
+package main
```
carbonin, line 1: I don't think we need a `main` or an executable for either of these commands as we run them directly from the `next_step_runner`.

src/download_boot_artifacts/download_boot_artifacts_test.go
```diff
@@ -0,0 +1,112 @@
...
+		bootFile   string
+		bootFolder string
+		rootfsUrl  string
+	)
+
+	BeforeEach(func() {
+		tempDir = os.TempDir()
+		kernelFile = "vmlinuz"
+		initrdFile = "initrd"
+		rootfsUrl = "http://test.com/rootfs?arch=x86_64&version=4.10"
```
carbonin, line 27: These can just be set in the `var` block assuming they are not changing per test.

carbonin: Alternatively you could use the values from the source file if they will always be the same.

src/download_boot_artifacts/download_boot_artifacts_test.go
```diff
@@ -0,0 +1,112 @@
...
+				folder, err = os.Stat(bootFolder)
+				Expect(err).To(BeNil())
+				Expect(folder.IsDir()).To(BeTrue())
+			})
+		})
+	})
+
+	Context("Create bootloader config file", func() {
+		AfterEach(func() {
+			err := os.RemoveAll(bootFolder)
```
carbonin, line 65: I think this is handled by the `RemoveAll` call above. You shouldn't need to do it twice.

src/download_boot_artifacts/download_boot_artifacts_test.go
```diff
@@ -0,0 +1,112 @@
...
+import (
+	"os"
+	"path"
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Download boot artifacts test", func() {
```
carbonin, line 12: Generally `Describe` should be used with the function you're testing, so I'd expect a structure like:
```go
var _ = Describe("createFolders", func() {
        It("creates all the folders", func() {
                ...
        })
})

var _ = Describe("createBootLoaderConfig", func() {
})

...
```

src/download_boot_artifacts/download_boot_artifacts_test.go
```diff
@@ -0,0 +1,112 @@
...
+		bootFolder = path.Join(tempDir, "loader")
+		bootFile = path.Join(bootFolder, "00-assisted-discovery.conf")
+	})
+
+	AfterEach(func() {
+		os.RemoveAll(artifactsFolder)
+		os.RemoveAll(bootFolder)
+	})
+
+	Context("Create Folders", func() {
```
carbonin, line 38: `Context` is only needed when you're defining some shared setup or teardown code for all the `It`s in the context.

In this case you only have one `It` block and no shared setup/teardown so it's not necessary.

src/commands/actions/download_boot_artifacts_cmd.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+type downloadBootArtifacts struct {
+	args        []string
+	agentConfig *config.AgentConfig
+}
+
+func (a *downloadBootArtifacts) Validate() error {
+	modelToValidate := models.DownloadBootArtifactsRequest{}
+	err := ValidateCommon("download boot artifacts", 1, a.args, &modelToValidate)
```
carbonin, line 16: Super minor, but you can just send `&models.DownloadBootArtifactsRequest{}` directly here if you don't care about reading the contents after this call.

src/commands/actions/reboot_for_reclaim.go
```diff
@@ -0,0 +1,37 @@
...
+func (a *rebootForReclaim) Validate() error {
+	err := ValidateCommon("reboot for reclaim", 0, a.args, nil)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (a *rebootForReclaim) Run() (stdout, stderr string, exitCode int) {
+	err := syscall.Chroot("/host")
```
carbonin, line 22: I know it's late to mention this, but I'm wondering if this should come from the service in an arg.

As it is, this mountpoint is defined in two places and we probably only want it in the service (since that's where we're creating the container).

carbonin (commented):
Looking good, mostly minor test changes.

Two larger things I want to bring up here, just so they don't get lost:
1. I don't think we need an executable
2. I'm thinking we should provide the mountpoint to chroot to as an arg to the reboot command.

src/download_boot_artifacts/download_boot_artifacts.go
```diff
@@ -0,0 +1,110 @@
...
+	"net/http"
+	"os"
+	"path"
+	"syscall"
+
+	"github.com/openshift/assisted-service/models"
+	log "github.com/sirupsen/logrus"
+)
+
+var (
```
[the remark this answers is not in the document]
CrystalChun: Yes, but I think the folders (e.g. `hostArtifactsFolder`) should be variables in case we want to use a different directory (arg passed from service) in the future. Wdyt?

src/download_boot_artifacts/download_boot_artifacts_test.go
```diff
@@ -0,0 +1,112 @@
...
+		initrdFile = "initrd"
+		rootfsUrl = "http://test.com/rootfs?arch=x86_64&version=4.10"
+		artifactsFolder = path.Join(tempDir, "artifacts")
+		bootFolder = path.Join(tempDir, "loader")
+		bootFile = path.Join(bootFolder, "00-assisted-discovery.conf")
+	})
+
+	AfterEach(func() {
+		os.RemoveAll(artifactsFolder)
+		os.RemoveAll(bootFolder)
```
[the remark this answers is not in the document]
CrystalChun: Good idea, modified it, thank you!

src/download_boot_artifacts/download_boot_artifacts.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	if err := CreateBootLoaderConfig(*req.RootfsURL, artifactsFolder, bootLoaderFolder); err != nil {
+		return fmt.Errorf("failed creating bootloader config file on host: %w", err)
+	}
+	log.Infof("Successfully downloaded boot artifacts and created bootloader config.")
+	return nil
+}
+
+func Download(filePath, url string) error {
```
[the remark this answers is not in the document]
CrystalChun: Gotcha, I made them private and moved them

src/commands/actions/download_boot_artifacts_cmd_test.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+var _ = Describe("download boot artifacts", func() {
+	var param string
+
+	BeforeEach(func() {
+		param = "{\"initrd_url\":\"http://test.com/api/v2/pxe-initrd?api_key=123&arch=x86_64&version=4.10\"," +
+			"\"rootfs_url\":\"http://test.com/rootfs?arch=x86_64&version=4.10\"," +
+			"\"kernel_url\":\"http://test.com/kernel?arch=x86_64&version=4.10\"}"
+	})
```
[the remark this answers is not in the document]
CrystalChun: Good catch, modified it, thank you!

Makefile
CrystalChun: Gotcha, removed it!

src/download_boot_artifacts/main/main.go
CrystalChun: Gotcha, removed this!

src/download_boot_artifacts/download_boot_artifacts_test.go
CrystalChun: Thanks for the catch, modified it!

src/download_boot_artifacts/download_boot_artifacts_test.go
CrystalChun: Gotcha, removed this!

src/download_boot_artifacts/download_boot_artifacts_test.go
CrystalChun: Gotcha, good to know! Thanks!

src/commands/actions/reboot_for_reclaim.go
CrystalChun: I can add this as an argument we send from the service to the agent - wdyt? Is it worth the time/effort to do it in this iteration?

src/download_boot_artifacts/download_boot_artifacts.go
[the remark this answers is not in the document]
carbonin: I don't think that means we can't make them constants now.

src/commands/actions/download_boot_artifacts_cmd.go
```diff
@@ -0,0 +1,145 @@
...
+
+// Unused, but required as part of ActionInterface
+func (a *downloadBootArtifacts) Args() []string {
+	return a.args
+}
+
+var (
+	hostArtifactsFolder string = "/host/boot/discovery"
+	artifactsFolder     string = "/boot/discovery"
+	bootLoaderFolder    string = "/host/boot/loader/entries"
```
carbonin, line 52: So I suppose we should do the same here for `/host` if we're moving it into an arg for reboot.

CrystalChun:
Switching to draft until https://github.com/openshift/assisted-service/pull/4263 is merged

src/commands/actions/download_boot_artifacts_cmd.go
```diff
@@ -0,0 +1,139 @@
...
+
+	if err := os.WriteFile(bootLoaderConfigFile, []byte(bootLoaderConfig), 0644); err != nil {
+		return fmt.Errorf("failed writing bootloader config content to %s: %w", bootLoaderConfigFile, err)
+	}
+	return nil
+}
+
+func createFolderIfNotExist(folder string) error {
+	if _, err := os.Stat(folder); os.IsNotExist(err) {
+		return os.MkdirAll(folder, 0765)
```
carbonin, line 124: This seems like a strange choice of permissions, wouldn't `0755` be more standard?
Do we specifically need group write for some reason?

carbonin (commented):
One comment about permissions. Otherwise looks good.

@omertuc did you want to take another look?

omertuc:
No further comments

carbonin (approved):

PR #419, opened 2022-08-10
MGMT-11575: Ignore empty MAC address for host UUID generation

javipolo:
When falling back to MAC generated host UUID , we have loopback device in the list of interfaces with empty UUID

If this interface gets picked up to be the one to generate the UUID (and it does, since the sorting provokes that it ends up always being the first element), it will create always the same UUID for all hosts:

```
Aug 10 13:27:38 109-74-204-122 agent[5130]: time="10-08-2022 13:27:38" level=info msg="Using mac from interface lo to provide node-uuid" file="machine_uuid_scanner.go:118"
```

cc: @avishayt 
cc: @filanov

### DIFF
--- a/src/scanners/machine_uuid_scanner.go
+++ b/src/scanners/machine_uuid_scanner.go
@@ -14,6 +14,7 @@ import (
 	"github.com/jaypipes/ghw/pkg/util"
 	"github.com/openshift/assisted-installer-agent/src/inventory"
 	agent_utils "github.com/openshift/assisted-installer-agent/src/util"
+	"github.com/openshift/assisted-service/models"
 	log "github.com/sirupsen/logrus"
 )
 
@@ -24,6 +25,10 @@ const (
 	SerialNotSpecified               = "not specified"                        // Linode
 	ZeroesUUID                       = "00000000-0000-0000-0000-000000000000"
 	KaloomUUID                       = "03000200-0400-0500-0006-000700080009" // All hosts of this type have the same UUID
+)
+
+var (
+	FailureUUID = strfmt.UUID("deaddead-dead-dead-dead-deaddeaddead")
 )
 
 var unknownSerialCases = []string{"", util.UNKNOWN, "none",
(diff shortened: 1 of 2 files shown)


src/scanners/machine_uuid_scanner.go
```diff
@@ -115,9 +115,16 @@ func ReadId(d SerialDiscovery, dependencies agent_utils.IDependencies) *strfmt.U
 		sort.Slice(interfaces, func(i, j int) bool {
 			return interfaces[i].MacAddress < interfaces[j].MacAddress
 		})
+
+		// ignore empty MacAddress
+		var i int
+		for i = 0; i < len(interfaces); i++ {
+			if (interfaces[i].MacAddress != "") { break }
```
omertuc, line 122: no `(` `)` around `if` conditions needed

src/scanners/machine_uuid_scanner.go
```diff
@@ -115,9 +115,16 @@ func ReadId(d SerialDiscovery, dependencies agent_utils.IDependencies) *strfmt.U
...
+		}
+
 		log.Infof("Using %s mac from interface %s to provide node-uuid",
-			interfaces[0].MacAddress, interfaces[0].Name)
-		ret = md5GenerateUUID(interfaces[0].MacAddress)
+			interfaces[i].MacAddress, interfaces[i].Name)
```
omertuc, line 126: out of bounds access when non of the interfaces have a mac address

src/scanners/machine_uuid_scanner.go
javipolo: Fixed!

src/scanners/machine_uuid_scanner.go
javipolo: Fixed. At least now it wont break code, even if there are no network interfaces with a mac address, we'll still be generating the same UUID always

Do you think it's worth to add a couple of extra lines to check for that corner case and output some meaningful log?

omertuc: >Fixed. At least now it wont break code, even if there are no network interfaces with a mac address, we'll still be generating the same UUID always

Great, thanks

>Do you think it's worth to add a couple of extra lines to check for that corner case and output some meaningful log?

Yes

src/scanners/machine_uuid_scanner.go
```diff
@@ -115,9 +115,16 @@ func ReadId(d SerialDiscovery, dependencies agent_utils.IDependencies) *strfmt.U
 		sort.Slice(interfaces, func(i, j int) bool {
 			return interfaces[i].MacAddress < interfaces[j].MacAddress
 		})
+
+		// ignore empty MacAddress
+		var i int
+		for i = 0; i < len(interfaces)-1; i++ {
```
omertuc, line 121: prefer for-each loops over index loops, e.g. `for _, interface := range interfaces {`

src/scanners/machine_uuid_scanner.go
javipolo: Ok, added a warning
I'm returning a dummy UUID anyways, I dont know whathappens if I dont return a UUID ...

src/scanners/machine_uuid_scanner.go
javipolo: agree, changed!

src/scanners/machine_uuid_scanner.go
```diff
@@ -115,9 +115,20 @@ func ReadId(d SerialDiscovery, dependencies agent_utils.IDependencies) *strfmt.U
 		sort.Slice(interfaces, func(i, j int) bool {
 			return interfaces[i].MacAddress < interfaces[j].MacAddress
 		})
-		log.Infof("Using %s mac from interface %s to provide node-uuid",
-			interfaces[0].MacAddress, interfaces[0].Name)
-		ret = md5GenerateUUID(interfaces[0].MacAddress)
+
+		// ignore empty MacAddress
+		for _, iface := range(interfaces) {
```
omertuc, line 120: no parenthesis around range

javipolo: ACK ... fixed

PR #422, opened 2022-08-18
MGMT-11447: Agent should report error when trying to pull ignition

omertuc:
Implement service API introduced in
https://github.com/openshift/assisted-service/commit/d21ae1ef34e6cda0e613a89590bc284c59d7bf4a

### DIFF
--- a/src/apivip_check/apivip_check.go
+++ b/src/apivip_check/apivip_check.go
@@ -6,7 +6,7 @@ import (
 	b64 "encoding/base64"
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"net/http"
 
 	"github.com/coreos/ignition/v2/config/v3_2"
@@ -17,6 +17,10 @@ import (
 )
 
 const ignitionVersion string = "3.2.0"
+
+const responseErrorSizeLimit int = 512
+
+const ignitionDownloadErrorStderr string = "ignition download error"
 
 func getIgnitionFile(ignition *ignition_types.Config, path string) *ignition_types.File {
 	for _, file := range ignition.Storage.Files {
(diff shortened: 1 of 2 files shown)


src/apivip_check/apivip_check.go
```diff
@@ -183,26 +207,43 @@ func httpDownload(connectivityReq models.APIVipConnectivityRequest) (string, err
...
-	if res.StatusCode != http.StatusOK {
-		return "", errors.Errorf("HTTP download failure. Status Code: %v", res.StatusCode)
+		return "", errors.Wrap(err, fmt.Sprintf("error while reading response body, read %d bytes", len(bytes)))
 	}
 
-	defer res.Body.Close()
-	bytes, err := ioutil.ReadAll(res.Body)
-	if err != nil {
-		return "", errors.Wrap(err, "File read failure")
+	if res.StatusCode < 200 || res.StatusCode >= 300 {
```
ori-amizur, line 223: 300 codes are redirection messages.  Not sure they should be treated as errors.
100 codes are information messages. Not sure they should be treated as errors.

src/apivip_check/apivip_check.go
```diff
@@ -100,65 +102,84 @@ func CheckAPIConnectivity(checkAPIRequestStr string, log logrus.FieldLogger) (st
 	var checkAPIRequest models.APIVipConnectivityRequest
 
 	if err := json.Unmarshal([]byte(checkAPIRequestStr), &checkAPIRequest); err != nil {
-		wrapped := errors.Wrap(err, "Error unmarshaling APIVipConnectivityRequest")
-		log.WithError(err).Error(wrapped.Error())
-		return createResponse(false, ""), wrapped.Error(), -1
+		return createResponse("<unknown URL due to internal error>", false,
```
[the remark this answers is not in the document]
omertuc: The point of this change was to move the error into the response struct rather than stderr so it can be displayed as part of the validation, there's no reason to duplicate it in stderr as well, stderr has no use for this step anymore

src/apivip_check/apivip_check.go
omertuc: go http is supposed to follow redirects automatically, so if they reached our application code, it means something went wrong

Not too familiar with 100 codes so I can't say for sure but I don't think they're supposed to reach us as well.

omertuc:
> maybe easier just to check for some ignition substring in the answer without need to open it?

Parsing the ignition is something we already did before this PR, I did not change tihs

src/apivip_check/apivip_check.go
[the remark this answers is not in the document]
ori-amizur: I still think in case of error some error string should be returned as part of stderr

src/apivip_check/apivip_check.go
ori-amizur: HTTP2 uses 101 code in order to switch.  I am not sure how it should be handled.

src/apivip_check/apivip_check.go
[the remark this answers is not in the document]
omertuc: Alright, changed to return a constant error message

src/apivip_check/apivip_check.go
omertuc: Alright. Since I don't know what we should do in case of 1xx codes, I think for now it's best if we treat it as an unexpected error

PR #425, opened 2022-08-26
MGMT-10738: Include secure transport for downloading boot artifacts

CrystalChun:
[MGMT-10738](https://issues.redhat.com/browse/MGMT-10738)
https://github.com/openshift/assisted-installer-agent/pull/415
Adds a secure transport option when downloading boot artifacts
that was missing in the first PR.

### DIFF
--- a/src/commands/actions/download_boot_artifacts_cmd.go
+++ b/src/commands/actions/download_boot_artifacts_cmd.go
@@ -1,6 +1,8 @@
 package actions
 
 import (
+	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


src/commands/actions/download_boot_artifacts_cmd.go
```diff
@@ -86,8 +94,31 @@ func run(infraEnvId, downloaderRequestStr string) error {
 	return nil
 }
 
-func download(filePath, url string) error {
-	res, err := http.Get(url)
+func createAssistedServiceClient(caCertPath string) (*http.Client, error) {
+	client := &http.Client{}
+	if caCertPath != "" {
+		caCert, err := ioutil.ReadFile(caCertPath)
```
carbonin, line 100: ioutil is deprecated.

I think the choice for new code is [`os.ReadFile`](https://pkg.go.dev/os@go1.19#ReadFile)

src/commands/actions/download_boot_artifacts_cmd.go
```diff
@@ -86,8 +94,31 @@ func run(infraEnvId, downloaderRequestStr string) error {
 	return nil
 }
 
-func download(filePath, url string) error {
-	res, err := http.Get(url)
+func createAssistedServiceClient(caCertPath string) (*http.Client, error) {
```
carbonin, line 97: It's not technically an assisted service client. So maybe just call it `createHTTPClient` or something.

carbonin (approved):

PR #427, opened 2022-09-02
MGMT-11185: disk_speed_check: escape colons when calling fio

vrutkovs:
fio's filename can be a colon-separated list of devices to test. This however breaks some paths with colons, so these need to be escaped.
See https://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-arg-filename

### DIFF
--- a/src/disk_speed_check/disk_speed_check.go
+++ b/src/disk_speed_check/disk_speed_check.go
@@ -2,6 +2,7 @@ package disk_speed_check
 
 import (
 	"encoding/json"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


omertuc:
Also let's mention in the commit message that it only seem to have affected OKD for an unknown reason

vrutkovs:
I think RHCOS is affected too - its just OKD quickly runs out of space when `fio` writes file to a location until the first colon - its just we didn't notice it (and were always testing root disk performance instead of target disk)

vrutkovs:
/hold

For jira ticket and linting

omertuc:
> I think RHCOS is affected too - its just OKD quickly runs out of space when `fio` writes file to a location until the first colon - its just we didn't notice it (and were always testing root disk performance instead of target disk)

Ah I see. well copy this paragraph into the commit message then :) It helps to know what is being fixed (and how things worked previosuly) when trying to understand whether things should be backported / reverted / prioritized

omertuc:
/approve

https://github.com/openshift/assisted-installer-agent/pull/428

PR #431, opened 2022-09-06
Bug-2105358: Don't use addressless interfaces in network checks

mkowalski:


mkowalski:
/retitle Bug 2105358: Don't use addressless interfaces in network checks

src/connectivity_check/connectivity_check.go
```diff
@@ -30,6 +30,15 @@ func getOutgoingNics(dryRunConfig *config.DryRunConfig) []string {
 		if !(intf.IsPhysical() || intf.IsBonding() || intf.IsVlan()) {
 			continue
 		}
+
+		// In order to remediate issue with polluting ARP table by using enslaved interfaces
```
carbonin, line 34: Is this the best place to do this?

Could there be some callers that might want interfaces without addresses?

mkowalski: As far as my knowledge goes, nowhere in the service we should ever try to use addressless interface. We only ever communicate using IP stack (we don't directly use any L2 protocols) therefore I can't think of a scenario of some component trying to use no-IP interface

mkowalski: Just a nit to the comment above - it does not have to go as far as "nowhere in the service" because this is only a "connectivity check" package

carbonin: > this is only a "connectivity check" package

Ah, missed that. I thought this was some kind of utility that might end up being used for something other than connecting.
Looks fine in that case then.

carbonin:
Looks good, but can you add a test?

src/connectivity_check/connectivity_check.go
```diff
@@ -18,9 +18,11 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-func getOutgoingNics(dryRunConfig *config.DryRunConfig) []string {
+func getOutgoingNics(dryRunConfig *config.DryRunConfig, d util.IDependencies) []string {
 	ret := make([]string, 0)
-	d := util.NewDependencies(dryRunConfig, "")
+	if d == nil {
+		d = util.NewDependencies(dryRunConfig, "")
```
mkowalski, line 24: I need to do this because otherwise the function is non-testable. Its original design is ugly but it is what it is

eranco74:
/hold
Fix lint...

src/connectivity_check/connectivity_check.go
eranco74: Despite the comment it's unclear why and when `d` will be nil, why not pass `util.NewDependencies(dryRunConfig, "")` as `d` from the test?

mkowalski: copy-pasta from slack:

I need to do the trick with d as function param because this function was originally written in the most untestable way ever. I must not pass util.NewDependencies(dryRunConfig, "") in the test, I need to use MockIDepencencies instead. So in the test I create and pass util.MockIDependencies and in real run I don't pass anything, so the code runs with util.NewDependencies(dryRunConfig, "")

mkowalski:
> /hold Fix lint...

Not clear to me...
```
 tivity_check/connectivity_check_test.go:528:10: NewFilledInterfaceMock not declared by package util]" 
```
is the error but this code compiles and runs this function correctly. It's defined in `src/util/mocks.go` which has `package util` at the very top

eranco74:
UT has the same issue:
```
# github.com/openshift/assisted-installer-agent/src/connectivity_check [github.com/openshift/assisted-installer-agent/src/connectivity_check.test]
src/connectivity_check/connectivity_check_test.go:524:5: undefined: util.NewFilledInterfaceMock
src/connectivity_check/connectivity_check_test.go:525:5: undefined: util.NewFilledInterfaceMock
src/connectivity_check/connectivity_check_test.go:526:5: undefined: util.NewFilledInterfaceMock
src/connectivity_check/connectivity_check_test.go:527:5: undefined: util.NewFilledInterfaceMock
src/connectivity_check/connectivity_check_test.go:528:5: undefined: util.NewFilledInterfaceMock
```
perhaps it's a version issue?

mkowalski:
LOL how come

```
$ skipper make unit-test
[...]
ok      github.com/openshift/assisted-installer-agent/src/apivip_check  37.381s coverage: 75.7% of statements
?       github.com/openshift/assisted-installer-agent/src/apivip_check/main     [no test files]
?       github.com/openshift/assisted-installer-agent/src/commands      [no test files]
?       github.com/openshift/assisted-installer-agent/src/config        [no test files]
?       github.com/openshift/assisted-installer-agent/src/connectivity_check/main       [no test files]
?       github.com/openshift/assisted-installer-agent/src/container_image_availability/main     [no test files]
?       github.com/openshift/assisted-installer-agent/src/dhcp_lease_allocate/main      [no test files]
?       github.com/openshift/assisted-installer-agent/src/disk_speed_check/main [no test files]
?       github.com/openshift/assisted-installer-agent/src/domain_resolution/main        [no test files]
?       github.com/openshift/assisted-installer-agent/src/free_addresses/main   [no test files]
?       github.com/openshift/assisted-installer-agent/src/inventory/main        [no test files]
?       github.com/openshift/assisted-installer-agent/src/logs_sender/main      [no test files]
?       github.com/openshift/assisted-installer-agent/src/next_step_runner/main [no test files]
?       github.com/openshift/assisted-installer-agent/src/ntp_synchronizer/main [no test files]
?       github.com/openshift/assisted-installer-agent/src/session       [no test files]
?       github.com/openshift/assisted-installer-agent/src/util/nmap     [no test files]

DONE 14 tests in 39.409s
[...]
```

mkowalski:
/unhold

Fixed build

mkowalski:
/retitle MGMT-11953: Don't use addressless interfaces in network checks

eranco74 (commented):
/lgtm

PR #433, opened 2022-09-11
MGMT-11970: Ensure that we handle an IP conflict correctly.

paul-maidment:
When there is an L2 IP address conflict, resulting in an arping result similar to the one shown below, we need to ensure that this is correctly reported to asssisted-service so that we can advise the user of the issue correctly.

```arping -I eth0 192.168.150.74
ARPING 192.168.150.74 from 192.168.150.75 eth0
Unicast reply from 192.168.150.74 [00:50:56:95:BA:89]  0.685ms
Unicast reply from 192.168.150.74 [00:50:56:85:16:66]  0.837ms
Unicast reply from 192.168.150.74 [00:50:56:85:16:66]  0.754ms
```

This change will attempt to remedy this by making sure that assisted-service is aware of the IP collission so that the user may be advised. Presently we do not do this correctly due to the following issues:

1: The existing code incorrectly reuses pointers to L2Connectity and L3Connectivity structs, meaning that the content of these structs is incorrectly interpreted, leading to a misreporting of actual results.
2: Insufficient time is given to gather unicast replies, this commit changes that by altering `arping` parameters to make this work correctly. The parameter "-c" has been altered to have a value of "10" to allow this.

The net effect of this change is that deduplicated representation of all unicast replies will be sent back to assisted-service for the host. This will allow checks to be made in assisted-service to determine whether or not there are IP collisions for the host.

In conjunction with the changes in https://github.com/openshift/assisted-service/pull/4459
The intention is to add a validation that there are no IP collisions.

### DIFF
--- a/src/connectivity_check/connectivity_check.go
+++ b/src/connectivity_check/connectivity_check.go
@@ -81,8 +81,8 @@ type connectivity struct {
 	dryRunConfig *config.DryRunConfig
 }
 
-func (c *connectivity) l3CheckAddressOnNic(address string, outgoingNic string, innerChan chan *models.L3Connectivity, conCheck connectivityCmd) {
-	ret := &models.L3Connectivity{
+func (c *connectivity) l3CheckAddressOnNic(address string, outgoingNic string, innerChan chan models.L3Connectivity, conCheck connectivityCmd) {
+	ret := models.L3Connectivity{
 		OutgoingNic:     outgoingNic,
 		RemoteIPAddress: address,
 	}
@@ -101,7 +101,7 @@ func (c *connectivity) l3CheckAddressOnNic(address string, outgoingNic string, i
 		innerChan <- ret
 		return
 	}
-	err = parsePingCmd(ret, string(b))
+	err = parsePingCmd(&ret, string(b))
 	if err != nil {
 		log.Error(err)
 		innerChan <- ret
@@ -156,7 +156,7 @@ func (c *connectivity) l3CheckConnectivity(addresses []string, dataCh chan any, 
 
 func (c *connectivity) l3CheckAddress(address string, outgoingNics []string, dataCh chan any, wg *sync.WaitGroup, conCheck connectivityCmd) {
 	defer wg.Done()
-	innerChan := make(chan *models.L3Connectivity)
+	innerChan := make(chan models.L3Connectivity)
 	for _, nic := range outgoingNics {
 		go c.l3CheckAddressOnNic(address, nic, innerChan, conCheck)
 	}
@@ -169,7 +169,7 @@ func (c *connectivity) l3CheckAddress(address string, outgoingNics []string, dat
 		}
 	}
 	if !successful {
-		ret := &models.L3Connectivity{
+		ret := models.L3Connectivity{
 			RemoteIPAddress: address,
 		}
 		dataCh <- ret
@@ -197,7 +197,7 @@ func (c *connectivity) l2CheckAddressOnNic(dstAddr string, dstMAC string, allDst
 
 func analyzeNmap(dstAddr string, dstMAC string, allDstMACs []string, srcNIC string, dataCh chan any, conCheck connectivityCmd, dryRunEnabled bool) {
 
-	ret := &models.L2Connectivity{
+	ret := models.L2Connectivity{
 		OutgoingNic:     srcNIC,
 		RemoteIPAddress: dstAddr,
 	}
@@ -256,7 +256,7 @@ func (c *connectivity) l2CheckAddress(dstAddr string, dstMAC string, allDstMACs 
 	for numDone := 0; numDone != len(conCheck.getOutgoingNICs()); {
 		iret := <-innerChan
 		switch ret := iret.(type) {
-		case *models.L2Connectivity:
+		case models.L2Connectivity:
 			received = true
 			dataCh <- ret
 		case done:
@@ -264,7 +264,7 @@ func (c *connectivity) l2CheckAddress(dstAddr string, dstMAC string, allDstMACs 
 		}
 	}
 	if !received {
-		dataCh <- &models.L2Connectivity{
+		dataCh <- models.L2Connectivity{
 			RemoteIPAddress: dstAddr,
 		}
 	}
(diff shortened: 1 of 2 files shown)


ori-amizur:
Please make sure that the returned L2 results are unique

omertuc:
Please edit commit message / PR description to explain what we did before, and what we do now differently, and why it helps detect IP conflicts

Also PR title is wrong - this is not a mac conflict detection, this is an IP conflict detection

paul-maidment:
> Please edit commit message / PR description to explain what we did before, and what we do now differently, and how making this change helps us detect IP conflicts.
> 
> Please mention in brief any upcoming service changes that might need to follow this PR
> 
> Also PR title is wrong - this is not a mac conflict detection, this is an IP conflict detection

I have updated the PR message and title as requested.

omertuc:
- The commit message doesn't match the PR description
- The commit message doesn't mention the most important point, which is changing the `"-c", "10"` parameter to get more than one arping result
- The commit message keeps mentioning "mac collisions" which is technically incorrect

paul-maidment:
"The commit message doesn't mention the most important point, which is changing the "-c", "10" parameter to get more than one arping result, which with enough luck, will allow us to detect IP collision because we'll likely see responses arping responses from 2 different hosts sharing the same IP"
I felt this was covered by "Insufficient time is given to gather unicast replies, this commit changes that by altering arping parameters to make this work correctly." but I can change this to mention the specific parameters used

paul-maidment:
I have updated again with all the changes you requested

1: commit message now matches the PR
2: I mention the "-c" "10" change
3: I remove references to "mac collision" and replace with appropriate text

mkowalski (approved):
/lgtm
/hold

As soon as @omertuc signs-off on his comments, feel free to unhold

paul-maidment:
@omertuc Do my recent changes satisfy your concerns or do further changes need to be made?

src/connectivity_check/connectivity_check.go
```diff
@@ -401,14 +406,15 @@ func (c *connectivity) connectivityCheck(_ string, args ...string) (stdout strin
...
-	hostChan := make(chan *models.ConnectivityRemoteHost)
+	hostChan := make(chan models.ConnectivityRemoteHost)
 	for _, host := range params {
 		h := hostChecker{outgoingNICS: nics, host: host}
 		go c.checkHost(h, hostChan)
 	}
 	ret := models.ConnectivityReport{RemoteHosts: []*models.ConnectivityRemoteHost{}}
 	for i := 0; i != len(params); i++ {
-		ret.RemoteHosts = append(ret.RemoteHosts, <-hostChan)
+		h := <-hostChan
```
omertuc, line 416: What is the reason for this change?

src/connectivity_check/connectivity_check.go
```diff
@@ -353,19 +354,23 @@ func (c *connectivity) checkHost(conCheck connectivityCmd, outCh chan *models.Co
 	dataCh := make(chan any)
 	go c.l3CheckConnectivity(addresses, dataCh, conCheck)
 	go c.l2CheckConnectivity(dataCh, conCheck)
+	l2C := make(map[string]models.L2Connectivity)
```
omertuc, line 357: I think some background for the choice of using a map here is warranted and not so obvious

```suggestion
        // We use a map to de-duplicate arping responses from the same mac. They are redundant
        //  because the reason we're reporting more than one line of arping in the first place is to allow
        // the service to also detect devices in the network that have IP conflict with cluster hosts
	l2C := make(map[string]models.L2Connectivity)
```

src/connectivity_check/connectivity_check_test.go
```diff
@@ -483,13 +487,44 @@ var _ = Describe("check host parallel validation", func() {
 					},
 				},
 			},
+			simulateL2IPConflict: false,
+		},
+		{name: "Should correctly report on L2 IP conflicts",
```
omertuc, line 492: Can we also add a test that deduplication works?

src/connectivity_check/connectivity_check.go
```diff
@@ -353,19 +354,23 @@ func (c *connectivity) checkHost(conCheck connectivityCmd, outCh chan *models.Co
...
+			ret.L3Connectivity = append(ret.L3Connectivity, &value)
+		case models.L2Connectivity:
+			l2C[value.RemoteMac] = value
 		case done:
 			numDone++
 		}
-
 	}
-	outCh <- ret
+	for _, value := range l2C {
```
omertuc, line 369: Map iteration happens in random order. This will cause the response to (potentially) constantly change. Need to sort the array after constructing it

src/connectivity_check/connectivity_check.go
omertuc: Ignore me

src/connectivity_check/connectivity_check_test.go
paul-maidment: Added "Should deduplicate L2 entries and ensure consistent order of L2 entries" which uses the newly introduced test var `strictMatchingL2` to introduce a mode where the expected array must be an exact match.

This, combined with a change to the simulated data to add a duplicate arping response should be enough to prove both deduplication and appropriate order.

src/connectivity_check/connectivity_check_test.go
```diff
@@ -483,17 +491,79 @@ var _ = Describe("check host parallel validation", func() {
...
 	for i := range tests {
 		t := tests[i]
 		It(t.name, func() {
-			h := testHostChecker{outgoingNICS: t.nics, host: t.hosts, success: t.success}
+			h := testHostChecker{outgoingNICS: t.nics, host: t.hosts, success: t.success, simulateL2IPConflict: t.simulateL2IPConflict}
 			c := connectivity{dryRunConfig: &config.DryRunConfig{}}
 			go c.checkHost(h, hostChan)
 			r := <-hostChan
-			Expect(r.L2Connectivity).Should(ContainElements(t.expected.L2Connectivity))
+			if !t.strictMatchingL2 {
```
omertuc, line 562: On the surface this seems unnecessary, why not always be strict and adjust expectations in the rest of the tests?

paul-maidment: Mainly because the majority of the tests do not require this level of "exact matching". Moreover if we are strict about L2 then we have to be strict about L3 (by the principle you suggest) and I think will then require us to add lots more explicit (and potentially irrelevant) stuff to the tests.

The other way to go is to write this as it's own explicit test case and bypass the need to change the other test cases to be aware of this parameter.

Though I think what we have here is fine in all honesty.

PR #438, opened 2022-09-27
MGMT-11551: Retry when service fails

jhernand:
Currently if the service fails to respond the agent stops trying and waits forever. This patch changes it so that it will retry with an exponential backoff.

Related: https://issues.redhat.com/browse/MGMT-11551

### DIFF
--- a/src/commands/step_processor.go
+++ b/src/commands/step_processor.go
@@ -10,6 +10,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/cenkalti/backoff/v4"
+	"github.com/go-openapi/swag"
 	"github.com/patrickmn/go-cache"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -227,28 +229,27 @@ func (s *stepSession) getMountpointSourceDeviceFile() (string, error) {
 	return source, nil
 }
 
-func (s *stepSession) processSingleSession() (int64, string) {
+func (s *stepSession) processSingleSession() (delay time.Duration, exit bool, err error) {
 	s.Logger().Info("Query for next steps")
 	result, err := s.serviceAPI.GetNextSteps(&s.InventorySession)
 	if err != nil {
 		invalidateCache(s.stepCache)
 		switch err.(type) {
 		case *installer.V2GetNextStepsNotFound:
-			s.Logger().WithError(err).Errorf("Infra-env %s was not found in inventory or user is not authorized, going to sleep forever", s.agentConfig.InfraEnvID)
-			return -1, ""
+			err = errors.Wrapf(err, "infra-env %s was not found in inventory or user is not authorized", s.agentConfig.InfraEnvID)
 		case *installer.V2GetNextStepsUnauthorized:
-			s.Logger().WithError(err).Errorf("User is not authenticated to perform the operation, going to sleep forever")
-			return -1, ""
+			err = errors.Wrapf(err, "user is not authenticated to perform the operation")
 		case *installer.V2GetNextStepsForbidden:
-			s.Logger().WithError(err).Errorf("User is forbidden to perform the operation, going to sleep forever")
-			return -1, ""
+			err = errors.Wrapf(err, "user is forbidden to perform the operation")
 		default:
-			s.Logger().Warnf("Could not query next steps: %s", getErrorMessage(err))
+			err = fmt.Errorf("could not query next steps: %s", getErrorMessage(err))
 		}
-		return int64(s.agentConfig.IntervalSecs), ""
+		return
 	}
 	s.handleSteps(result)
-	return result.NextInstructionSeconds, *result.PostStepAction
+	delay = time.Duration(result.NextInstructionSeconds * int64(time.Second))
+	exit = swag.StringValue(result.PostStepAction) == models.StepsPostStepActionExit
+	return
 }
 
 // requiresRestart checks if the agent needs to be restarted after completing this step. Currently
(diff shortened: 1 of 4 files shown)


jhernand:
@omertuc I'd appreciate your review.

src/commands/step_processor.go
```diff
@@ -227,28 +229,27 @@ func (s *stepSession) getMountpointSourceDeviceFile() (string, error) {
...
+func (s *stepSession) processSingleSession() (delay time.Duration, exit bool, err error) {
 	s.Logger().Info("Query for next steps")
 	result, err := s.serviceAPI.GetNextSteps(&s.InventorySession)
 	if err != nil {
 		invalidateCache(s.stepCache)
 		switch err.(type) {
 		case *installer.V2GetNextStepsNotFound:
-			s.Logger().WithError(err).Errorf("Infra-env %s was not found in inventory or user is not authorized, going to sleep forever", s.agentConfig.InfraEnvID)
-			return -1, ""
+			err = errors.Wrapf(err, "infra-env %s was not found in inventory or user is not authorized, going to sleep forever", s.agentConfig.InfraEnvID)
```
omertuc, line 239:
```suggestion
			err = errors.Wrapf(err, "infra-env %s was not found in inventory or user is not authorized", s.agentConfig.InfraEnvID)
```

omertuc: Not just here

jhernand: Done.

src/commands/step_processor.go
```diff
@@ -276,20 +277,28 @@ func (s *stepSession) requiresRestart(reply models.StepReply) bool {
 func ProcessSteps(ctx context.Context, cancel context.CancelFunc, agentConfig *config.AgentConfig, toolRunnerFactory ToolRunnerFactory, wg *sync.WaitGroup, log log.FieldLogger) {
 	defer wg.Done()
 
-	var nextRunIn int64
 	c := newCache()
-	for afterStep := ""; afterStep != models.StepsPostStepActionExit; {
+
```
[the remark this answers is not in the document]
jhernand: I can add that comment, but note that what this code is doing is not exactly what you describe. The outer loop that already existed and that is controlled by the delay returned by the service remains the same. What is new is the inner loop to potentially retry every iteration of the outer loop. I can change the code to have just one loop with the delay calculated as you describe.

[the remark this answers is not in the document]
omertuc: I fail to see the difference between what you said and what I said, they sound equivalent to me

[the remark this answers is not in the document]
jhernand: There is no difference in behavior, only in the implementation details. I will add the comment.

src/commands/step_processor.go
```diff
@@ -276,20 +277,33 @@ func (s *stepSession) requiresRestart(reply models.StepReply) bool {
...
+		s := newSession(cancel, agentConfig, toolRunnerFactory, c, log)
+		var err error
+		delay, exit, err = s.processSingleSession()
+		return err
+	}
+	notify := func(err error, delay time.Duration) {
+		log.Errorf("Step processing failed, will try again in %s: %v", delay, err)
+	}
+	for !exit {
+		backOff := backoff.NewExponentialBackOff()
```
tsorya, line 299: It will be not enough to use http transport retry feature? 
https://github.com/openshift/assisted-installer/blob/master/src/inventory_client/inventory_client.go#L122 ?

jhernand: @tsorya I think it would be enough if all we want is retry at the HTTP level. If we do that then we will not be handling other non-HTTP related errors. I think that in this particular case the only thing that we wouldn't be doing between attempts is calling the "invalidateCache" function. Is that important?

tsorya: i don't think invalidating cache is relevant in this case. All we want is to retry in case service is not reachable in that case  we have nothing to invalidate as we didn't talk to service no?

jhernand: OK, I will change this to just retry at the HTTP level. Thanks @tsorya.

jhernand: @tsorya @omertuc the alternative implementation using HTTP level retries is in #442. Let me know which one you prefer and I will close the other one.

omertuc: While #442 looks "elegant" I think it's too silent / implicit / hard to follow than explicit retries, so I vote for #438

flaper87: I actually vote for #442 as I don't think we should retry on every failure. Some of them may hide real issues and it would be frustrating to just have the agent retry forever. Probably also not a good UX.

I hinted on my concerns about retrying on 401 errors [here](https://github.com/openshift/assisted-installer-agent/pull/442#discussion_r990582339) and I think there may be other similar errors that we want to skip

omertuc: The "#442 vs #438" argument is not about which failures we are retrying on, it's about whether we're using the silent rehttp transport layer library that silently retries without any indication, or whether we have a retry mechanism in our code that is explicit and includes a log for every failed request (which is what we want). The former is simpler but worse (no logs, implicit and hard to follow), the latter is more complex but better (logs, explicit and easy to follow).

More importantly, In both approaches we can theoretically pick and choose which errors get special treatment and which not.

I agree with you that retrying on a 401 code is not necessarily ideal and it's up for debate whether it's the behavior we want or not - but that goes completely against the ticket that this PR is trying to solve, the ticket that says we never sleep forever, no matter what the error is. 

I don't think this PR / the other PR is is the right place to have this debate, as they're just implementing the Jira ticket. If we don't like the Jira ticket (which is completely fair) let's take the discussion to Jira, Slack or the architecture forum meeting

flaper87: >  or whether we have a retry mechanism in our code that is explicit and includes a log for every failed request. 

To some extent, it is about what failures we are retrying/catching because the mechamism as it is in this PR, we are not differentiating.

To @jhernand point on whether retrying at HTTP level or not, I think I agree with @tsorya that using the HTTP transport is probably what we want in this case. Or rather, the HTTP retries should probably be separated from other retries/failures as they are not the same. One is an error coming from the server, another one is an error that could happen locally..

>  but that goes completely against the ticket that this PR is trying to solve, the ticket that says we never sleep forever, no matter what the error is.

Then we probably need 2 PRs?. One PR implements the retries and a second PR implements a logic for which we fail (and don't sleep) for some errors. I just don't think adding 401 here "just because the ticket says so" is the right approach. (I know that's not what's being suggested).

Small derail from the actual topic of this PR:

> I don't think this PR / the other PR is is the right place to have this debate, as they're just implementing the Jira ticket. 

I mentioned this in #442 and I will mention here again. The Jira ticket you are referring to is not visible to everyone looking at this PR. The implementation does something and opens up things for discussion on this platform so, it is a place where we can have (and probably should have) the discussion. Let's make sure we report back everything we discuss. Let's schedule a meeting if that's really necessary.

omertuc: >To some extent, it is about what failures we are retrying/catching because the mechamism as it is in this PR, we are not differentiating.

Right, that's why I said:

> More importantly, In both approaches we can **theoretically** pick and choose which errors get special treatment and which not.

This "picking which errors" difference between the two PRs is I believe just an accident / implementation error and not the point of having two PRs. The reason for having two PRs is manual implementation vs using `rehttp`, which is why I don't see the reason to prefer one PR over the other just because of "here we pick which errors and here we don't", we can pick which errors in both PRs.

>To @jhernand point on whether retrying at HTTP level or not, I think I agree with @tsorya that using the HTTP transport is probably what we want in this case

But `rehttp` is silent and implicit. Why would we want to have retries without any logs? 

>Or rather, the HTTP retries should probably be separated from other retries/failures as they are not the same. One is an error coming from the server, another one is an error that could happen locally..
Then we probably need 2 PRs?. One PR implements the retries and a second PR implements a logic for which we fail (and don't sleep) for some errors. I just don't think adding 401 here "just because the ticket says so" is the right approach. (I know that's not what's being suggested).

Not sure I understand this point, what failures?

>I mentioned this in https://github.com/openshift/assisted-installer-agent/pull/442 and I will mention here again. The Jira ticket you are referring to is not visible to everyone looking at this PR.

I'm all for public discussions and managing our tickets on GitHub Issues, but we don't do that, we manage our tasks on Jira and that's where we plan what we need to do, currently.

> The implementation does something and opens up things for discussion on this platform 

I 100% agree 

>so, it is a place where we can have (and probably should have) the discussion. 

We can initiate the discussion where the implementation is (as we did), but right now your it seems your argument is "the jira ticket is wrong" (jira says don't discriminate between any errors, you say we should discriminate) and I think that calls for going back to the drawing board, as we don't seem to agree and I want to involve more people that were involved in the original decision to make this ticket.

Let's have a short session about this in the architecture forum, no need for a full meeting

flaper87:
/hold 

I do not think we should merge this while we are still skipping on all the errors and before having a proper discussion on this. 

I'm going to be offline for a couple of weeks so I will let you all decide and override my hold if needed. Just want to state here that we should not merge this before discussing what errors we want to retry on. At least, let me have this and remove 401 from the retries :smile:  :pray:

jhernand:
> /hold
> 
> I do not think we should merge this while we are still skipping on all the errors and before having a proper discussion on this.

@flaper87 in light of the discussion we had last week about this would it be OK to move forward with this change? If so please unhold.

PR #439, opened 2022-09-30
MGMT-8329: Disks with mounted parititions are ineligible

jhernand:
Currently disks with mounted partitions are considered as eligible for installation, but they aren't. This patch fixes that.

This is the result in the UI:

image

image

Related: https://issues.redhat.com/browse/MGMT-8329

### DIFF
--- a/src/inventory/disks.go
+++ b/src/inventory/disks.go
@@ -241,18 +241,39 @@ func (d *disks) checkEligibility(disk *ghw.Disk) (notEligibleReasons []string, i
 		notEligibleReasons = append(notEligibleReasons, "Disk is an LVM logical volume")
 	}
 
-	if funk.Contains(funk.Map(disk.Partitions, func(p *ghw.Partition) bool {
-		return p.Type == "iso9660"
-	}), true) {
-		notEligibleReasons = append(notEligibleReasons, "Disk appears to be an ISO installation media (has partition with type iso9660)")
-		isInstallationMedia = true
-	}
+	// Check disk partitions for type, name, and mount points:
+	for _, partition := range disk.Partitions {
+		if partition.Type == "iso9660" {
+			notEligibleReasons = append(
+				notEligibleReasons,
+				"Disk appears to be an ISO installation media (has partition with "+
+					"type iso9660)",
+			)
+			isInstallationMedia = true
+		}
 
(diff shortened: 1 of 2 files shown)


carbonin (approved):

src/inventory/disks.go
```diff
@@ -255,6 +255,23 @@ func (d *disks) checkEligibility(disk *ghw.Disk) (notEligibleReasons []string, i
 		isInstallationMedia = true
 	}
 
+	// We also want to check that disks don't have partitions mounted, but only for disks that
+	// aren't the installation media because we already know that those have partitions mounted
+	// and we don't want to have duplicated reasons.
+	if !isInstallationMedia {
```
carbonin, line 261: Would it be a bit cleaner to combine these two checks?

jhernand: @carbonin I am not sure to understand what is your suggestion. Can you elaborate?

carbonin: The previous two checks are also iterating through the partitions on the disks to determine if they are installation media and here you're explicitly excluding the cases where `isInstallationMedia` is set already.

I'm thinking it would be more clear to loop through the partitions once and check for all of these conditions at at the same time. Something like:

```go
// check disk partitions for type, name, and mount points
for _, p := range disk.Partitions {
  if p.Type == "iso9660" {
    notEligibleReasons = append(notEligibleReasons, "Disk appears to be an ISO installation media (has partition with type iso9660)")
    isInstallationMedia = true
  }

  if strings.HasSuffix(p.MountPoint, "iso") {
    notEligibleReasons = append(notEligibleReasons, "Disk appears to be an ISO installation media (has partition with mountpoint suffix iso)")\
    isInstallationMedia = true
  }

  if isInstallationMedia {
    continue
  }

  if partition.MountPoint != "" {
    notEligibleReasons = append(
      notEligibleReasons,
      fmt.Sprintf(
        "Disk has partition '%s' mounted on '%s'",
        partition.Name, partition.MountPoint,
      ),
  }
}
```

jhernand: OK, makes sense, I will do that, thanks @carbonin.

jhernand: Done.

PR #442, opened 2022-10-07
MGMT-11551: Retry when service fails

jhernand:
This patch changes the agent so that it will retry with an exponential backoff when the server respods with 401, 503 or 504.

Related: https://issues.redhat.com/browse/MGMT-11551

### DIFF
--- a/src/session/inventory_session.go
+++ b/src/session/inventory_session.go
@@ -148,14 +148,37 @@ func createBmInventoryClient(agentConfig *config.AgentConfig, inventoryUrl strin
 		},
 	})
 
+	// This function transforms a regular delay function into one that does the same thing but
+	// also writes a log message indicating that the request will be retried.
+	delayLog := func(delayFn rehttp.DelayFn) rehttp.DelayFn {
+		return func(attempt rehttp.Attempt) time.Duration {
+			delay := delayFn(attempt)
+			logrus.WithFields(logrus.Fields{
+				"method":  attempt.Request.Method,
+				"url":     attempt.Request.URL,
+				"status":  attempt.Response.StatusCode,
+				"error":   attempt.Error,
+				"attempt": fmt.Sprintf("%d of %d", attempt.Index+1, retries+1),
+				"delay":   delay,
+			}).Info("Request will be retried")
+			return delay
+		}
+	}
+
 	// Add retry settings
 	tr := rehttp.NewTransport(
 		transport,
 		rehttp.RetryAll(
 			rehttp.RetryMaxRetries(retries),
-			rehttp.RetryTemporaryErr(),
+			rehttp.RetryAny(
+				rehttp.RetryTemporaryErr(),
+				rehttp.RetryStatuses(
+					http.StatusServiceUnavailable,
+					http.StatusGatewayTimeout,
+				),
+			),
 		),
-		rehttp.ExpJitterDelay(minDelay, maxDelay),
+		delayLog(rehttp.ExpJitterDelay(minDelay, maxDelay)),
 	)
 
 	clientConfig.Transport = tr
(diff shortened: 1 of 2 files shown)


src/session/inventory_session.go
```diff
@@ -153,7 +153,14 @@ func createBmInventoryClient(agentConfig *config.AgentConfig, inventoryUrl strin
 		transport,
 		rehttp.RetryAll(
 			rehttp.RetryMaxRetries(retries),
-			rehttp.RetryTemporaryErr(),
+			rehttp.RetryAny(
+				rehttp.RetryTemporaryErr(),
+				rehttp.RetryStatuses(
+					http.StatusUnauthorized,
```
flaper87, line 159: I'm wondering if we really want to retry on 401s. A retry is useful for eventual reconciliation. I'm unsure when a 401 may occur that could be solved by various exponential retries. Could you elaborate on this?

omertuc: Treating these errors as irrecoverable is our existing behavior and is what this PR (and it's alternative #438) are trying to change.

The Jira ticket goes into the details, but the tl;dr is we've come to the conclusion we don't **really** understand the nature of those errors, and they may actually happen due to temporary circumstances, so it's better to have exponential backoff than simply freezing forever.

jhernand: 401 is quite frequent when the client uses token authentication and the token is close to expire: the client thinks it is still valid but the time it reaches the server it is already expired. Time skew between client and server makes that even more frequent. Not sure what kind of authentication is used for this communication between the agent and the service. If it isn't time sensitive then this isn't needed or helpful.

jhernand:
Note that this is an alternative to #438, please check that if you are interested on the subject.

src/session/inventory_session.go
flaper87: > Treating these errors as irrecoverable is our existing behavior 

This part I know 

> The Jira ticket goes into the details,

Part of the reason I was asking is that the Jira ticket is not public,  which is why I always stress so much about having PRs detail explicitly what they do and why.. The other part is realted and is that I did not have access to Jira when I first reviewed the PR :) 

> we don't really understand the nature of those errors,

And this is where I'm getting at with my questions. It is clear that we *don't* understand why they happen and that worries me. A 502, 504, or some other availability error I can understand and I can see how a Retry would help but, I don't understand why we should retry on a 401 `StatusUnauthorized` error. 

I think we should let the agent fail on authentication errors. It may be someone is trying to attack, or using an expired token, or who knows what. An authorization error does not sound like something that the service/agent would eventually recover from.

@jhernand thanks for the details. To me, exponential retries on authorization failures are a bit weird. That said, assuming this is what we want, I think they are useful *only* if the agent will actually recover from it eventually. I don't think that's the case and I guess that's the gist of my question:

- Can we recover from these authorization issues automatically? 
- Does the agent renew the token when it expires? (IIRC, it doesn't)
- Does it recover from authentication issues by just synchronizing clocks? (a bit weird but :shrug: )

flaper87 (commented):
> Note that this is an alternative to #438, please check that if you are interested on the subject.

Will do! Let's close this one if it's not the preferred solution

jhernand:
@omertuc @flaper87 the new version of the patch doesn't retry for 401 errors and writes a log message when a request is retried. I think that addresses your concerns. Please take another look.

flaper87:
This works for me! I will wait for @omertuc feedback before approving.

flaper87:
> We've agreed to discuss this in the architecture forum

I couldn't attend the last architecture forum call. Was this discussed?

jhernand:
> I couldn't attend the last architecture forum call. Was this discussed?

No, it wasn't discussed, all the time was dedicated to other items.

omertuc:
Anyway this better than what we have today. So let's merge this but let's also not close the ticket

omertuc:
>all the time was dedicated to other items.

Actually it wasn't on the agenda, just in the future topics section

PR #449, opened 2022-11-03
MGMT-8460: Automatically rename discovered hosts

jhernand:
Currently hosts that have a forbidden name like `hostname` need to be manually renamed by users before proceeding to the installation of the cluster. That isn't convenient for users. To avoid it this patch changes the code so that the agent will replace those names with automatically generated ones from a MAC address of a network interface card of the host that has a non-local IP address.

Related: https://issues.redhat.com/browse/MGMT-8460

### DIFF
--- a/src/inventory/inventory.go
+++ b/src/inventory/inventory.go
@@ -2,10 +2,14 @@ package inventory
 
 import (
 	"encoding/json"
+	"net"
+	"sort"
+	"strings"
 
 	"github.com/openshift/assisted-installer-agent/src/config"
 	"github.com/openshift/assisted-installer-agent/src/util"
 	"github.com/openshift/assisted-service/models"
+	"github.com/sirupsen/logrus"
 )
 
 func ReadInventory(subprocessConfig *config.SubprocessConfig, c *Options) *models.Inventory {
@@ -24,6 +28,7 @@ func ReadInventory(subprocessConfig *config.SubprocessConfig, c *Options) *model
 		Routes:       GetRoutes(d),
 		TpmVersion:   GetTPM(d),
 	}
+	processInventory(&ret)
 	return &ret
 }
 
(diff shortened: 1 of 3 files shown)


src/inventory/inventory.go
```diff
@@ -24,6 +28,7 @@ func ReadInventory(subprocessConfig *config.SubprocessConfig, c *Options) *model
 		Routes:       GetRoutes(d),
 		TpmVersion:   GetTPM(d),
 	}
+	processInventory(&ret)
```
omertuc, line 31: Any reason this is done as a "post-processing" transformation of the inventory rather than simply as part of `GetHostname`?

omertuc: Oh because it combines information from the NICs, okay. Nevermind

jhernand: The only reason is that `GetHostname` doesn't receive the network interfaces. I can change that.

src/inventory/inventory.go
```diff
@@ -41,3 +46,111 @@ func CreateInventoryInfo(subprocessConfig *config.SubprocessConfig) []byte {
...
+	sort.Slice(nics, func(i, j int) bool {
+		return strings.Compare(nics[i].Name, nics[j].Name) < 0
+	})
+
+	// Find the first NIC that has a MAC address an a global IP address:
+	for _, nic := range nics {
+		hasMAC := nic.MacAddress != ""
+		hasV4 := false
+		for _, ip := range nic.IPV4Addresses {
+			hasV4, err = isGlobalCIDR(ip)
```
omertuc, line 120: Can `nic.Type == "physical"` be used instead?

jhernand: I'd say that it can be added as an additional criteria, but not to replace the existing: a physical network device can have a non-global IP address assigned, and we want to exclude those.

jhernand: Done, added the check.

src/inventory/inventory.go
```diff
@@ -41,3 +46,112 @@ func CreateInventoryInfo(subprocessConfig *config.SubprocessConfig) []byte {
...
+	}
+	result = strings.ToLower(strings.ReplaceAll(nic.MacAddress, ":", "-"))
+	return
+}
+
+// findUsableNIC returns a physical network interface card of the given host inventory that has a
+// MAC address and a non-local IP address. Returns nil if the host doesn't have such network
+// interface card, and an error if the process fails, for example if some of the IP addresses of the
+// host can't be parsed.
+func findUsableNIC(inventory *models.Inventory) (result *models.Interface, err error) {
```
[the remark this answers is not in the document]
jhernand: The contract is similar, but the implementation is quite different: this one checks for IP addresses and physical vs. virtual interfaces. I assume those checks wheren't a requirement when creating that function, so I'd say we want another function.

omertuc:
/lgtm

@avishayt you can unhold if you want

PR #458, opened 2022-11-28
WIP MGMT-12714: Unblock software RAID devices

avishayt:
Also allow removable disks because assisted-service was changed to make that a warning rather than error.


omertuc:
/uncc @omertuc 

recc when not wip

PR #459, opened 2022-11-30
MGMT-12760: Don't panic when retrying after no response

jhernand:
Currently the code that retries requests writes a log message that contains the response status code. But that causes a panic if there is no response, for example if the server didn't respond at all or there was a timeout. This patch fixes that issue.

Related: https://issues.redhat.com/browse/MGMT-12760


jhernand:
@tsorya please take a look.

jhernand:
This is a fix for a bug that I introduced in #442.

PR #462, opened 2022-12-01
NO-ISSUE: Stop retry when backoff expires or context is cancelled

jhernand:
The retry loop that was introduced in a previous patch doesn't stop when the backoff expires and the error still happens, and also when the service explicitly asks the agent to upgrade and restart. This patch fixes that.


src/commands/step_processor.go
```diff
@@ -300,9 +300,11 @@ func ProcessSteps(ctx context.Context, cancel context.CancelFunc, agentConfig *c
 		err := backoff.RetryNotify(operation, backOff, notify)
 		if err != nil {
 			log.Errorf("Step processing failed, will exit: %v", err)
+			return
 		}
 		select {
 		case <-ctx.Done():
+			return
```
filanov, line 307: maybe a log in this case?

jhernand: Done.

PR #464, opened 2022-12-01
MGMT-12339: disk_speed_check: escape colons when calling fio (#427)

javipolo:
With https://github.com/openshift/assisted-installer/pull/583 we can reapply @vrutkovs's fix to disk_speed_check so FIO runs properly 

fio's filename can be a colon-separated list of devices to test. This however breaks some paths with colons, so these need to be escaped. See https://fio.readthedocs.io/en/latest/fio_doc.html#cmdoption-arg-filename

Previously `fio` was creating `/dev/disk/by-path/pci-0000` for `/dev/disk/by-path/pci-0000:06:0000.0` disk, which was a) testing the wrong disk speed and
b) could fill up the root partition

/cc @vrutkovs 
/cc @osherdp

### DIFF
--- a/src/disk_speed_check/disk_speed_check.go
+++ b/src/disk_speed_check/disk_speed_check.go
@@ -2,6 +2,7 @@ package disk_speed_check
 
 import (
 	"encoding/json"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


osherdp (approved):

vrutkovs (approved):
/approve
/lgtm

PR #467, opened 2022-12-05
MGMT-12850: Add host id header

jhernand:
This patch changes the agent so that it sends the host identifier in the `X-Host-ID` header. This will be used for rate limiting in the server side.

Related: https://issues.redhat.com/browse/MGMT-12850
Related: https://github.com/openshift/assisted-service/pull/4724

### DIFF
--- a/src/commands/service_api.go
+++ b/src/commands/service_api.go
@@ -32,6 +32,7 @@ func (v *v2ServiceAPI) RegisterHost(s *session.InventorySession) (*models.HostRe
 	}
 
 	params := &installer.V2RegisterHostParams{
+		XHostID:               &hostID,
 		InfraEnvID:            strfmt.UUID(v.agentConfig.InfraEnvID),
 		DiscoveryAgentVersion: &v.agentConfig.AgentVersion,
 		NewHostParams: &models.HostCreateParams{
@@ -48,8 +49,10 @@ func (v *v2ServiceAPI) RegisterHost(s *session.InventorySession) (*models.HostRe
 }
 
 func (v *v2ServiceAPI) GetNextSteps(s *session.InventorySession) (*models.Steps, error) {
+	hostID := strfmt.UUID(v.agentConfig.HostID)
 	params := installer.V2GetNextStepsParams{
-		HostID:                strfmt.UUID(v.agentConfig.HostID),
+		XHostID:               &hostID,
+		HostID:                hostID,
 		InfraEnvID:            strfmt.UUID(v.agentConfig.InfraEnvID),
 		DiscoveryAgentVersion: &v.agentConfig.AgentVersion,
 		Timestamp:             swag.Int64(time.Now().Unix()),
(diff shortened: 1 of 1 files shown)


src/commands/service_api.go
```diff
@@ -48,8 +49,10 @@ func (v *v2ServiceAPI) RegisterHost(s *session.InventorySession) (*models.HostRe
 }
 
 func (v *v2ServiceAPI) GetNextSteps(s *session.InventorySession) (*models.Steps, error) {
+	hostID := strfmt.UUID(v.agentConfig.HostID)
 	params := installer.V2GetNextStepsParams{
-		HostID:                strfmt.UUID(v.agentConfig.HostID),
+		XHostID:               &hostID,
+		HostID:                hostID,
```
gamli75, line 55: why do we need both HostID & XHostID?

jhernand: `XhostID` goes in the `X-Host-ID` HTTP header, and `HostID` goes in the request path. It is not strictly necessary to have the header here, but it is convenient to simplify the [Envoy](https://github.com/openshift/assisted-service/pull/4725) and [Limitador](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/53343) configuration. The rate limiting configuration in the Envoy side specifies what pieces of information are sent to the rate limiting service. For example, for this endpoint this is what I am suggesting in [#4725](https://github.com/openshift/assisted-service/pull/4725) :

```yaml
# Check rate limits for requests to get next steps and post results:
- name: get_and_post_steps
  match:
    safe_regex:
      regex: "^/api/assisted-install/v2/infra-envs/.*/hosts/.*/instructions$"
  route:
    cluster: assisted_service
    rate_limits:
    - actions:
      - generic_key:      
          descriptor_key: operation
          descriptor_value: get_and_post_steps
      - request_headers:
          header_name: x-host-id
          descriptor_key: host_id
```

That is asking Envoy so send the following pieces of information (_descriptors_ in Envoy rate limiting jargon) to the rate limiting service:

```
(operation, get_and_post_steps)
(host_id, <the value of the X-Host-ID header>)
```

I am also suggesting a very similar thing for the host registration endpoint:

```yaml
# Check rate limits for requests to register hosts: 
 - name: post_host
   match:
     headers:
     - name: :method    
       string_match:      
         exact: POST
     safe_regex:
       regex: "^/api/assisted-install/v2/infra-envs/.*/hosts$"
   route:
     cluster: assisted_service
     rate_limits:
     - actions:
       - generic_key:
           descriptor_key: operation
           descriptor_value: post_host
       - request_headers: 
           header_name: x-host-id
           descriptor_key: host_id
```

But in that case, for host registration, the header would be required because Envoy doesn't know how to extract it from the request body.

If we don't add that `X-Host-ID` header then we would need to send different pieces of information, in particular we would need to send the complete request path instead of the header:

```yaml
 - generic_key:
     descriptor_key: operation
     descriptor_value: get_and_post_steps
 - request_headers: 
     header_name: :path
     descriptor_key: path
```

The actual rate limits in Limitador would also need to be slightly different: ones using the `host_id` descriptor, and others using the `path` descriptor. Not a big deal, just a bit more complicated and potentially confusing.

gamli75: thanks a lot for the detailed answer

PR #475, opened 2023-01-03
MGMT-12273: Log more from the PreNetworkConfigScript

paul-maidment:
In situations where users are struggling to determine why their network configuration is not taking effect it would be helpful to log a bit more verbosely from the service that sets up the network config files. Specifically if a host does not match any of the provided mappings we currently write "None of host directories are a match for the current host" which is not particularly helpful. It should at least say something about possible mac address mismatches. We probably could also find a few more places to log something useful.

This PR attempts to address these concerns by adding more verbose logging.


rccrdpccl:
I'm not extremely familiar with the codebase, but wouldn't it be best to output error and warning messages  to stderr instead?

vendor/github.com/openshift/assisted-service/internal/constants/scripts.go
```diff
@@ -34,6 +35,7 @@ MAC_NIC_MAPPING_FILE=mac_interface.ini
 
 if [ ! -d "$NMCONNECTIONS_DIR" ]
 then
+  echo "Error (exiting): Expected to find the directory '$NMCONNECTIONS_DIR' on the host but this was not present."
   exit 0
```
rccrdpccl, line 39: Is it an error or an early finish without error? I am asking because it exists with exit code 0

omertuc:
Why are we modifying the copy of the assisted-service repo in the vendor directory in the agent repo instead of the assisted-service repo itself?

paul-maidment:
Accidentally worked against the wrong repo, closing this PR completely.

PR #476, opened 2023-01-03
MGMT-12535: Add step verify vips

ori-amizur:
Free addresses scan consume a lot of resources and
scans big networks in order to know if the designated IP addresses
for the VIPs are free.
This change adds a step that receives the list of IPs to check and checks them one by
one.

/cc @filanov 
/cc @avishayt


src/vips_verifier/vips_verifier.go
```diff
@@ -0,0 +1,118 @@
...
+		VipType:      vipToVerify.VipType,
+	}, nil
+}
+
+type vipVerifier struct {
+	exe Executer
+}
+
+func (v *vipVerifier) verifyVip(vipToVerify *models.VerifyVip) (*models.VerifiedVip, error) {
+	var (
```
omertuc, line 51: Skip on dry run?

omertuc: Oh nevermind missed that this is just a function for one VIP

src/vips_verifier/vips_verifier.go
```diff
@@ -0,0 +1,118 @@
...
+func VerifyVips(dryRunConfig *config.DryRunConfig, _ string, args ...string) (stdout string, stderr string, exitCode int) {
+	if len(args) != 1 {
+		return bailOut(fmt.Sprintf("expected 1 argument.  Received %d", len(args)))
+	}
+	var verifier VipVerifier
+	if dryRunConfig.DryRunEnabled {
+		verifier = &dryVipVerifier{}
+	} else {
+		verifier = &vipVerifier{exe: &executer{}}
+	}
```
omertuc, line 116: Can be shortened a bit

```suggestion
	var verifier VipVerifier = &vipVerifier{exe: &executer{}}
	if dryRunConfig.DryRunEnabled {
		verifier = &dryVipVerifier{}
	}
```

omertuc:
Why UI/UX? This is a performance improvement, the user doesn't know anything about it

avishayt:
> Why UI/UX? This is a performance improvement, the user doesn't know anything about it

With this PR, the VIP verification is moving from a sync/immediate check to an async validation.  This means that users will need to wait 1-2 minutes after entering a VIP for full validation.

ori-amizur:
> > Why UI/UX? This is a performance improvement, the user doesn't know anything about it
> 
> With this PR, the VIP verification is moving from a sync/immediate check to an async validation. This means that users will need to wait 1-2 minutes after entering a VIP for full validation.

The users will not have to wait more than they wait today in order to add VIPs.  They just will be validated up to a minute after VIPs are entered instead of immediate validation.

src/vips_verifier/vips_verifier.go
ori-amizur: OK

omertuc:
> > Why UI/UX? This is a performance improvement, the user doesn't know anything about it
> 
> With this PR, the VIP verification is moving from a sync/immediate check to an async validation. This means that users will need to wait 1-2 minutes after entering a VIP for full validation.

Fair point

ori-amizur:
> > > Why UI/UX? This is a performance improvement, the user doesn't know anything about it
> > 
> > 
> > With this PR, the VIP verification is moving from a sync/immediate check to an async validation. This means that users will need to wait 1-2 minutes after entering a VIP for full validation.
> 
> Fair point

@avishayt this change doesn't invoke the step - it is done in the service.
So if the desired solution should include this step that possibly it will be invoked depending on some conditions 
(i.e network size, dual stack, kube API), then we should go forward with this PR.
Otherwise, it should be cancelled.

avishayt:
/unhold
Right, we do want this behavior at least for some cases.  Can you add an ipv6 test?

ori-amizur:
> /unhold Right, we do want this behavior at least for some cases. Can you add an ipv6 test?

Added

PR #478, opened 2023-01-06
MGMT-13111: Stop agent on `404 Not Found`

jhernand:
Currently when the user deletes a host the next step runner receives a `404 Not Found` response code from the service, logs it and tries again with exponential backoff. After approximately 15 minutes it stops retrying and exits, and then the main process of the agent will register the host again. The net result is that the host is automatically re-added after those 15 minutes, and we don't want that. To avoid that this patch changes the next step runner so that it stops the agent using the equivalent of `systemctl stop agent.service`. To register the host again the user will have to start the agent or else reboot the host.

Related: https://issues.redhat.com/browse/MGMT-13111

### DIFF
--- a/src/commands/step_processor.go
+++ b/src/commands/step_processor.go
@@ -18,6 +18,7 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	"github.com/openshift/assisted-installer-agent/src/config"
+	"github.com/openshift/assisted-installer-agent/src/service"
 	"github.com/openshift/assisted-installer-agent/src/session"
 	"github.com/openshift/assisted-installer-agent/src/util"
 	"github.com/openshift/assisted-service/client/installer"
@@ -236,7 +237,15 @@ func (s *stepSession) processSingleSession() (delay time.Duration, exit bool, er
 		invalidateCache(s.stepCache)
 		switch err.(type) {
 		case *installer.V2GetNextStepsNotFound:
-			err = errors.Wrapf(err, "infra-env %s was not found in inventory or user is not authorized", s.agentConfig.InfraEnvID)
+			s.Logger().WithError(err).Errorf(
+				"infra-env %s or host %s was not found in inventory or user is "+
+					"not authorized, will stop the agent",
+				s.agentConfig.InfraEnvID, s.agentConfig.HostID,
+			)
+			err = service.Stop()
+			if err != nil {
+				s.Logger().WithError(err).Errorf("Failed to stop the agent")
+			}
 		case *installer.V2GetNextStepsUnauthorized:
 			err = errors.Wrapf(err, "user is not authenticated to perform the operation")
 		case *installer.V2GetNextStepsForbidden:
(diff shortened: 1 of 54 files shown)


src/service/stop.go
```diff
@@ -0,0 +1,44 @@
...
+			"failed to stop systemd unit '%s', job is '%d' and result is '%s'",
+			systemdUnitName, systemdJob, systemdResult,
+		)
+	}
+
+	return nil
+}
+
+// systemdUnitName is the name of the systemd unit of the agent.
+const systemdUnitName = "agent.service"
```
carbonin, line 44: Nit, but can you define this before the function? It just seems strange to me to put it here.

carbonin (changes requested):
Doesn't this effectively revert the decisions made in https://issues.redhat.com/browse/MGMT-11551?

In that case it was a 404 that caused the agent to sleep forever and we decided that _wasn't_ the behavior we wanted.

Maybe we should at least hit the timeout before stopping the agent rather than doing the stop on the first 404 we get?

avishayt:
Why does it exit after 15 minutes?

jhernand:
@carbonin yes, it partially reverts those decisions, in particular for 404.

@avishayt it exists after 15 minutes because that is the maximum total retry time.

If we don't want to stop the agent in this case we can change that maximum time to infinite.

src/commands/step_processor.go
```diff
@@ -236,7 +237,15 @@ func (s *stepSession) processSingleSession() (delay time.Duration, exit bool, er
 		invalidateCache(s.stepCache)
 		switch err.(type) {
 		case *installer.V2GetNextStepsNotFound:
-			err = errors.Wrapf(err, "infra-env %s was not found in inventory or user is not authorized", s.agentConfig.InfraEnvID)
+			s.Logger().WithError(err).Errorf(
+				"infra-env %s or host %s was not found in inventory or user is "+
+					"not authorized, will stop the agent",
+				s.agentConfig.InfraEnvID, s.agentConfig.HostID,
+			)
+			err = service.Stop()
```
carbonin, line 245: This looks like the agent service is being stopped as soon as we get a 404 from the service. Where is the 15 minutes logic?

src/service/stop.go
jhernand: Done.

carbonin:
I still think we should keep the 15 minutes retry around, but stop the whole agent process after that time rather than the previous behavior which would lead to it restarting.

Or am I misunderstanding something?

avishayt:
The issue of backup/DR has come up recently.  In that case, it makes sense that the agent would receive 404 errors, and then the service would recover.  I suggest we keep the behavior of retrying forever for 404.

jhernand:
> The issue of backup/DR has come up recently. In that case, it makes sense that the agent would receive 404 errors, and then the service would recover. I suggest we keep the behavior of retrying forever for 404.

Then this change doesn't make sense, I am closing it.

PR #483, opened 2023-01-19
MULTIARCH-3239 -- Add multipath FC disk check for Power

cs-zhang:
Add multipath FC disk check to enable RHCOS installation for PowerVM with SAN disk.

Signed-off-by: CS Zhang <zhangcho@us.ibm.com>


CrystalChun (approved):
/lgtm

PR #484, opened 2023-01-22
MGMT-13420: Adding cleanup of next-step-runner before starting it.

tsorya:
[MGMT-13420](https://issues.redhat.com//browse/MGMT-13420): Adding cleanup of next-step-runner before starting it.
In 4.12 for unknown reason podman fails to cleanup next-step-runner on agent restart and --rm doesn't help. To workaround the problem we are adding cleanup before starting it.

### DIFF
--- a/src/commands/actions/next_step_runner.go
+++ b/src/commands/actions/next_step_runner.go
@@ -2,6 +2,7 @@ package actions
 
 import (
 	"fmt"
+	log "github.com/sirupsen/logrus"
 	"strconv"
 
 	"github.com/openshift/assisted-installer-agent/src/util"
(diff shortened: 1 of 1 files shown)


src/commands/actions/next_step_runner.go
```diff
@@ -12,6 +13,8 @@ import (
 	"github.com/openshift/assisted-service/models"
 )
 
+const containerName = "next-step-runner"
```
osherdp, line 16: let's reuse this var on line 78, to make sure we don't change it only on one place (and forget the other place)

osherdp (changes requested):

osherdp (approved):

omertuc:
Please check if this also resolves https://issues.redhat.com/browse/MGMT-13183

tsorya:
@omertuc seems like it should handle it

tsorya:
/retitle MGMT-13427:Adding cleanup of next-step-runner before starting it

PR #340, opened 2022-04-26
MGMT-9561: VSphere platform lacking setting disk.EnableUUID guidance

michaellevy101:

PR #497, opened 2023-02-21
OCPBUGS-7495: add sg3_utils package in the Docker file

mhanss:
On vSphere cloud, disk UUID verification is failing due to sg3_utils package is missing in the docker file.

### DIFF
--- a/Dockerfile.ocp
+++ b/Dockerfile.ocp
@@ -19,4 +19,4 @@ COPY --from=builder /go/src/github.com/openshift/assisted-installer-agent/script
 
 RUN if [ "$(arch)" -eq "x86_64" ]; then dnf install -y biosdevname dmidecode; fi
 RUN if [ "$(arch)" -eq "aarch64" ]; then dnf install -y dmidecode; fi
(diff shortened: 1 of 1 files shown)


zaneb:
Looks like adding this was missed in #340. That was not long after this Dockerfile was added, so probably most folks did not know about it yet.
/lgtm
/cherry-pick release-4.12

omertuc:
Please link the corresponding downstream Dockerfile PR

gamli75:
@mhanss I think we should handle this fix also in https://github.com/openshift/assisted-installer-agent/blob/master/Dockerfile.assisted_installer_agent and in the DS image because we probably have the same issue in SaaS.
@eliorerz please check it and if the answer is yes - we should clone this issue for the SaaS as well and fix it

zaneb:
> @mhanss I think we should handle this fix also in https://github.com/openshift/assisted-installer-agent/blob/master/Dockerfile.assisted_installer_agent and in the DS image because we probably have the same issue in SaaS. @eliorerz please check it and if the answer is yes - we should clone this issue for the SaaS as well and fix it

It's already there. This is copying the fix in #340 from the SaaS to the OCP Dockerfile.

eliorerz (approved):

PR #498, opened 2023-02-22
Make the agent team be the OWNERS of the Dockerfile.ocp file

omertuc:
Also there's no "downstream" for this Dockerfile so that label is
unnecessary

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -27,3 +27,9 @@ aliases:
     - yevgeny-shnaidman
     - lranjbar
     - ybettan
+  agent_team:
+    - andfasano
+    - bfournie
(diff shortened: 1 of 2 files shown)


OWNERS_ALIASES
```diff
@@ -27,3 +27,9 @@ aliases:
     - yevgeny-shnaidman
     - lranjbar
     - ybettan
+  agent_team:
+    - andfasano
+    - bfournie
+    - lranjbar
```
zaneb, line 33: Lisa has moved on from the agent team, but pawanpinjarkar is still here.

zaneb:
Awesome, thank you!
/lgtm

PR #501, opened 2023-02-23
MGMT-13758: In connectivity check, ARPING checks should not run on nics without valid IPv4 address

ori-amizur:
When check L2 IPv4 with ARPING, testing on interfaces that do not have valid IPv4 addresses causes the arp table to have duplicate entries, which may cause troubles in connectivity checks and packet losses. This change ensures that only interfaces with valid addresses are used for L2 check.
In addition, same restriction is applied when doing IPv6 L2 check.

/cc @filanov

### DIFF
--- a/src/connectivity_check/util.go
+++ b/src/connectivity_check/util.go
@@ -1,6 +1,8 @@
 package connectivity_check
 
 import (
+	"fmt"
+	"net"
 	"os/exec"
 	"regexp"
 	"strings"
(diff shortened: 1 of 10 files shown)


src/connectivity_check/util.go
```diff
@@ -50,11 +72,28 @@ func getOutgoingNics(dryRunConfig *config.DryRunConfig, d util.IDependencies) []
...
+				continue
+			}
+			if isIpv4 {
+				outgoingNic.HasIpv4Addresses = true
+			} else {
+				outgoingNic.HasIpv6Addresses = true
+			}
+		}
+		if !outgoingNic.HasIpv4Addresses && !outgoingNic.HasIpv6Addresses {
+			log.Infof("Skipping NIC %s (MAC %s) because of no valid addresses", intf.Name(), intf.HardwareAddr().String())
```
eranco74, line 93: Please add a test case for this scenario

ori-amizur: OK

PR #506, opened 2023-03-01
vendor agent-installer-utils to v0.0.0-20230228190135-5b30a39dadf6

andfasano:
Bumping up the agent-installer-utils version

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/kelseyhightower/envconfig v1.4.0
 	github.com/onsi/ginkgo v1.16.5
 	github.com/onsi/gomega v1.27.1
-	github.com/openshift/agent-installer-utils v0.0.0-20230227172155-0e307b16d36b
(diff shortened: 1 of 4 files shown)


gamli75 (approved):

osherdp (approved):

andfasano:
@osherdp would be possible to override the edge-e2e-metal-assisted? The failure does not seem related to the patch

osherdp:
why not retest?
/retest

andfasano:
Just to save same time, because 

> why not retest? /retest

Just to save some time and reduce the overall roundtrip (in case of some other fix), as the e2e-metal-assisted will take another couple of hours to complete approx. The assisted-installer-agent image will then be required (and thus implicitly this patch) by https://github.com/openshift/installer/pull/6898, which in turn will take other 2/3 hrs to complete and verify that everything works ok (we're integrating the agent-tui)

PR #549, opened 2023-06-01
MGMT-14751: Detect DASD disks (s390x)

avishayt:
Detect DASD disks for s390x, which have 3 flavors - ECKD, ECKD (ESE), and FBA.


src/inventory/disks_test.go
```diff
@@ -918,6 +918,39 @@ var _ = Describe("Disks test", func() {
...
+		dasdbDisk := createDisk("dasdb", 2, "2", "0x2")
+		dependencies.On("EvalSymlinks", "/sys/block/dasdb").Return("/sys/devices/css0/0.0.000f/0.0.5237/block/dasdb", nil)
+		dependencies.On("ReadFile", "/sys/devices/css0/0.0.000f/0.0.5237/discipline").Return([]byte("ECKD\n"), nil)
+		dependencies.On("ReadFile", "/sys/devices/css0/0.0.000f/0.0.5237/ese").Return([]byte("1\n"), nil)
+		// dasdc is FBA
+		dasdcDisk := createDisk("dasdc", 3, "3", "0x3")
+		dependencies.On("EvalSymlinks", "/sys/block/dasdc").Return("/sys/devices/css0/0.0.000f/0.0.5238/block/dasdc", nil)
+		dependencies.On("ReadFile", "/sys/devices/css0/0.0.000f/0.0.5238/discipline").Return([]byte("FBA\n"), nil)
+
+		mockGetWWNCallForSuccess(dependencies, make(map[string]string))
```
gamli75, line 937: can you please add some negative tests to verify that we will get models.DriveTypeUnknown?

AmadeusPodvratnik:
Code is looking good so far and should work for FCP and DASD drives. We need a note for the ESE-feature  documented that for ESE devices they need to be formated in full mode because the check will fail. 
I will discuss this with Holger Wolf if there is a requirement to format disks with mode=quick. Nevertheless this fix will unblock the zVM installation.

src/inventory/disks_test.go
avishayt: Good call, done

gamli75 (approved):

PR #553, opened 2023-06-05
MGMT-13009: Auto rename host with VLAN

jhernand:
Currently the host auto rename feature doesn't work for hosts that only have IP addresses in VLAN network interface cards. This patch fixes that limitation.

Related: https://issues.redhat.com/browse/MGMT-13009

### DIFF
--- a/src/inventory/inventory.go
+++ b/src/inventory/inventory.go
@@ -119,6 +119,7 @@ func findUsableNIC(inventory *models.Inventory) (result *models.Interface, err e
 	// Find the first NIC that has a MAC address an a global IP address:
 	for _, nic := range nics {
 		isPhysical := nic.Type == "physical"
+		isVLAN := nic.Type == "vlan"
 		hasMAC := nic.MacAddress != ""
 		hasV4 := false
 		for _, ip := range nic.IPV4Addresses {
@@ -140,7 +141,7 @@ func findUsableNIC(inventory *models.Inventory) (result *models.Interface, err e
 				break
 			}
 		}
-		if isPhysical && hasMAC && (hasV4 || hasV6) {
+		if (isPhysical || isVLAN) && hasMAC && (hasV4 || hasV6) {
 			result = nic
 			return
 		}
--- a/src/inventory/inventory_test.go
+++ b/src/inventory/inventory_test.go
@@ -198,5 +198,23 @@ var _ = Describe("Hostname processing", func() {
 			},
 			"localhost",
 		),
+		Entry(
+			"Renames with VLAN",
+			"localhost",
+			[]*models.Interface{
+				{
+					Name:       "eth0",
+					Type:       "physical",
+					MacAddress: "71:A0:A4:6F:BE:C8",
+				},
+				{
+					Name:          "eth0.42",
+					Type:          "vlan",
+					MacAddress:    "71:A0:A4:6F:BE:C8",
+					IPV4Addresses: []string{"192.168.0.2/24"},
+				},
+			},
+			"71-a0-a4-6f-be-c8",
+		),
 	)
 })


jhernand:
@filanov @rccrdpccl please review.

rccrdpccl:
of course feel free to unhold whenever @jhernand

ori-amizur:
@jhernand , Did you check that after reboot we always get the same MAC address for VLANs ? What about static networking ?

jhernand:
@ori-amizur I didn't check explicitly if the MAC address is the same after reboot, but the MAC address for a VLAN is the same of the underlying network interface, so I assume it will preserved.

Not sure what is your concern about static networking, can you elaborate?

ori-amizur:
> but the MAC address for a VLAN is the same of the underlying network interface, so I assume it will preserved.

Was it checked? I am not sure this is the case

> Not sure what is your concern about static networking, can you elaborate?

If we use nmstate

ori-amizur:
I saw that by default the same mac address is assigned, but it can be changed

jhernand:
@ori-amizur to try to answer your questions I created a RHEL 9 virtual machine with two network interfaces, the first configured with an static IP address and the second unconfigured:

```
# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: enp1s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 02:00:00:00:00:0a brd ff:ff:ff:ff:ff:ff
    inet 192.168.122.10/24 brd 192.168.122.255 scope global noprefixroute enp1s0
       valid_lft forever preferred_lft forever
3: enp2s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 02:00:00:00:01:0a brd ff:ff:ff:ff:ff:ff
```

The first is to be able to SSH to the machine, and the second is to test this. As you can see the second network interface has MAC address `02:00:00:00:01:0a`. I then added the VLAN virtual interface, same that `nmstate` would do:

```
# nmcli conn add type vlan con-name enp2s0.42 dev enp2s0 id 42
Connection 'enp2s0.42' (75ed967f-c914-4c02-9369-80ab3c25e60c) successfully added.
```

The resulting VLAN virtual interface has the same MAC than the underlying one:

```
1: enp2s0.42@enp2s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default qlen 1000
    link/ether 02:00:00:00:01:0a brd ff:ff:ff:ff:ff:ff
```

I then rebooted the machine and verified that after the reboot the VLAN interface still has the same MAC address.

This MAC address can indeed be changed, for example:

```
# nmcli conn modify enp2s0.42 802-3-ethernet.cloned-mac-address 02:00:00:00:01:0b
```

Or with nmstate:

```yaml
interfaces:
  - name: enp2s0
    type: ethernet
    state: down
    ipv4:
      enabled: false
    ipv6:
      enabled: false
  - name: enp2s0.42
    type: vlan
    state: up
    mac-address: "02:00:00:00:01:0e"
    ipv4:
      enabled: true
      address:
        - ip: 192.168.100.1
          prefix-length: 24
    ipv6:
      enabled: false
    vlan:
      base-iface: enp2s0
      id: 42
```

I can't imagine a situation where this would be required or useful, but if there is then I think it is reasonable to assume that the user will configure correctly that MAC address. Worst that can happen is that the host would get a non unique host name, and then the assisted service would detect that and refuse to install.

All in all I think it isn't necessary to do any special handling of this.

ori-amizur:
OK. We can go forward with this PR.

jhernand:
Thanks @ori-amizur!

PR #558, opened 2023-06-14
OCPBUGS-14848: Filter out hidden devices

avishayt:
Devices that have the "hidden" attribute should not be returned.  They actually have no corresponding block device file under /dev.  This property is set for the underlying components of multipath devices (usually NVMe).


Gleway:
can you cherry pick to 4.13.x train? OCPBUGS-14848 we expect this fix to be available for 4.13.x

avishayt:
Created #559 for 4.13

PR #559, opened 2023-06-14
[release-4.13] OCPBUGS-14992: Filter out hidden devices

avishayt:
Devices that have the "hidden" attribute should not be returned.  They actually have no corresponding block device file under /dev.  This property is set for the underlying components of multipath devices (usually NVMe).


zaneb:
/retitle [release-4.13] OCPBUGS-14992: Filter out hidden devices

PR #560, opened 2023-06-15
MGMT-13111: Retry get next steps for ever

jhernand:
Currently when the request to get next steps fails we retry it with exponential back-off and with a maximum elapsed time of 15 minutes. After that the next step runner exits, and then the agent will register again. A side effect of that is that if the host is explicitly deleted then it will be automatically re-added aftor those 15 minutes, because the agent will receive a 404 error. That isn't desired. To address it this patch changes the back-off configuration so that it doesn't have a maximum elapsed time. That means that we will now retry every 5 minutes, for ever.

Note that the back-off library that we use has a default randomization factor of 0.5, so the actual intervals will not be exactly five minutes, they will be between 2m30s and 7m30s.

Related: https://issues.redhat.com/browse/MGMT-13111

### DIFF
--- a/src/commands/step_processor.go
+++ b/src/commands/step_processor.go
@@ -296,7 +296,12 @@ func ProcessSteps(ctx context.Context, cancel context.CancelFunc, agentConfig *c
 		log.Errorf("Step processing failed, will try again in %s: %v", delay, err)
 	}
 	for !exit {
+		// Note that this back-off configuration, with the max elapsed time set to zero,
+		// means that there is no limit; after reaching the five minute interval we will
+		// keep retrying every five minutes for ever.
 		backOff := backoff.NewExponentialBackOff()
+		backOff.MaxInterval = 5 * time.Minute
+		backOff.MaxElapsedTime = 0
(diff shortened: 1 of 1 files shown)


filanov:
Without the ability to disable this form the service i don't think that we should retry forever, otherwise image a bug or just Agents that retry forever even if cluster doesn't exist

carbonin:
The original request around this feature IIRC was that we don't want the agent to retry forever in the SaaS if the cluster has been removed. I don't think that requirement has changed.

To also add a retry forever for on-prem use cases I think we should add some configuration that the service can pass to the agent either through env vars when the service writes out the agent ignition or through some configuration passed to the agent when it registers or gets steps.

This will allow us to continue to stop agents from polling the service forever in the SaaS while also solving the issues on-prem (agents reregistering and DR).

oourfali:
@jhernand what's next on this pr?

jhernand:
> @jhernand what's next on this pr?

I need to implement the capability to configure this from the service as requested by @filanov and @carbonin.

PR #563, opened 2023-06-21
AGENT-594: remove agent based installer code

andfasano:
This patch removes all the code (and vendored dependencies) related to the agent-based installer (to ship the agent-tui).

Requires https://github.com/openshift/release/pull/40513 (where the CI part was cleaned up)


andfasano:
/assign @osherdp 

https://github.com/openshift/release/pull/40513 landed, so we could proceed with this PR now

osherdp (approved):

PR #578, opened 2023-07-20
DO NOT MERGE: Attempt to debug next_step_runner crash on RHEL 9

zaneb:


zaneb:
Bingo:

```
next_step_runner: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by next_step_runner)
next_step_runner: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by next_step_runner)
```

zaneb:
It's running next_step_runner in the container `registry.build05.ci.openshift.org/ci-op-jwgyd7ry/stable@sha256:1435334e1034219e0bf364f68dc20ae9777bb15fcad4374247821c14a8c97884`, which is a RHEL 8 container. Since next_step_runner is a RHEL 9 binary, that explains the error.

The reason is that we're somehow building with the RHEL9 build container but a RHEL8 base container. The [log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_assisted-installer-agent/578/pull-ci-openshift-assisted-installer-agent-master-e2e-agent-compact-ipv4/1681855463798345728/artifacts/build-logs/agent-installer-node-agent.log) says:

```
2023-07-20T02:38:05.128828022Z Replaced Dockerfile FROM image registry.ci.openshift.org/ocp/4.14:base-rhel9
```

and although it doesn't say what it's replacing it with (presumably `4.14:base`), it's clearly RHEL 8. Presumably we have to change this in the release repo before what we say here will take effect? I thought a bot would make that change, but thus far it has not. Maybe it's pulling the change from the repo here and not directly from the ART config as I had assumed?

zaneb:
https://github.com/openshift/release/pull/41477 should be enough to prevent this error I think.
/close

PR #604, opened 2023-09-19
MGMT-15809: change base image to stream9

osherdp:
Now that D/S building system fully supports RHEL-9 base images (and allows easy installation of nmstate and other libs that were previously unsupported) we can move U/S to stream9 and D/S to ubi9-minimal.

It doesn't seem possible to easily use ubi9-minimal U/S for the assisted-service, because installation of nmstate packages either requires entitlements running on a RHEL node, or alternatively an access to RedHat's internal network (which can be a major problem for U/S development). For a better consistency we're sticking with the same for the agent, installer, etc.

### DIFF
--- a/subsystem/lease_test.go
+++ b/subsystem/lease_test.go
@@ -238,7 +238,7 @@ func startTcpDump(output chan string, ifaceName string, timeoutSecs, count int) 
 	if count > 0 {
 		countStr = fmt.Sprintf("-c %d", count)
 	}
-	dump := fmt.Sprintf("timeout %d tcpdump -l -i %s %s -v 'udp dst port 67' | awk '/DHCP-Message Option 53/{print $6}'", timeoutSecs, ifaceName, countStr)
+	dump := fmt.Sprintf("timeout %d tcpdump -l -i %s %s -v 'udp dst port 67' | awk '/DHCP-Message \\(53\\)/{print $5}'", timeoutSecs, ifaceName, countStr)
 
 	s, e, errorCode := agentUtils.Execute("docker", []string{"exec", "agent", "bash", "-c", dump}...)
 	fmt.Println(s, e, errorCode)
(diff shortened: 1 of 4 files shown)


osherdp:
/hold
syncing with D/S changes

osherdp:
``mce-images`` will get fixed with https://github.com/openshift/release/pull/43781

subsystem/lease_test.go
```diff
@@ -238,7 +238,7 @@ func startTcpDump(output chan string, ifaceName string, timeoutSecs, count int)
 	if count > 0 {
 		countStr = fmt.Sprintf("-c %d", count)
 	}
-	dump := fmt.Sprintf("timeout %d tcpdump -l -i %s %s -v 'udp dst port 67' | awk '/DHCP-Message Option 53/{print $6}'", timeoutSecs, ifaceName, countStr)
+	dump := fmt.Sprintf("timeout %d tcpdump -l -i %s %s -v 'udp dst port 67' | awk '/DHCP-Message \\(53\\)/{print $5}'", timeoutSecs, ifaceName, countStr)
```
adriengentil, line 241: good catch ! :+1:

adriengentil (approved):

subsystem/lease_test.go
adriengentil: could be nice to `tee` tcpdump results in a file and put it in the artifacts?

osherdp: that's a great idea, but it will require some extra coding (as I don't think we currently support transferring artifacts from the skipper container to the outside host)

adriengentil: yeah, it's what I guessed, we can do it later in a follow up work

osherdp:
```
github.com/containers/image/v5@v5.20.0: stream error: stream ID 4253; INTERNAL_ERROR; received from peer
```
not sure what this is about, maybe a minor connectivity issue
/test edge-subsystem-test

adriengentil (approved):

PR #605, opened 2023-09-19
NO-ISSUE: fix lint target and code

osherdp:
For some reason we applied ``--fix`` to the lint code, which fixes the issues and return a 0 code (so the job always passes).

This removes it and fixes the lint issues in code.


eliorerz (approved):
LGTM

PR #617, opened 2023-10-22
Shrink agent image

omertuc:
[MGMT-16011](https://issues.redhat.com//browse/MGMT-16011)

See commit message for information on what changed

From **606MB** to **315MB** (48% reduction)

# Analysis

## Original additional layers total size: **452MB**
* **245MB** dnf installations layers
* **207MB** agent binaries (**53MB** + **13MB** + **14MB** + **56MB** + **59MB** + **12MB**)

## New additional layers total size: **160MB** 
* **118MB** dnf installations layer
* **42MB** agent binary

## Summary

65% reduction in added size.

## This seems to be rock bottom:

* Half the total size is now just the base image
* It's difficult to make the agent binary any smaller, go binaries are just big
* 50% of the size DNF installations layer is just podman which is not something we can replace
* The rest of the dnf installations layer is full of small tools we rely on

# Downstream

We need to perform similar work also on the downstream Dockerfile, which creates images that weigh much more

### DIFF
--- a/Dockerfile.assisted_installer_agent
+++ b/Dockerfile.assisted_installer_agent
@@ -10,36 +10,48 @@ RUN go mod download
 
 COPY . .
 
-RUN TARGETPLATFORM=$TARGETPLATFORM make build
+RUN TARGETPLATFORM=$TARGETPLATFORM make build-release
 
 FROM quay.io/centos/centos:stream9
 ARG TARGETPLATFORM
-RUN if [[ "$TARGETPLATFORM" == "linux/amd64" || -z "$TARGETPLATFORM" ]] ; then  dnf install -y biosdevname dmidecode ; fi
-RUN if [[ "$TARGETPLATFORM" == "linux/arm64" || -z "$TARGETPLATFORM" ]] ; then  dnf install -y dmidecode ; fi
 
-RUN dnf install --setopt=install_weak_deps=False --setopt=tsdocs=False -y \
-		findutils iputils \
-		podman \
-		# inventory
-		ipmitool file fio \
-		sg3_utils \
-		# free_addresses
-		nmap \
-		# dhcp_lease_allocate
-		dhclient \
-		# logs_sender
-		tar openssh-clients \
-		# ntp_synchronizer
-		chrony \
-		# for the 'nsenter' executable
-		util-linux-core \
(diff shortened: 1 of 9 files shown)


avishayt:
This is awesome, thank you!

You have a few other suggestions - are they not worth doing?
- Strip debug information from the agent binaries
- We don't need the entire /usr/share/misc/magic database, stop using file and use a simpler detection for MBR partition
- Remove X11
- Remove licenses (or maybe compress them if it's legally problematic)
- Remove grub , this is a container, it doesn't boot
- I'm not sure where all the ceph stuff is coming from
- Remove Python - where are we even using Python in the agent?
- Look into the libicudata and libmozjs things, I'm not sure we need them (or what they are)

avishayt:
Also, is there any reason we're using centos as a base?  Why not ubi?  There are some really small ubi images.

omertuc:
> This is awesome, thank you!
> 
> You have a few other suggestions - are they not worth doing?
> 
> * Strip debug information from the agent binaries

I'm doing that, just forgot to mention it in the commit message, it's the `-ldflags "-s -w"` part

> * We don't need the entire /usr/share/misc/magic database, stop using file and use a simpler detection for MBR partition
> * Remove X11
> * Remove grub , this is a container, it doesn't boot
> * Remove Python - where are we even using Python in the agent?

I saw these are part of the base image, nothing we can do. 

> * Remove licenses (or maybe compress them if it's legally problematic)

Also mostly base image. We only add like 100KB of licenses during our steps

> * I'm not sure where all the ceph stuff is coming from
> * Look into the libicudata and libmozjs things, I'm not sure we need them (or what they are)

Doesn't seem like these are part of the upstream image, I'll have to see what's going on in the downstream image that makes those appear

omertuc:
>Also, is there any reason we're using centos as a base? Why not ubi? There are some really small ubi images.

We are using ubi [downstream](https://gitlab.cee.redhat.com/rh-ecosystem-edge/assisted-installer/assisted-installer-projects/blob/e86dda51453a30575861225faf6ae49319b08364/distgit/containers/assisted-installer-agent/Dockerfile.in#L17), currently `registry.redhat.io/ubi9-minimal:9.2-750.1696515534`

avishayt (approved):
Great work - thanks!

omertuc:
Downstream PR https://gitlab.cee.redhat.com/rh-ecosystem-edge/assisted-installer/assisted-installer-projects/-/merge_requests/352

PR #624, opened 2023-11-08
MGMT-15271: Get cmdline and set IBM in vendor

AmadeusPodvratnik:
With the support of zVM in Assisted Installer, the necessary kernel arguments to boot a zVM nodes need to be considered. During the first boot of a zVM node a parm file will be provided to activate the necessary devices or specify a static IP w/o nmstate. This parm file looks like:

rd.neednet=1 console=ttysclp0 coreos.live.rootfs_url=http://172.23.236.156:8080/assisted-installer/rootfs.img ip=10.14.6.3::10.14.6.1:255.255.255.0:master-0.boea3e06.lnxero1.boe:encbdd0:none nameserver=10.14.6.1 ip=[fd00::3]::[fd00::1]:64::encbdd0:none nameserver=[fd00::1] zfcp.allow_lun_scan=0 rd.znet=qeth,0.0.bdd0,0.0.bdd1,0.0.bdd2,layer2=1 rd.dasd=0.0.5235 rd.dasd=0.0.5236 random.trust_cpu=on rd.luks.options=discard ignition.firstboot ignition.platform.id=metal console=tty1 console=ttyS1,115200n8

Nevertheless the cmdline will be taken independently of the architecture. For s390x gwh is returning for getProduct():
product family=unknown name=unknown vendor=unknown sku=unknown version=unknown
and need to be set to the correct values. To do that you need to parse:
cat /proc/cpuinfo |grep vendor
vendor_id : IBM/S390
sudo cat /proc/sysinfo |grep 'VM.*Control Program'
VM00 Control Program: z/VM 7.2.0
For KVM following values are returned:
cat /proc/sysinfo |grep 'VM.*Control Program'
VM00 Control Program: KVM/Linux

List all the issues related to this PR

[x ] New Feature
      Enhancement
      Bug fix
      Tests
      Documentation
      CI/CD 

How was this code tested?

   assisted-test-infra environment
   dev-scripts environment

[x ] Reviewer's test appreciated
Waiting for CI to do a full test run
[x ] Manual (Elaborate on how it was tested)
- Using boot_test.go
No tests needed

PR #625, opened 2023-11-09
WIP Dummy PR to check CI

omertuc:
#624 is failing in CI but it's unclear whether it's a job issue or PR issue,
this dummy PR that doesn't change anything should help confirm that


omertuc:
It passed

PR #627, opened 2023-11-21
OCPBUGS-23537: backport of agent retries

tsorya:
Backporting 5 prs,
https://github.com/openshift/assisted-installer-agent/pull/442
https://github.com/openshift/assisted-installer-agent/pull/462
https://github.com/openshift/assisted-installer-agent/pull/459
https://github.com/openshift/assisted-installer-agent/pull/438
https://github.com/openshift/assisted-installer-agent/pull/583


zaneb:
#583 is not present in 4.13, so it needs to be backported there first to avoid a regression for users going from 4.12->4.13.

PR #631, opened 2023-11-30
NO-ISSUE: fix dependabot update

eifrach:


eifrach:
> @eifrach Was `ghw.MemoryInfo` name was changed to `ghw.MemoryArea` in `ghw` update that was made by dependabot ?

yes, the name was changed and which breaks in the new version

filanov:
To avoid most lint issues i suggest to add a file watcher to your editor, it will automatically fix most of the lint issues for you

filanov (approved):

PR #639, opened 2023-12-12
[release-ocm-2.9] MGMT-16370: fix golangci-lint failing to be pulled

adriengentil:
app-sre made their repo mirroring golangci-lint private for legal reasons, and
now our jobs are failing to install it. Installing it now from the cli.

We could also have pulled golangci-lint from dockerhub, but rate
limitation there is quite aggressive.

### DIFF
--- a/Dockerfile.assisted_installer_agent-build
+++ b/Dockerfile.assisted_installer_agent-build
@@ -4,7 +4,7 @@ ENV GOFLAGS=""
 
 RUN yum install -y docker docker-compose && \
     yum clean all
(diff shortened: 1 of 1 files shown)


adriengentil:
/override ci/prow/images ci/prow/lint ci/prow/mce-images ci/prow/subsystem-test ci/prow/unit-test ci/prow/e2e-ai-operator-ztp
Failing because prow does not take the Dockerfile from the PR, it's an expected behavior

gamli75 (approved):

PR #649, opened 2024-01-22
MULTIARCH-4089: Fix LPAR handling

AmadeusPodvratnik:
On LPAR there is a glitch preventing systemd-detect-virt to detect if there is a virtualized system or not. Even if the command returns correct results if executed via command line directly, it might happen that executing during agent run returns an error (-1). This is not a problem because for s390x architecture zVM and LPAR are not virtualized but only KVM.

With this fix LPAR will be set as Product to make LPAR more visible and add new test case.

List all the issues related to this PR

[ ] New Feature
Enhancement
[x ] Bug fix
Tests
Documentation
CI/CD

How was this code tested?

assited-installer-agent unit-tests

[x ] Reviewer's test appreciated
Waiting for CI to do a full test run
[x ] Manual (Elaborate on how it was tested)

    Unit testing using 
    skipper make unit-test TEST=./src/inventory
    JUnit report was created: /root/code/Assisted_Installer/latest/assisted-installer-agent/src/inventory/junit_unit_test.xml

Ran 103 of 103 Specs in 0.180 seconds
SUCCESS! -- 103 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestInventory (0.20s)

### DIFF
--- a/src/inventory/system_vendor.go
+++ b/src/inventory/system_vendor.go
@@ -16,6 +16,7 @@ const (
 	CTRL_PRG      = "Control Program"
 	KVM_VIRT      = "KVM/Linux"
 	VENDOR_IBM_ID = "IBM/S390"
+	LPAR          = "LPAR"
 )
 
 // For oVirt VMs the correct platform can be detected only by the Family value,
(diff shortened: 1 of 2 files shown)


src/inventory/system_vendor.go
```diff
@@ -51,7 +52,10 @@ func GetVendorFors390x(dependencies util.IDependencies, ret *models.SystemVendor
 		}
 	}
 
-	// get virtualization type
+	// get virtualization type but set Product to LPAR (will be overwritten with zVM or KVM if sysinfo contains virt entry)
+	ret.ProductName = LPAR
```
ori-amizur, line 56: If the platform is s390x we set the ProductName unconditionally to LPAR? Is this what you meant?

AmadeusPodvratnik: Yes ... this can be done ... if nothing is in sysinfo, it's LPAR otherwise I will have KVM or zVM/7.2.0 (or a new version of zVM therefor I'm only checking for KVM/Linux because of missing version info).
It will be expected that LPAR will be the default because of the reduction of virtualization and maintenance and can be called, to some extend, as bare metal on e.g. LinuxOne.

PR #663, opened 2024-02-20
ACM-9505: CVE-2023-49568, CVE-2023-49569 - Remove dependency on go-git

paul-maidment:
Assisted-service has a depedency on the vulnerable go-git library This PR makes the minimal changes required to use the latest assisted-service in the ocm-2.8 branch


gamli75:
@CrystalChun please review

src/commands/actions/domain_resolution_cmd_test.go
```diff
@@ -22,7 +22,7 @@ var _ = Describe("domain resolution", func() {
 	})
 
 	It("domain resolution - bad domain name", func() {
-		param = "{\"domains\":[{\"domain_name\":\"aaaaaa\"}]}"
```
adriengentil, line 25: why this line has changed?

paul-maidment: The way domains are validated is different, whatever the new validation code is doing, doesn't find "aaaaaa" to be invalid (because it's not actually, if "localhost" can be a valid domain name then so can "aaaaaa")

I had to make the domain name actually invalid in order for the test to pass, I chose to add a "$" at the start for this purpose.

And yes, this is due to new service behaviour.

adriengentil (approved):

CrystalChun (approved):
/lgtm

adriengentil:
> I wouldn't override the `lint` job, it will leave the build broken for this branch
> 
> @paul-maidment I think in `release-ocm-2.8` branch the test configuration is different than the newer branches in the way that it doesn't just build `Dockerfile.assisted_installer_agent-build` and runs on top of it but has a different base image: https://github.com/openshift/release/blob/86aa93ca83501e32509fbcc773ac8aadbe95a578/ci-operator/config/openshift/assisted-installer-agent/openshift-assisted-installer-agent-release-ocm-2.8.yaml#L58
> 
> I think only afterwards we organized it a little better

the lint binary is copied from the build root (through test-bin) which is built from `Dockerfile.assisted_installer_agent-build`. This PR updates `Dockerfile.assisted_installer_agent-build`, so I think the lint job will use the new lint version once this PR is merged.

adriengentil:
@osherdp can you review my comment and if that's correct, override the lint job ? Thanks

osherdp:
> @osherdp can you review my comment and if that's correct, override the lint job ? Thanks

ack, thanks!
please assure post-merge that it is indeed passing
/override ci/prow/lint

osherdp (approved):

paul-maidment:
Performing a dummy commit to make sure all is well

https://github.com/openshift/assisted-installer-agent/pull/669

PR #679, opened 2024-04-03
OCPBUGS-31423: Try to pull installer image before installing

CrystalChun:
https://issues.redhat.com/browse/OCPBUGS-31423
Split the install command into two parts: pulling the installer image then running the install command. Pulling separately and ahead of time prevents possible flakes when pulling the image. Mostly relevant for day 2 hosts since we don't run the container image availability command before installing.

/cc @tsorya

### DIFF
--- a/src/commands/actions/install_cmd.go
+++ b/src/commands/actions/install_cmd.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/alessio/shellescape"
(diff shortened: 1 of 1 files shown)


src/commands/actions/install_cmd.go
```diff
@@ -276,3 +288,29 @@ func (a *install) Command() string {
...
+	var err error
+	for attempts := 0; attempts < retry; attempts++ {
+		if err = pullImage(pullTimeoutSeconds, image); err == nil {
+			break
+		}
+	}
+	return err
+}
+
+func pullImage(pullTimeoutSeconds int64, image string) error {
```
tsorya, line 302: Lets validate if image exists beforehand, another way we will pull it all the time

CrystalChun: Good idea! Added it!

CrystalChun:
Oh @tsorya may I get your approval? 😄

PR #733, opened 2024-06-27
MGMT-17867: fix multipath inventory to include wwn

linoyaslan:
Currently, if the cluster has multipath configured, the multipath inventory does not include the WWN. This occurs because the `ghw` package returns "unknown" if it fails to detect the WWN, and we then set it to empty. With my changes, if `ghw` returns "unknown," the code retrieves the WWN differently and returns it instead of an empty string.

### DIFF
--- a/src/inventory/disks.go
+++ b/src/inventory/disks.go
@@ -19,6 +19,8 @@ import (
 
 const (
 	applianceAgentPrefix = "agent"
+	byIdLocation         = "/dev/disk/by-id"
+	wwnPrefix            = "wwn-"
 )
 
 type disks struct {
@@ -30,9 +32,17 @@ func newDisks(subprocessConfig *config.SubprocessConfig, dependencies util.IDepe
 	return &disks{dependencies: dependencies, subprocessConfig: subprocessConfig}
 }
 
+func getPureWWN(wwn string, byId string) string {
+	if wwn == ghwutil.UNKNOWN {
+		prefix := byIdLocation + "/" + wwnPrefix
+		withoutPrefix := strings.TrimPrefix(byId, prefix)
+		return withoutPrefix
+	}
+	return wwn
+}
+
 func (d *disks) getDisksWWNs() map[string]string {
-	const ByIdLocation = "/dev/disk/by-id"
-	filesInfo, err := d.dependencies.ReadDir(ByIdLocation)
+	filesInfo, err := d.dependencies.ReadDir(byIdLocation)
 
 	if err != nil {
 		logrus.Warnf("Cannot get disk/by-id information: %s", err)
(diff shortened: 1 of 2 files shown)


src/inventory/disks.go
```diff
@@ -145,9 +146,12 @@ func (d *disks) hasUUID(path string) bool {
...
 }
 
-func unknownToEmpty(value string) string {
+func unknownToEmpty(value string, defaultVal string) string {
 	if value == ghwutil.UNKNOWN {
-		return ""
+		if strings.HasPrefix(defaultVal, "wwn-") {
+			withoutWWNPrefix := strings.TrimPrefix(defaultVal, "wwn-")
+			return withoutWWNPrefix
+		}
```
avishayt, line 154: This logic doesn't really belong in "unknownToEmpty"

src/inventory/disks.go
```diff
@@ -145,9 +146,12 @@ func (d *disks) hasUUID(path string) bool {
 	return true
 }
 
-func unknownToEmpty(value string) string {
+func unknownToEmpty(value string, defaultVal string) string {
```
avishayt, line 149: It's no longer changing to "Empty"

src/inventory/disks.go
```diff
@@ -68,7 +69,7 @@ func (d *disks) getDisksWWNs() map[string]string {
 			return "", ""
 		}
 
-		return diskPath, diskId
+		return diskPath, wwn
```
avishayt, line 72: I think it would be cleaner to leave this, and add a method to remove "/dev/disk/by-id" and "wwn-"

avishayt (changes requested):

src/inventory/disks.go
```diff
@@ -30,6 +30,15 @@ func newDisks(subprocessConfig *config.SubprocessConfig, dependencies util.IDepe
 	return &disks{dependencies: dependencies, subprocessConfig: subprocessConfig}
 }
 
+func getPureWWN(wwn string, byId string) string {
```
danmanor, line 33: Why is this parameter called `byId` ?

avishayt:
Please use consts and add a unit test

src/inventory/disks.go
linoyaslan: @danmanor Because the argument being sent is the by-id param of the disk
```
ByID:                    diskPath2diskWWN[path],
Wwn:                     getPureWWN(disk.WWN, diskPath2diskWWN[path]),
```

avishayt (approved):

src/inventory/disks.go
```diff
@@ -30,9 +32,17 @@ func newDisks(subprocessConfig *config.SubprocessConfig, dependencies util.IDepe
...
+	if wwn == ghwutil.UNKNOWN {
+		prefix := byIdLocation + "/" + wwnPrefix
+		withoutPrefix := strings.TrimPrefix(byId, prefix)
```
zaneb, line 38: This will return a wrong value if the disk does not start with `wwn-` (`nvme-eui` is also a valid possibility from line 55).

zaneb (commented):
I believe this code is effectively dead since #936 updated to a version of ghw that supports returning the WWN for multipath disks. We should probably consider reverting to reduce confusion in future.

PR #738, opened 2024-07-02
[release-ocm-2.9] NO-ISSUE: Replace base golang image in the build Dockerfile

danmanor:
Replace base golang image in the build Dockerfile as it is based on centos7 and it is EOL, which results in yum can't reach its repositories


paul-maidment:
I think this may no longer be needed? Can you verify @danmanor ?

PR #750, opened 2024-07-07
MGMT-18332: Use centos stream 9 as the base image for test container

danmanor:
- Use centos stream 9 as the base image for test container
- Replace deprecated scopelint linter


danmanor:
/override ci/prow/e2e-agent-compact-ipv4 ci/prow/edge-e2e-ai-operator-ztp ci/prow/edge-e2e-metal-assisted

PR #759, opened 2024-07-25
[release-ocm-2.8] MGMT-18313: Replace golang base image as it is based on Centos Linux 7

danmanor:
This PR moves centos base image from stream8 to stream 9 as well as stream8 is EOL


paul-maidment (approved):
/lgtm

PR #761, opened 2024-07-31
MGMT-18837: Fix broken build due to base image change.

paul-maidment:
The image registry.access.redhat.com/ubi9/go-toolset:1.20 does not default to the root user as registry.ci.openshift.org/openshift/release:golang-1.20 used to This results in a permissions issue when attempting to run `go mod download`

By setting the user directive prior to this operation, this PR resolves that issue

### DIFF
--- a/Dockerfile.assisted_installer_agent
+++ b/Dockerfile.assisted_installer_agent
@@ -3,12 +3,10 @@ ARG TARGETPLATFORM
 ENV GO111MODULE=on
 ENV GOFLAGS=""
 
-WORKDIR /go/src/github.com/openshift/assisted-installer-agent
-
-COPY go.mod .
+COPY --chown=1001 go.mod .
 RUN go mod download
 
-COPY . .
+COPY --chown=1001 . .
 
 RUN TARGETPLATFORM=$TARGETPLATFORM make build-release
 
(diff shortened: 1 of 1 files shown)


rccrdpccl:
/hold

where is this failing?

rccrdpccl:
For the records we discuss this and it happens also outside skipper, due to misalignment of permissions.

danmanor:
why not just use `root` in the base image as done initially ?

Dockerfile.assisted_installer_agent
```diff
@@ -3,12 +3,10 @@ ARG TARGETPLATFORM
 ENV GO111MODULE=on
 ENV GOFLAGS=""
 
-WORKDIR /go/src/github.com/openshift/assisted-installer-agent
-
-COPY go.mod .
+COPY --chown=1001 go.mod .
```
eifrach, line 6: let's keep it simple and same as other builds
just set the user for the build stage 

same as the controller
https://github.com/openshift/assisted-installer/blob/master/Dockerfile.assisted-installer#L7

eifrach (changes requested):

rccrdpccl:
> why not just use root in the base image as done initially ? @rccrdpccl @paul-maidment
The base image is meant to be used differently, using the home folder instead of pre-go mod style of folder for the module being compiled.
I think it would be easier to understand to use it that way and there is no real need to use root user.

That said, in this specific case it's not a real disadvantage using root.

Dockerfile.assisted_installer_agent
rccrdpccl: I understand the root user approach is used everywhere else, but as [stated in the comments](https://github.com/openshift/assisted-installer/blob/master/Dockerfile.assisted-installer#L6) it's a workaround, while this would be the expected way of building golang projects with this base image.

rccrdpccl:
/hold
Let's make sure we agree on the solution before unholding
I'm ok with both but I think the current one (using the home directory) is better.

/retest

flake

PR #767, opened 2024-08-21
MGMT-18521: fix NTP sources are not valid

eifrach:
Fix issue where the IP of the NTP source has a domain and cannot be matched to the IP set

```console
> nslookup security.cloudflare-dns.com
...
...
Name:	security.cloudflare-dns.com
Address: 1.1.1.2

```

- move all subsystem test IP's to internal local IP's


danmanor:
your change is only affecting tests, overriding agent job

danmanor:
@eifrach please cherry-pick ocm-2.11 - ocm-2.8

PR #772, opened 2024-08-22
MGMT-18475: disabling CGO

eifrach:


Makefile
```diff
@@ -10,9 +10,7 @@ BIN = $(ROOT_DIR)/build
 # Multiarch support.  Transform argument passed from docker buidx tool to go build arguments to support cross compiling
 GO_BUILD_ARCHITECTURE_VARS := $(if ${TARGETPLATFORM},$(shell echo ${TARGETPLATFORM} | awk -F / '{printf("GOOS=%s GOARCH=%s", $$1, $$2)}'),)
 CGO_FLAG := $(or $(CGO_FLAG),0)
```
danmanor, line 12: no need for this right

Makefile
```diff
@@ -10,9 +10,7 @@ BIN = $(ROOT_DIR)/build
...
-ifeq ($(TARGETPLATFORM),linux/amd64)
-	CGO_FLAG=1
-endif
+
 GO_BUILD_VARS := CGO_ENABLED=$(CGO_FLAG) $(GO_BUILD_ARCHITECTURE_VARS)
```
danmanor, line 14: No need for `CGO_ENABLED=$(CGO_FLAG)` also

eifrach: the default is `CGO_ENABLED=1` so you must disable it

Makefile
eifrach: the default is CGO_ENABLED=1 so you must disable it

Makefile
danmanor: so can we just `CGO_ENABLED=0` ?

PR #776, opened 2024-09-03
MGMT-17103: Fixing boot issue for IBMZ hosts in reclaim

veera-damisetti:
Fixing boot issue for Z hosts in reclaim
- Removed single quotes for coreos.live.rootfs_url in bootLoaderConfigTemplate
- Added zipl incase of s390x for handling boot issue in reclaiming agents (HCP)

### DIFF
--- a/src/commands/actions/download_boot_artifacts_cmd.go
+++ b/src/commands/actions/download_boot_artifacts_cmd.go
@@ -12,6 +12,8 @@ import (
 	"syscall"
 
 	"github.com/openshift/assisted-installer-agent/src/config"
+	inv "github.com/openshift/assisted-installer-agent/src/inventory"
+	"github.com/openshift/assisted-installer-agent/src/util"
 	"github.com/openshift/assisted-service/models"
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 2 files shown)


carbonin:
Let's wait for the new CI test to be added in https://github.com/openshift/release/pull/56410 so we can avoid regressions.

src/commands/actions/download_boot_artifacts_cmd.go
```diff
@@ -141,7 +148,13 @@ func createBootLoaderConfig(rootfsUrl, artifactsPath, bootLoaderPath string) err
 	kernelPath := path.Join(artifactsPath, kernelFile)
 	initrdPath := path.Join(artifactsPath, initrdFile)
 	bootLoaderConfigFile := path.Join(bootLoaderPath, bootLoaderConfigFileName)
-	bootLoaderConfig := fmt.Sprintf(bootLoaderConfigTemplate, rootfsUrl, kernelPath, initrdPath)
+	var bootLoaderConfig string
+	dependencies := &util.Dependencies{}
+	cpuInfo := inv.GetCPU(dependencies)
```
carbonin, line 153: Is there an upside to doing this rather than `runtime.GOARCH`?

veera-damisetti: yeah, as there is an existing function GetCPU() to get the architecture information , we just wanted to reuse it rather than getting from a new module runtime

carbonin: Runtime is part of the [standard library](https://pkg.go.dev/runtime@go1.23.1#pkg-constants)

veera-damisetti: Sure @carbonin  
I will do the changes to use runtime.GOARCH for fetching architecture details

carbonin (approved):

PR #778, opened 2024-09-09
MGMT-18839: Retry downloading boot artifacts

CrystalChun:
Adds a retry when downloading the boot artifacts from assisted service in case there are connection issues or a server disruption.

/cc @carbonin

### DIFF
--- a/src/commands/actions/download_boot_artifacts_cmd.go
+++ b/src/commands/actions/download_boot_artifacts_cmd.go
@@ -11,6 +11,7 @@ import (
 	"path"
 	"runtime"
 	"syscall"
+	"time"
 
 	"github.com/openshift/assisted-installer-agent/src/config"
 	"github.com/openshift/assisted-service/models"
@@ -45,6 +46,8 @@ func (a *downloadBootArtifacts) Args() []string {
 }
 
 const (
+	retryDownloadAmount                  = 5
+	defaultDownloadRetryDelay            = 1 * time.Minute
 	artifactsFolder               string = "/boot/discovery"
 	kernelFile                    string = "vmlinuz"
 	initrdFile                    string = "initrd"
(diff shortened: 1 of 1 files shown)


src/commands/actions/download_boot_artifacts_cmd.go
```diff
@@ -116,20 +117,28 @@ func createHTTPClient(caCertPath string) (*http.Client, error) {
...
 }
 
-func download(httpClient *http.Client, filePath, url string) error {
-	res, err := httpClient.Get(url)
-	if err != nil {
-		return fmt.Errorf("failed getting %s: %w", url, err)
+func download(httpClient *http.Client, filePath, url string, retry int) error {
+	var downloadErr error
+	var res *http.Response
+	for attempts := 0; attempts < retry; attempts++ {
```
carbonin, line 123: I think maybe add a delay here so we don't do the 5 requests all right after each other.

CrystalChun: Good idea, thanks! Modified

src/commands/actions/download_boot_artifacts_cmd.go
```diff
@@ -116,20 +119,29 @@ func createHTTPClient(caCertPath string) (*http.Client, error) {
...
+		res, downloadErr = httpClient.Get(url)
+		if downloadErr == nil && (res.StatusCode >= 200 && res.StatusCode < 300) {
+			break
+		}
+		downloadErr = fmt.Errorf("failed downloading boot artifact from %s, status code received: %d, attempt %d/%d, download error: %w",
+			url, res.StatusCode, attempts, retry, downloadErr)
+		log.Warn(downloadErr.Error())
+		time.Sleep(defaultDownloadTimeout)
```
paul-maidment, line 133: Should we call this `defaultDownloadRetryDelay` ?

carbonin: Indeed, it's not a timeout.

CrystalChun: Good idea, thanks! Changed it

carbonin (approved):

PR #783, opened 2024-09-18
OCPBUGS-42157: Switch to github.com/docker/distribution/reference to Mitigate CVE-2024-3727

paul-maidment:
The library github.com/containers/image/v5 has a vulnerability that has as of yet been unresolved.

Thankfully, it is possible to change the part of the library that we use

We can change github.com/containers/image/v5/docker/reference for github.com/docker/distribution/reference

In the case of assisted-installer-agent, we achieve this by changing the dependency to the latest assisted-installer


paul-maidment:
@zaneb @andfasano can you help with the required label? thanks

zaneb:
/lgtm
Is this blocking an urgent backport to 4.16? If so I can add the staff-eng-approved label.

danmanor (approved):

PR #784, opened 2024-09-22
MGMT-18466: improve depndabot grouping

eifrach:
- Improve dependabot groups.
Allowing security update in a separate group give us smaller change ( minor version change ) which will be easier to auto merge.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,13 +2,12 @@
 # package ecosystems to update and where the package manifests are located.
 # Please see the documentation for all configuration options:
 # https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
-
 version: 2
 updates:
   - package-ecosystem: "gomod"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -2,13 +2,12 @@
...
 # Please see the documentation for all configuration options:
 # https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
-
 version: 2
 updates:
   - package-ecosystem: "gomod"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "daily"
```
danmanor, line 10: Are you sure about it ? seems like weekly should be enough no ? I am worried about duplicated PRs

eifrach: it not - due to grouping

if there's a PR open it will rebase it with the changes

danmanor (approved):

PR #787, opened 2024-09-23
MGMT-18923: Refactor getGPUs

mlorenzofr:
These changes fix an issue in `inventory` when the GPUs are discovered on the system. Due to how ghw library works, only devices with a DRM interface are considered GPUs, which excludes devices like nvidia A100 or Habana Gaudi due to them not create that interface. `inventory` will now check all PCI devices looking for GPU devices. To provide more flexibility in the future, a new configuration option has been provided to `inventory`. The configuration can be set via configuration file or via **INVENTORY_GPU_CONFIG** environment variable (documentation and examples provided in the README)

The `inventory` binary has been built (`make build`) and tested on my own laptop, a beaker machine, a VM and also a physical machine with Nvidia A100 hardware. In all cases, the GPUs were detected correctly, and filtering via environment variable and configuration file was also successful.

# List all the issues related to this PR
- [ ] New Feature
- [x] Enhancement
- [x] Bug fix
- [x] Tests
- [x] Documentation
- [ ] CI/CD

# How was this code tested?
- [ ] Reviewer's test appreciated
- [ ] Waiting for CI to do a full test run
- [x] Manual (Elaborate on how it was tested)
- [ ] No tests needed

# Checklist
- [x] Title and description added to both, commit and PR.
- [ ] Relevant issues have been associated (see CONTRIBUTING guide)
- [ ] This change does not require a documentation update (docstring, docs, README, etc)
- [x] Does this change include unit-tests (note that code changes require unit-tests)

# Reviewers Checklist
* Are the title and description (in both PR and commit) meaningful and clear?
* Is there a bug required (and linked) for this change?
* Please take a careful look if something is unclear in the code or in its workflow


src/inventory/gpu_test.go
```diff
@@ -45,73 +84,90 @@ var (
...
-		gpus := GetGPUs(dependencies)
+		var yamlData = "---\nvendors:\n  - '1200 1da3'"
+
+		tmpFile, err := os.CreateTemp("", "gpus*.yaml")
+		if err != nil {
+		}
+		defer os.Remove(tmpFile.Name())
+
+		_, err = tmpFile.WriteString(yamlData)
+		if err != nil {
```
rccrdpccl, line 144: Please remove empty condition

rccrdpccl:
/lgtm

/hold

Looks good to me, @CrystalChun would u mind taking a quick look in case I am missing something?

src/inventory/gpu.go
```diff
@@ -1,37 +1,144 @@
...
+}
+
+// isSupportedClass validates if the PCI Class is supported
+func isSupportedClass(class string) bool {
+	return slices.Contains(supportedConfig.Classes, class)
+}
+
+// isSupportedVendor validates if the PCI Vendor ID is in the list of supported GPUs
+func isSupportedVendor(vendorID string) bool {
+	return slices.Contains(supportedConfig.Vendors, vendorID)
```
ori-amizur, line 94: Since some of these may contain whitespace as separator, maybe we can ignore the whitespaces, so different number of sequential whitespaces wouldn't fail the comparison?

ori-amizur (commented):
Does the change allow to detect other PCI devices (non GPU)?

src/inventory/gpu.go
```diff
@@ -1,37 +1,144 @@
...
+	"gopkg.in/yaml.v2"
 )
 
-func GetGPUs(dependencies util.IDependencies) []*models.Gpu {
+// PCI classes, combination of base class (2) + subclass (2)
+const (
+	PCI_CLASS_DISPLAY_VGA            = "0300"
+	PCI_CLASS_DISPLAY_3D             = "0302"
+	PCI_CLASS_PROCESSING_ACCELERATOR = "1200"
+	INVENTORY_CONFIG_ENV_VAR         = "INVENTORY_GPU_CONFIG"
```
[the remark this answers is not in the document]
mlorenzofr: These PCI classes are used for default settings, in case no argument is set. Also, the names for the magic numbers are meant to provide the reader an explanation about their meaning.

> How is this env var set? Who sets it?
Currently, the environment var is not set by any process. This change opens the possibility that in the future it will be added on the `installer`, which is where we usually call the `inventory`.

> Why not use command line parameter?
I added two possibilities: using an environment variable or a configuration file. I added the environment variable because iinm the inventory is launched inside of a container during the `installer` process. By having the option to get the setting from an environment variable, it would be easy to implement it on a service like we do with the `PULL_SECRET` variable in the agent.

src/inventory/gpu.go
```diff
@@ -1,37 +1,144 @@
...
+	if envConfiguration != "" {
+		err := json.Unmarshal([]byte(envConfiguration), &config)
+		if err != nil {
+			return defaultConfig, fmt.Errorf("failed parsing JSON on %s: %v", INVENTORY_CONFIG_ENV_VAR, err)
+		}
+
+		return config, nil
+	}
+
+	if path != "" {
```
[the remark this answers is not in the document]
mlorenzofr: Yes it is. Should I specifically write it in the documentation to avoid confusion?
The reason is that I think the environment variable will have better integration in future, if we execute the inventory in a containerized environment

src/inventory/gpu.go
mlorenzofr: I added the whitespaces to improve readability when someone adds new entries. We can see 3 blocks (pci class + vendor-id + device-id), and identify in on shot the 3 properties. At first I even had 4 blocks because I split the 1st block into a pci base class + pci subclass.

But beyond readability, you're absolutely right, I could remove the whitespaces and it should work. We can avoid problems with sequential whitespaces.

README.md
```diff
@@ -30,6 +30,49 @@ identifier (UUID) of the host that the agent is running on.
...
+# Include Nvidia (0x10de) 3D controllers (0x302)
+vendors:
+  - '0300 10de'
+  - '0302 10de'
+```
+
+This configuration can also be set using the **environment variable** `INVENTORY_GPU_CONFIG`.
+```yaml
+export INVENTORY_GPU_CONFIG='{"vendors": ["0300 10de", "0302 10de"]}'
+```
```
CrystalChun, line 69: Question for my understanding: this must be set on the host before the Agent runs the inventory command? Can it be changed while it's running or afterwards?

src/inventory/gpu.go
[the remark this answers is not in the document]
CrystalChun: > Should I specifically write it in the documentation to avoid confusion?

Yes that would be great!

src/inventory/gpu.go
```diff
@@ -1,37 +1,144 @@
...
+		}
+
+		fileSize := fileInfo.Size()
+		data := make([]byte, fileSize)
+		_, err = file.Read(data)
+		if err != nil {
+			return defaultConfig, fmt.Errorf("failed to read file: %v", err)
+		}
+
+		err = yaml.Unmarshal(data, &config)
```
CrystalChun, line 76: Is the file expected to be in yaml format?

README.md
```diff
@@ -30,6 +30,49 @@ identifier (UUID) of the host that the agent is running on.
 * *--cacert*: Path to a custom CA certificate file in PEM format.
 * *--help*: Print help message and exit.
 
+### Inventory Flags
+* *--gpu-config-file*: Path to a configuration file to filter the GPU discovery process.
```
CrystalChun, line 34: Should this file exist before the agent runs? Can it be changed while the agent is running?

README.md
mlorenzofr: It needs to be set before running the `inventory` command. iinm, the `inventory` is executed in one of the steps of `agent.service`, during the `installer` process. We could set the environment variable when we start the service (like we do with `PULL_SECRET`) and the configuration would be available to the `inventory` when it's launched.

TBH, I need to carefully review all the agent workflow on the `installer`. I don't have a clear design how to integrate it, or the requirements by the hypothetical product that will use this custom feature.

src/inventory/gpu.go
mlorenzofr: Yes, the configuration file is in yaml format. I used this format for the configuration file because I think it is easier to manipulate by a human than json, and it allow comments.
I followed the policy of files/manifests in yaml, environment variables in json, as we do in other components.

src/inventory/gpu_test.go
```diff
@@ -45,73 +84,88 @@ var (
...
+		os.Setenv("INVENTORY_GPU_CONFIG", `{"vendors": ["0302 10de"]}`)
+		gpus := GetGPUs(subprocessConfig, dependencies)
+
+		Expect(gpus).ToNot(BeNil())
+		Expect(gpus).To(ConsistOf(&gpu2))
 	})
 
-	table.DescribeTable("should handle incomplete data", func(card ghw.GraphicsCard, expectedGpu models.Gpu) {
-		dependencies.On("GPU").Return(&ghw.GPUInfo{GraphicsCards: []*ghw.GraphicsCard{&card}}, nil).Once()
+	It("should handle error gracefully", func() {
```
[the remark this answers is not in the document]
mlorenzofr: If we get an error when listing PCI devices (`nil` in the first returned value, an error in the second), then `gpus` object should exist, but be empty.

src/inventory/gpu.go
CrystalChun: Agreed, it's a good idea for sure and makes sense!

README.md
mlorenzofr: The `inventory` is executed at a very specific step of the agent workflow, it's not running as a process waiting for a signal. Therefore, configuration file should exist before of `inventory` execution, but can be created/modified while the `agent` is running.

iinm the agent execs the `inventory` inside of a container, `inventory`/`agent` are the same binary but depending on the name of the binary, the main function jumps to one entrypoint or another.

src/inventory/gpu_test.go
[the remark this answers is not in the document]
CrystalChun: Ah ok, maybe I'm missing what's causing it to error? Or maybe that doesn't matter?

README.md
CrystalChun: Yeah I'm worried the ux of this might not be ideal in case the user doesn't have access to the host beforehand to add this environment variable or the file before our agent runs. Or the user might not do this in time before the `inventory` command is ran.

Perhaps a flow to consider is maybe having the user create their GPU configs that pass to assisted-service, which then passes it to the Agent here?

CrystalChun: Not opposed to keeping what you have for now until this is integrated into assisted-service, though!

README.md
CrystalChun: Similar concern as the env var https://github.com/openshift/assisted-installer-agent/pull/787#discussion_r1777119682

Maybe we can mention in the doc the timing of these things existing?

README.md
CrystalChun: Also similarly, can we mention that this must be set before running the agent?

src/inventory/gpu_test.go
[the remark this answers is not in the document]
mlorenzofr: The error would be generated by the `ghw` library, but something very strange must happen to get this situation, maybe the `sysfs` is not mounted or populated. It could happen with an error mounting the volume in a container for example.

mlorenzofr:
> Does the change allow to detect other PCI devices (non GPU)?

Yes, indeed Gaudi devices are FPGAs and classified under the _processing accelerators_ branch (`0x12` instead of `0x03`). However they will be added to GPU list.

But for this development the requirements were to detect Nvidia and Gaudi cards as GPUs to find them in the inventory. RHEL AI uses them and detecting this hardware using the inventory would do some work easier.

The changes added here should:
* Detect and add Nvidia, Gaudi and standard GPUs by default (previously Nvidia/Gaudi cards were not detected)
* Provide a new mechanism to add custom hardware, of filter it, without changing the agent code.

README.md
```diff
@@ -30,6 +30,54 @@ identifier (UUID) of the host that the agent is running on.
 * *--cacert*: Path to a custom CA certificate file in PEM format.
 * *--help*: Print help message and exit.
 
+### Inventory Flags
+> [!WARNING]
+> The inventory process is tipically executed during the agent provisioning workflow. Therefore, the inventory configuration (either via configuration file or environment variable) must be set before reaching the step of calling the `inventory` binary.
```
CrystalChun, line 35: Looks good! Minor spelling 
```suggestion
> The inventory process is typically executed during the agent provisioning workflow. Therefore, the inventory configuration (either via configuration file or environment variable) must be set before reaching the step of calling the `inventory` binary.
```

README.md
CrystalChun: Looks good, thanks!

README.md
mlorenzofr: Warning added on [L34](https://github.com/openshift/assisted-installer-agent/pull/787/files#r1777176199)

CrystalChun: Ah sorry, is `installer` the assisted-service or openshift-installer?

src/inventory/gpu.go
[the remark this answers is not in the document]
mlorenzofr: I added a note box to the documentation. @ori-amizur does it look good to you or should I add some additional changes?

src/inventory/gpu_test.go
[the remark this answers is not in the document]
CrystalChun: Oh gotcha! Ok so you're simulating something going wrong with the ghw library and expecting the controller to still return a non-nil but empty gpu list?

README.md
mlorenzofr: :no_mouth: good question, I was thinking in `openshift-installer` but let me follow my people's tradition (galician), reply a question with a new one:

does the `assisted-service` create the ignition file when the ISO is generated or does it delegate that function to an `openshift-installer` package?

CrystalChun: I also follow your people's traditions too 😆 
AFAIK `assisted-service` does create the ignition file ([code for creating discovery image](https://github.com/openshift/assisted-service/blob/master/internal/ignition/discovery.go))

CrystalChun: We can also pass things directly to this inventory command too https://github.com/openshift/assisted-service/blob/master/internal/host/hostcommands/inventory_cmd.go

src/inventory/gpu_test.go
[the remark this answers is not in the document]
CrystalChun: Ah ok, thanks for the explanations!

README.md
mlorenzofr: Thanks for sharing :top: !!!!
We would definitely need to add some integration here. IMHO the best place for the configuration would be in the discovery image creation code (feeding the beast), but it will be possible to do it in the `inventory_cmd` as well.
I need to read this code, it's long but very interesting for future integrations. I didn't know about it.

CrystalChun: Sweet! I have a question about this configuration - sorry I didn't fully understand: should the gpu configuration be something that's generated by the service then? Or is it something we expect the user to give to us and assisted service should pass it along?

mlorenzofr: There is no proposal yet, we should have a discussion in the team on how to do it when the roadmap is clearer. Maybe @gamli75 or @oourfali have more information.

In my opinion, and it's just my two cents, I would use a set of known hardware that we can control. As product we can say: "We have this certified hardware, so we generate a configuration with their IDs". These AI cards are not just a hardware that we can plug&play, we depend on drivers and tools provided in the RHEL AI solution.

We maintain other configurations like the case of releases, images or srv-io hardware. I don't believe this hardware list is complicated, at least not more than the drivers themselves. Besides that, I have included support for PCI classes and vendors, it can simplify a lot the configuration if we use generic targets.

CrystalChun: That makes sense. Thank you for the explanation!

CrystalChun (commented):
Nice work! Looks good, but will wait for Ori!

src/inventory/gpu.go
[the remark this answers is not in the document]
ori-amizur: The other env variables are consumed as environment variables here: https://github.com/openshift/assisted-service/blob/8d9c72cddb29e2096e936d446efd8a0ed254f796/internal/ignition/templates/agent.service#L6-L12

In addition, these vars are consumed by all the agent processes.  The GPU var is only used by the inventory.  So I think it is better to put it as command line argument to the inventory process.

src/inventory/gpu.go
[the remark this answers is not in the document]
ori-amizur: I just wonder if we wonder if we need to allow both, or maybe we should warn the user when there is such attempt?

src/inventory/gpu.go
ori-amizur: I don't think you should remove whitespaces support, but I think you should be able to handle multiple whitespaces as if they were single whitespace.

mlorenzofr: Ok, I understand. I can trim the whitespace and parse the individual characters. No problem.

src/inventory/gpu.go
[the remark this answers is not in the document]
mlorenzofr: Ok, I can implement it via command line argument. Currently, I have implemented the configuration file as argument, but I implemented the environment variable for some environments (containerized) where using an environment variable can be easier than mounting a volume with an external configuration file.
In your opinion, would you pass an argument with a configuration file or would you directly pass the configuration (e.g. in json format) as argument?

src/inventory/gpu.go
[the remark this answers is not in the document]
mlorenzofr: Actually, I added this to provide more flexibility to the user in multiple scenarios (containerized or not). I could add a log message warning about the environment variable being applied over the configuration file, but if I remove the environment variable option (discussed at https://github.com/openshift/assisted-installer-agent/pull/787#discussion_r1780048856), we can ignore this.

PR #792, opened 2024-10-01
ACM-14598: Libraries bump to mitigate CVE-2024-27289

paul-maidment:
The goal of this PR is to ensure that the github.com/jackc/pgx library is no longer on 4.16.0 which is vulnerable to CVE-2024-27289

By replacing the library, we achieve this.

Also, as we were unable to build a recent version of this project due to incompatible versions, the go version has been bumped to 1.21

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/openshift/assisted-installer-agent
 
-go 1.20
+go 1.21
+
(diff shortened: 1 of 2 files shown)


paul-maidment:
Will be covered by changes being made in https://github.com/openshift/assisted-installer-agent/pull/801

PR #817, opened 2024-10-28
ACM-15078: Fixing agents reuse problem in z/VM after reclaim

veera-damisetti:
- Updated bootLoaderConfigTemplateS390x to add ai.ip_cfg_override to take care of IP and nameserver parameters , which is a mandatory parameter in z/VM as mac address is not persistent. 
- Updated paramaters which are being passed to zipl , to include parameters from /proc/cmdline which will help in reusing the agents after reclaim.

### DIFF
--- a/src/commands/actions/reboot_for_reclaim.go
+++ b/src/commands/actions/reboot_for_reclaim.go
@@ -3,7 +3,9 @@ package actions
 import (
 	"encoding/json"
 	"fmt"
+	"regexp"
 	"runtime"
+	"strings"
 	"syscall"
 
 	"github.com/openshift/assisted-installer-agent/src/util"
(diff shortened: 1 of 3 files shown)


CrystalChun (approved):
/lgtm

CrystalChun:
Would be great to see how and why these additional parameters are needed specifically for Z in order for the hosts to be reusable! Maybe a doc or blog post would be the route to show everyone?

veera-damisetti:
Sure @CrystalChun 
In order to reuse the hosts, hosts should be booted with proper/desired network and storage configurations.

In case of Z ( z/VM) , 
- IP and nameserver should be added to cmdline parameters , as using NMState is not possible, because mac address is not persistent in z/VM. 
- Network configurations should be included in cmdline parameters, using rd.znet and ip . which through we can give desired network cards for connectivity.
- Similarly, we should give storage configuraions also with rd.zfcp,rd.dasd and zfcp.allow_lun_scan, depends on the type of the disk which we want to configure for z/VM.

HCP IBMZ doc for z/VM for cmdline parameter reference: https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.11/html/clusters/cluster_mce_overview#hosted-bare-metal-adding-agents-ibmz-zvm-lpar

RH doc for explaining more details about each param: https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/7/html/installation_guide/chap-installer-booting-ipl-s390#chap-installer-booting-ipl-s390

src/commands/actions/reboot_for_reclaim.go
```diff
@@ -29,21 +31,64 @@ func (a *rebootForReclaim) Run() (stdout, stderr string, exitCode int) {
 	}
 
 	if runtime.GOARCH == "s390x" {
+		var options string
+		cmdline_params := make(map[string]string)
+		var requiredCmdline string
+
+		stdout, stderr, exitCode = util.Execute("cat", "/boot/loader/entries/00-assisted-discovery.conf")
+		if exitCode != 0 {
+			return stdout, stderr, exitCode
```
eifrach, line 40: the rest of the code is depending on this.
we should add a stop here

src/commands/actions/reboot_for_reclaim.go
```diff
@@ -29,21 +31,64 @@ func (a *rebootForReclaim) Run() (stdout, stderr string, exitCode int) {
...
+		lines := strings.Split(stdout, "\n")
+		for _, line := range lines {
+			if strings.HasPrefix(line, "options") {
+				options = strings.TrimSpace(strings.TrimPrefix(line, "options"))
+				break
+			}
+		}
+		stdout, stderr, exitCode := util.Execute("cat", "/proc/cmdline")
+		if exitCode != 0 {
+			return stdout, stderr, exitCode
```
eifrach, line 51: same here, if failed no point of continuing

src/commands/actions/reboot_for_reclaim.go
```diff
@@ -29,21 +31,64 @@ func (a *rebootForReclaim) Run() (stdout, stderr string, exitCode int) {
...
+			regex := regexp.MustCompile(fmt.Sprintf(`\b%s=([^\s]+)`, param))
+			match := regex.FindStringSubmatch(stdout)
+			if len(match) > 1 {
+				cmdline_params[param] = match[1]
+			}
+		}
+
+		for key, value := range cmdline_params {
+			requiredCmdline += fmt.Sprintf("%s=%s ", key, value)
+		}
```
eifrach, line 74: I think we should create a function for this and a unit test.

once we start to use regex it can break in future changes  + make this function more readable

src/commands/actions/reboot_for_reclaim.go
```diff
@@ -29,21 +31,64 @@ func (a *rebootForReclaim) Run() (stdout, stderr string, exitCode int) {
...
+			"nameserver",
+			"rd.znet",
+			"zfcp.allow_lun_scan",
+			"rd.zfcp",
+			"rd.dasd",
+		}
+
+		for _, param := range paramsToExtract {
+			regex := regexp.MustCompile(fmt.Sprintf(`\b%s=([^\s]+)`, param))
+			match := regex.FindStringSubmatch(stdout)
```
eifrach, line 66: what happens if we don't have a match ?

eifrach (changes requested):
how is the unit-test passed without a change?
we are missing here a lot for testing to this logic

src/commands/actions/reboot_for_reclaim.go
veera-damisetti: If there’s no match for a parameter in stdout, the current code skips over assigning it, so cmdline_params[param] remains unset.
and this should be the expected behaviour, if paramsToExtract  are not present in /proc/cmdline

src/commands/actions/reboot_for_reclaim.go
veera-damisetti: Sure @eifrach 
Do you want me to write entire s390x logic in a seperate function and unit test ?  or just this fetching parameters using regexp as a seperate function ?

veera-damisetti:
@eifrach , thanks for the review. 

I did changes to have a separate function for the logic and added unit tests for the same, and 
Please have a look.

src/commands/actions/reboot_for_reclaim.go
eifrach: yes just the regex / paramter fatching. cause we do want make sure that changes wont break it later on

veera-damisetti: Sure

veera-damisetti:
Thanks @CrystalChun @eifrach , for the quick reviews.

PR #819, opened 2024-11-08
MGMT-18464: MGMT-18452: Add validations to make iSCSI disk eligible

adriengentil:
This PR adds 2 checks before making iSSI disks be be eligible as
installation disk.

[MGMT-18464](https://issues.redhat.com//browse/MGMT-18464): Check the state of the disk, we need to ensure that it is
"running" otherwise the installtion of RHCOS will fail on the disk.

[MGMT-18464](https://issues.redhat.com//browse/MGMT-18464): Check if the discovery as been booted using iBFT, this check
ensure that the iSCSI disk is actually a disk on which we will be able
to boot from.

### DIFF
--- a/src/inventory/disks.go
+++ b/src/inventory/disks.go
@@ -341,6 +341,9 @@ func (d *disks) checkEligibility(disk *ghw.Disk) (notEligibleReasons []string, i
 		notEligibleReasons = append(notEligibleReasons, "Disk is an LVM logical volume")
 	}
 
+	notEligibleReasons = append(notEligibleReasons, d.checkEligibilityISCSIState(disk)...)
+	notEligibleReasons = append(notEligibleReasons, d.checkEligibilityISCSIinIBFT(disk)...)
+
 	// Don't check partitions if this is an appliance disk, as those disks should be marked as eligible for installation.
 	for _, partition := range disk.Partitions {
 		if strings.HasPrefix(partition.Label, applianceAgentPrefix) {
@@ -530,24 +533,14 @@ func (d *disks) getBusPath(disks []*block.Disk, index int, busPath string) strin
 
 // getISCSIHostIPAddress retuns the IP address in use to connect on the iSCSI volume
 func (d *disks) getISCSIHostIPAddress(diskName string) string {
-	// resolve /sys/block/sda ->  /sys/devices/platform/host2/session1/target2:0:0/2:0:0:1/block/sda
-	sysBlockPath := filepath.Join("/sys", "block", diskName)
-	sysDevicesPath, err := d.dependencies.EvalSymlinks(sysBlockPath)
+	iSCSIHost, err := d.getISCSIHost(diskName)
 	if err != nil {
-		logrus.WithError(err).Errorf("Failed to evaluate symlink for iSCSI device")
-		return ""
-	}
-
-	// extract iSCSI host
-	const hostIndex = 4
-	splitPath := strings.Split(sysDevicesPath, "/")
-	if len(splitPath) < (hostIndex + 1) {
-		logrus.Errorf("Failed to resolve iSCSI host in path %s", sysDevicesPath)
+		logrus.WithError(err).Errorf("Failed to resolve iSCSI host")
 		return ""
 	}
 
 	// Read IP address
-	hostIPAdressFile := fmt.Sprintf("/sys/class/iscsi_host/%s/ipaddress", splitPath[hostIndex])
+	hostIPAdressFile := filepath.Join("/sys/class/iscsi_host", iSCSIHost, "ipaddress")
 	data, err := d.dependencies.ReadFile(hostIPAdressFile)
 	if err != nil {
 		logrus.WithError(err).Errorf("Failed to read host IP address file for iSCSI device")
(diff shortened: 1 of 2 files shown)


adriengentil:
/hold
Might had some more tests

src/inventory/disks.go
```diff
@@ -561,3 +554,108 @@ func (d *disks) getISCSIHostIPAddress(diskName string) string {
...
+
+	for _, file := range files {
+		if !file.IsDir() || !strings.Contains(file.Name(), "target") {
+			continue
+		}
+
+		ibftTargetNameFile := filepath.Join(ibftBasePath, file.Name(), "target-name")
+		ibftTargetData, err := d.dependencies.ReadFile(ibftTargetNameFile)
+		if err != nil {
+			logrus.WithError(err).Errorf("Failed to read iSCSI target name in iBFT")
```
adriengentil, line 624:
```suggestion
			logrus.WithError(err).Errorf("Failed to read iSCSI target name in iBFT")
			continue
```

src/inventory/disks.go
```diff
@@ -530,24 +533,14 @@ func (d *disks) getBusPath(disks []*block.Disk, index int, busPath string) strin
 
 // getISCSIHostIPAddress retuns the IP address in use to connect on the iSCSI volume
 func (d *disks) getISCSIHostIPAddress(diskName string) string {
-	// resolve /sys/block/sda ->  /sys/devices/platform/host2/session1/target2:0:0/2:0:0:1/block/sda
-	sysBlockPath := filepath.Join("/sys", "block", diskName)
-	sysDevicesPath, err := d.dependencies.EvalSymlinks(sysBlockPath)
+	iSCSIHost, err := d.getISCSIHost(diskName)
```
eifrach, line 536: maybe we should check this before entering this function.

src/inventory/disks.go
```diff
@@ -561,3 +554,108 @@ func (d *disks) getISCSIHostIPAddress(diskName string) string {
...
+	// # ls -l
+	// total 0
+	// drwxr-xr-x. 2 root root 0 Nov  7 10:14 acpi_header
+	// drwxr-xr-x. 2 root root 0 Nov  7 10:14 ethernet0
+	// drwxr-xr-x. 2 root root 0 Nov  7 10:14 initiator
+	// drwxr-xr-x. 2 root root 0 Nov  7 10:14 target0
+	//
+	// targetN directories contain "target-name" file, which will try to
+	// match with the target used by the iSCSI volume on the host.
+	const ibftBasePath = "/sys/firmware/ibft"
```
eifrach, line 610: this should move to top of the file with the other const

src/inventory/disks.go
```diff
@@ -341,6 +341,9 @@ func (d *disks) checkEligibility(disk *ghw.Disk) (notEligibleReasons []string, i
 		notEligibleReasons = append(notEligibleReasons, "Disk is an LVM logical volume")
 	}
 
+	notEligibleReasons = append(notEligibleReasons, d.checkEligibilityISCSIState(disk)...)
+	notEligibleReasons = append(notEligibleReasons, d.checkEligibilityISCSIinIBFT(disk)...)
+
```
eifrach, line 346: if those function retruns a string value - should we continue?

src/inventory/disks.go
```diff
@@ -561,3 +554,108 @@ func (d *disks) getISCSIHostIPAddress(diskName string) string {
 func GetDisks(subprocessConfig *config.SubprocessConfig, dependencies util.IDependencies) []*models.Disk {
 	return newDisks(subprocessConfig, dependencies).getDisks()
 }
+
+// Check if the state of the iSCSI disk is running
+func (d *disks) checkEligibilityISCSIState(disk *ghw.Disk) []string {
+	if !isISCSIDisk(disk) {
```
eifrach, line 560: second time we check this - we can check before those two functions once - if it false we can skip both functions

eifrach (changes requested):
We should think about moving most of the logic to the service and use the agent to collect the information.

In some cases, that might not be a great option. but, by leaving the logic in the service, this will make the agent simpler and allow us to patch the service without needing to redeploy a new agent.

adriengentil:
>We should think about moving most of the logic to the service and use the agent to collect the information.

in this case the service won't make use of this information, except for validations, that's why I put them here. I don't expect this code to change often, so I think it's ok for the moment to have them here, I will think about moving them in the service if we need further updates and if we need more flexibility.

src/inventory/disks.go
adriengentil: I wanted them to be seft-contained, but ok, that works too.

src/inventory/disks.go
adriengentil: this function gathers **all** the possible reasons for a disk to not be eligible for installation, so we need to continue.

src/inventory/disks.go
adriengentil: done

src/inventory/disks.go
adriengentil: done, moved the check out

PR #821, opened 2024-11-18
MGMT-19166:  Implement MTU checker

linoyaslan:
As part of the MTU validation implementation, the agent will include the MTU checker in the connectivity check. It will send reports to the service if a ping with the specified MTU size fails.

### DIFF
--- a/src/connectivity_check/mtu_checker.go
+++ b/src/connectivity_check/mtu_checker.go
@@ -1,0 +1,95 @@
+package connectivity_check
+
+import (
+	"net"
+	"sort"
+	"strconv"
+
+	"github.com/openshift/assisted-service/models"
+	log "github.com/sirupsen/logrus"
+)
+
+// Since the exact headers in use are uncertain, the total overhead is estimated to be a maximum of ~100 bytes.
+// This includes 18 bytes for VLAN, a maximum of 60 bytes for IPv4/IPv6, 8 bytes for ICMP, and potential additional headers.
+// To be on the safe side, we assume an overhead of 150 bytes.
+const headers = 150
+
+type mtuChecker struct {
+	executer Executer
+}
+
+func (m *mtuChecker) Features() Features {
+	return OutgoingNicFeature
+}
+
+func (m *mtuChecker) Check(attributes Attributes) ResultReporter {
+	ret := models.MtuReport{
+		RemoteIPAddress: attributes.RemoteIPAddress,
+		OutgoingNic:     attributes.OutgoingNIC.Name,
+	}
+
+	// Check if the remote IP address is IPv4 or IPv6
+	remoteIP := net.ParseIP(attributes.RemoteIPAddress)
+	if remoteIP == nil {
+		log.Errorf("MTU checker: Invalid remote IP address %s", attributes.RemoteIPAddress)
+		return newMtuResultReporter(&ret)
+	}
+	isRemoteIPv6 := remoteIP.To4() == nil // If To4() is nil, it's IPv6
+
+	var localIP string
+
+	for _, addr := range attributes.OutgoingNIC.Addresses {
+		ipN, ok := addr.(*net.IPNet)
+		if !ok {
+			log.Errorf("MTU checker: failed convert address %v", addr)
+			return newMtuResultReporter(&ret)
+		}
+
+		localIP = ipN.IP.String()
+
+		// Check if the local IP address is IPv4 or IPv6
+		isLocalIPv6 := ipN.IP.To4() == nil // If To4() is nil, it's IPv6
+		if isLocalIPv6 != isRemoteIPv6 {
+			continue
+		}
+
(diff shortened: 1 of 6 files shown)


src/connectivity_check/mtu_checker.go
```diff
@@ -0,0 +1,103 @@
...
+		// Check if the local IP address is IPv4 or IPv6
+		isLocalIPv6 := ipN.IP.To4() == nil // If To4() is nil, it's IPv6
+		if isLocalIPv6 != isRemoteIPv6 {
+			continue
+		}
+
+		// Check if remote IP is in the same network as the local IP
+		if isSameNetwork(localIP, attributes.RemoteIPAddress, subnetMask, isRemoteIPv6) {
+			mtu := attributes.OutgoingNIC.MTU
+			mtuWithoutHeaders := int(float64(mtu) * 0.9)
```
ori-amizur, line 55: Why use this calculation? The headers size should be fixed (IP headers + ICMP headers).  They are not proportional to the MTU value

src/connectivity_check/mtu_checker.go
```diff
@@ -0,0 +1,103 @@
...
+		localIP = ipN.IP.String()
+
+		// Check if the local IP address is IPv4 or IPv6
+		isLocalIPv6 := ipN.IP.To4() == nil // If To4() is nil, it's IPv6
+		if isLocalIPv6 != isRemoteIPv6 {
+			continue
+		}
+
+		// Check if remote IP is in the same network as the local IP
+		if isSameNetwork(localIP, attributes.RemoteIPAddress, subnetMask, isRemoteIPv6) {
```
ori-amizur, line 53: Why is this required? UMN allows different networks

linoyaslan: Because in the MTU check, I send a ping from a specific interface, that interface must be on the same network as the remote IP. Otherwise, the ping will fail, but not due to an MTU issue. To avoid this, I added this check.

src/connectivity_check/mtu_checker.go
eifrach: @ori-amizur 
The name is misleading, but we want to ensure that on some types of network we will consider encapsulation overhead  

For example:  dot1q ( 802.1q), Q&Q, VXLAN’s, different tunnel types,  etc…

removing 10% size will allow the overhead and make the test more reliable

src/connectivity_check/mtu_checker.go
linoyaslan: Regarding UMN, you are correct, but there is no conflict here. All IPs across all interfaces will be checked eventually, but the ping will target the correct IP addresses to avoid failures.

src/connectivity_check/mtu_checker.go
linoyaslan: Thanks @eifrach . I can rename it to make it clearer.

ori-amizur: I don't think that for MTU 1500 and MTU 9000 we should have different values of headers size (or whatever you want to call it).
In addition if for some kind of reason the MTU is very small, then the overhead might be too small since it might be more that 10%.

src/connectivity_check/mtu_checker.go
ori-amizur: ping (as opposed to arping) does not fail across different networks

src/connectivity_check/mtu_checker.go
```diff
@@ -0,0 +1,95 @@
...
+	"sort"
+	"strconv"
+
+	"github.com/openshift/assisted-service/models"
+	log "github.com/sirupsen/logrus"
+)
+
+// Since the exact headers in use are uncertain, the total overhead is estimated to be a maximum of ~100 bytes.
+// This includes 18 bytes for VLAN, a maximum of 60 bytes for IPv4/IPv6, 8 bytes for ICMP, and potential additional headers.
+// To be on the safe side, we assume an overhead of 150 bytes.
```
zaneb, line 14: This means the configured MTU can be up to 122 bytes too large for the network without us detecting it.

eifrach: You can’t predict the network - if they have VXLAN on the network it will also be too big for it.
We should try to cover most cases and not be 100% accurate cause it’s not possible and will result in lots of unwanted failures.

src/connectivity_check/mtu_checker.go
```diff
@@ -0,0 +1,95 @@
...
+import (
+	"net"
+	"sort"
+	"strconv"
+
+	"github.com/openshift/assisted-service/models"
+	log "github.com/sirupsen/logrus"
+)
+
+// Since the exact headers in use are uncertain, the total overhead is estimated to be a maximum of ~100 bytes.
```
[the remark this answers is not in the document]
eifrach: you can also have network overhead that the host is not aware of

PR #822, opened 2024-11-20
MGMT-19331: Add `secure_boot_state` to inventory

jhernand:
This patch changes the agent so that it checks if secure boot is enabled and adds that to the inventory. To do the check it uses reads the state from the `/sys/firmware/efi/efivars/SecureBoot-...` file.

Related: https://issues.redhat.com/browse/MGMT-19331
Related: https://issues.redhat.com/browse/MGMT-19056


src/commands/actions/inventory_cmd.go
```diff
@@ -62,10 +72,40 @@ func (a *inventory) Args() []string {
...
+	// unconditionally, as that will make the podman command fail.
+	const efivarsPath = "/sys/firmware/efi/efivars"
+	efivarsLogger := logrus.WithFields(logrus.Fields{
+		"path": efivarsPath,
+	})
+	_, err := a.filesystem.Stat(efivarsPath)
+	if os.IsNotExist(err) {
+		efivarsLogger.WithFields(logrus.Fields{
+			"path": efivarsPath,
+		}).Info("EFI variables filesystem isn't mounted")
```
adriengentil, line 87:
```suggestion
		efivarsLogger.Info("EFI variables filesystem isn't mounted")
```
`efivarsLogger` already contains this field

adriengentil (commented):
Looks great just a small comment

src/commands/actions/inventory_cmd.go
jhernand: Indeed, thanks @adriengentil.

jhernand: Done.

adriengentil (approved):

PR #857, opened 2024-12-16
OCPBUGS-45910: Fix cpu clockspeed for s390x

apodvrat:
For s390x architectures lscpu does not return the clock speed by "CPU MHz" or "CPU max MHz" field but have two fields (CPU static MHz and CPU dynamic MHz). 
"CPU static MHz" is the field representing the designed clock speed whereas the "CPU dynamic MHz" is representing the current CPU speed (throttled or energy saving).
New test case for s390x architecture was added. 
In addition the machine type was fixed, too.

The fix was tested using unit tests:

JUnit path was configured: ./junit_unit_test.xml

JUnit report was created: /root/code/Assisted_Installer/latest/assisted-installer-agent/src/inventory/junit_unit_test.xml

Ran 121 of 121 Specs in 0.268 seconds
SUCCESS! -- 121 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestInventory (0.30s)

### DIFF
--- a/src/inventory/cpu.go
+++ b/src/inventory/cpu.go
@@ -43,11 +43,11 @@ func GetCPU(dependencies util.IDependencies) *models.CPU {
 		switch f.Field[:len(f.Field)-1] {
 		case "Architecture":
 			ret.Architecture = f.Data
-		case "Model name":
+		case "Model name", "Machine type":
 			ret.ModelName = f.Data
 		case "CPU(s)":
 			ret.Count, _ = strconv.ParseInt(f.Data, 10, 64)
-		case "CPU MHz", "CPU max MHz":
+		case "CPU MHz", "CPU max MHz", "CPU static MHz":
 			f, _ := strconv.ParseFloat(f.Data, 64)
 			ret.Frequency = max(ret.Frequency, f)
 		case "Flags":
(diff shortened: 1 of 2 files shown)


src/inventory/cpu.go
```diff
@@ -43,11 +43,11 @@ func GetCPU(dependencies util.IDependencies) *models.CPU {
 		switch f.Field[:len(f.Field)-1] {
 		case "Architecture":
 			ret.Architecture = f.Data
-		case "Model name":
+		case "Model name", "Machine type":
```
paul-maidment, line 46: Might be worth a mention in the commit message that "Machine Type" is used in place of "Model name" in S390x `lscpu` responses.

apodvrat: Thank you for the comment. Done.

src/inventory/cpu_test.go
```diff
@@ -130,4 +173,17 @@ var _ = Describe("CPU test", func() {
 		}
 		Expect(ret).To(Equal(&expected))
 	})
+	It("s390x OK", func() {
+		dependencies.On("Execute", "lscpu", "-J").Return(s390xLscpuOutput, "", 0).Once()
+		ret := GetCPU(dependencies)
+		expected := models.CPU{
+			Architecture: "s390x",
+			Count:        4,
+			Flags: []string{"esan3", "zarch", "stfle" "msa", "ldisp", "eimm", "dfp", "edat", "etf3eh", "highgprs", "te", "vx",
```
paul-maidment, line 182: I missed this on first glance but the reason your tests are failing is that you are missing a comma between "stfle" and "msa"

paul-maidment (changes requested):

src/inventory/cpu_test.go
apodvrat: :-) Thank you ... I saw this bug but I'm dealing with two different development system because of my two different git account (IBM and RH) and fixed it only on one ... but it should work now ...

PR #877, opened 2024-12-22
NO-ISSUE: OCPBUGS-43352 calculate outgoing interface for arping

eifrach:
- arping is running on interfaces that are not in the same address range as the remote IP. 
In some proxy-arp enabled networks this causes false-positive IP duplication error.

This PR will validate the remote IP is in the same network as the outgoing interface

### DIFF
--- a/src/connectivity_check/ipv4_arping_checker.go
+++ b/src/connectivity_check/ipv4_arping_checker.go
@@ -1,6 +1,7 @@
 package connectivity_check
 
 import (
+	"net"
 	"os/exec"
 	"regexp"
 	"sort"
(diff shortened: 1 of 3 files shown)


src/connectivity_check/ipv4_arping_checker.go
```diff
@@ -24,61 +25,75 @@ func (p *arpingChecker) Check(attributes Attributes) ResultReporter {
...
+		if !nicNetwork.Contains(remoteIP) {
 			continue
 		}
-		remoteMAC := strings.ToLower(parts[2])
-		successful := macInDstMacs(remoteMAC, attributes.RemoteMACAddresses)
-		if !successful {
-			log.Warnf("Unexpected mac address for arping %s on nic %s: %s", attributes.RemoteIPAddress, attributes.OutgoingNIC.Name, remoteMAC)
+		result, err := p.executer.Execute("arping", "-c", "10", "-w", "5", "-I", attributes.OutgoingNIC.Name, attributes.RemoteIPAddress)
+		if err != nil {
+			// Ignore exit code of 1; only 2 or -1 are actual errors
```
paul-maidment, line 41: The `w` parameter is a timeout parameter and provided that this is higher than `c` then arping will await all responses, if it receives all responses, a zero exit code will be retuned.

At the very least I think we should mention the reason for receiving a non zero exit code.

```
paulmaidment@fedora ~ $ arping -c 6 -w 6 -I wlp0s20f3 192.168.20.1
ARPING 192.168.20.1 from 192.168.20.76 wlp0s20f3
Unicast reply from 192.168.20.1 [74:44:01:47:AA:34]  1.534ms
Unicast reply from 192.168.20.1 [74:44:01:47:AA:34]  1.489ms
Unicast reply from 192.168.20.1 [74:44:01:47:AA:34]  2.832ms
Unicast reply from 192.168.20.1 [74:44:01:47:AA:34]  4.550ms
Unicast reply from 192.168.20.1 [74:44:01:47:AA:34]  2.089ms
Unicast reply from 192.168.20.1 [74:44:01:47:AA:34]  2.830ms
Sent 6 probes (1 broadcast(s))
Received 6 response(s)
paulmaidment@fedora ~ $ echo $?
0
```

eifrach: I agree  but, I wonder where the information about the status code came from. ( on [commit](https://github.com/openshift/assisted-installer-agent/commit/d9b910b53544808ea726d8239d4b3cf3b31bc320) )

@zaneb  can you share where did you get the status codes? and/or what arpping version is it based on ?

src/connectivity_check/ipv4_arping_checker_test.go
```diff
@@ -122,7 +122,7 @@ Received 0 response(s)
 		attributes := Attributes{
 			RemoteIPAddress:    remoteIPAddress,
 			RemoteMACAddress:   remoteMACAddress,
-			OutgoingNIC:        OutgoingNic{Name: outgoingNIC, HasIpv4Addresses: true},
+			OutgoingNIC:        OutgoingNic{Name: outgoingNIC, HasIpv4Addresses: true, Addresses: getAddr(outgoingIPAddress, 24)},
```
linoyaslan, line 125: How about adding a new test where the remote address is not in the same network as the local IP? bc currently, all the scenarios only include cases where both the local and remote IPs are in the same network.

src/connectivity_check/ipv4_arping_checker.go
eifrach: after some checking,  it does fail on error code 1 - when there is no ping to the host. 
I've refactor it a bit and added a wait for ping before checking the arpping which makes it pass every time.

let me know what you think @paul-maidment

paul-maidment: Your changes in this area look good :)

src/connectivity_check/ipv4_arping_checker_test.go
eifrach: thanks for the input, Done I've added a test for remote address is not in the same subnet of the outgoing interface

src/connectivity_check/ipv4_arping_checker.go
```diff
@@ -24,61 +25,77 @@ func (p *arpingChecker) Check(attributes Attributes) ResultReporter {
...
 		}
-		remoteMAC := strings.ToLower(parts[2])
-		successful := macInDstMacs(remoteMAC, attributes.RemoteMACAddresses)
-		if !successful {
-			log.Warnf("Unexpected mac address for arping %s on nic %s: %s", attributes.RemoteIPAddress, attributes.OutgoingNIC.Name, remoteMAC)
+
+		result, err := p.executer.Execute("arping", "-c", "10", "-w", "5", "-I", attributes.OutgoingNIC.Name, attributes.RemoteIPAddress)
+		if err != nil {
+			if err.(*exec.ExitError).ExitCode() == 1 {
+				log.Warn("ARPPing exited with error code 1, ignoring")
```
paul-maidment, line 43: I still think we should make sure that we got at least one successful ping here. It is actually possible to force a return code of `1` when there is an erorr.

```
paulmaidment@paulmaidment:~$ sudo arping -c 6 -w 6 -I enp0s8 192.168.209.1
ARPING 192.168.209.1
Timeout
Timeout
Timeout
Timeout
Timeout
Timeout

--- 192.168.209.1 statistics ---
6 packets transmitted, 0 packets received, 100% unanswered (0 extra)

paulmaidment@paulmaidment:~$ echo $?
1
```

I think we should at least say 

```
"ARPPing did not receive all ping responses back but received at least one, ignoring.
```

paul-maidment (changes requested):

src/connectivity_check/ipv4_arping_checker.go
eifrach: @AlonaKaplan can we get you're input on this?

paul-maidment: Suggest we just add a comment above the log entry

```
// The return code of 1 will be returned when there are some missing replies (possibly all of them)
// as we are looking for "duplicate" replies and are not concerned with missing replies, we may ignore this.
```

paul-maidment (changes requested):

src/connectivity_check/ipv4_arping_checker.go
eifrach: thanks for the input

paul-maidment (approved):
/lgtm

src/connectivity_check/ipv4_arping_checker.go
zaneb: I believe [this](https://github.com/iputils/iputils/blob/20210202/arping.c) is the version in RHEL 9.
It has been largely rewritten since [the version in RHEL 8](https://github.com/iputils/iputils/blob/s20180629/arping.c), and now things that for decades used to work, don't.
I do not believe a single word the man page has to say about this.

zaneb: It looks like that man page quote comes from a [bug fix](https://github.com/iputils/iputils/commit/854873bdd28fcdd9cc3fe0c2d29c083a07d07a86) that is not (yet) present in RHEL9.

andfasano:
retitling the PR to allow the cherry-pick bot works as expected

/retitle OCPBUGS-43352: calculate outgoing interface for arping

PR #894, opened 2025-01-16
MGMT-19728: Change downstream dockerfiles to use rhel-minimal as the base image

giladravid16:
Change our downstream dockerfiles to use rhel-minimal as the base image, because it's smaller.
Also improve the README for rpm-prefetching.

Part-of [MGMT-19728](https://issues.redhat.com//browse/MGMT-19728)


giladravid16:
After testing this move, it adds CVEs to our final image, so canceling this task.

PR #897, opened 2025-01-17
MGMT-19100: Allow install to the existing root filesystem

carbonin:
This PR allows for installing coreos content to the existing root filesystem instead of only to an entire device.

When the install command contains the coreos image field it will be passed through to the installer.

Additionally the validation on eligible disks for iso type partitions is dropped.
This was done because it is not a valid way to detect installation media, but it is a way that config data can be provided to the host during a CAPI install.

For a full ISO the current install media partitions look like this:

```
[root@localhost core]# fdisk -l /dev/sr0
Disk /dev/sr0: 1.16 GiB, 1244659712 bytes, 607744 sectors
Disk model: QEMU DVD-ROM
Units: sectors of 1 * 2048 = 2048 bytes
Sector size (logical/physical): 2048 bytes / 2048 bytes
I/O size (minimum/optimal): 2048 bytes / 2048 bytes
Disklabel type: dos
Disk identifier: 0x47ad035f

Device     Boot Start     End Sectors  Size Id Type
/dev/sr0p1 *        0 2430975 2430976  4.6G  0 Empty
/dev/sr0p2        180   10139    9960 19.5M ef EFI (FAT-12/16/32)
```

Neither of these partitions are of type iso9660 so this check is doing nothing but making CAPI installs more difficult.

~~Leaving this as WIP until https://github.com/openshift/assisted-service/pull/7197 and https://github.com/openshift/assisted-installer/pull/1003 are merged~~ Merged


src/inventory/disks.go
```diff
@@ -360,15 +360,6 @@ func (d *disks) checkEligibility(disk *ghw.Disk) (notEligibleReasons []string, i
 
 	// Check disk partitions for type, name, and mount points:
 	for _, partition := range disk.Partitions {
-		if partition.Type == "iso9660" {
```
paul-maidment, line 363: I am not 100% sure but doesn't this code run for all block devices returned from the `ghw.Block(` call here?

https://github.com/openshift/assisted-installer-agent/blob/33758448dea620e5f9d939df11e14fc5b5a69680/src/inventory/disks.go#L441-L457

Is there ever a case where the installation media would be ISO9660 ?
Just for clarification, it's accurate to say that isInstallationMedia will only be set if the mountpoint of a partition has the suffix 'iso' is this the only scenario where we would consider the partition to be installation media?

carbonin: The installation media can be an iso but the partition never is.
I explain this in the commit an PR description, or are you asking about something else?

>Just for clarification, it's accurate to say that isInstallationMedia will only be set if the mountpoint of a partition has the suffix 'iso' is this the only scenario where we would consider the partition to be installation media?

Yes, that will be the situation after this commit is merged. And this will successfully eliminate the cd drive device. You can see this in the following example from a machine with the full ISO booted:
```
[root@localhost core]# mount | grep iso
/dev/sr0 on /run/media/iso type iso9660 (ro,relatime,nojoliet,check=s,map=n,blocksize=2048)
```

eranco74: Unsure why this change is part of this PR  or why this code is making CAPI installs more difficult.

eranco74:
@rccrdpccl unhold?

src/inventory/disks.go
carbonin: From the PR description:

>Additionally the validation on eligible disks for iso type partitions is dropped.
This was done because it is not a valid way to detect installation media, but it is a way that config data can be provided to the host during a CAPI install.

I also mention it in the commit message

PR #899, opened 2025-01-23
MGMT-19102: Add boot device info to inventory

carbonin:
Ephemeral devices are those backed by a loop device. This indicates that the system is not booted from a "real" disk. Otherwise the boot device is considered "persistent"

This indicates to assisted service how it should install this host. For ephemerally booted systems it will install directly to an entire device, but for persistently booted systems it will install to the existing root filesystem

https://issues.redhat.com/browse/MGMT-19102

~~WIP until https://github.com/openshift/assisted-service/pull/7221 is merged and the new APIs are properly included here.~~ Done


src/inventory/boot.go
```diff
@@ -103,12 +103,24 @@ func (b *boot) getSecureBootState() models.SecureBootState {
 	}
 }
 
+func (b *boot) getDeviceType() string {
+	stdOut, _, code := b.dependencies.ExecutePrivileged("findmnt", "--noheadings", "--output", "SOURCE", "/sysroot")
```
paul-maidment, line 107: I tried this command on my local commandline

```
$ findmnt --noheadings --output SOURCE --source '/doesnotexist'
$ echo $?
1
```

I'm assuming that we will always have a mount point called '/sysroot'?, though don't know the full context of how the BootDeviceTypePersistent hosts are going to be booted. 

My point is that if this mountpoint were to be absent (`code != 0`), we would not return a device type at all.

I wonder if we should at least log this scenario just in case?

rccrdpccl:
why WIP? Looks good to me :)

/lgtm
/hold

feel free to unhold

carbonin:
>why WIP?

image

PR #904, opened 2025-02-05
NO-ISSUE: Increase image pull timeout during install

CrystalChun:
Previously, there was a timeout of 30 seconds when pulling the installer image. This is not enough
time to pull the image and to mark the installation as failed.

This change increases the timeout to 600 seconds
(10 minutes), which is more accurate for a failed
image pull. The error message from failing to pull has also been updated to reflect the actual failure.

/cc @omertuc

### DIFF
--- a/src/commands/actions/install_cmd.go
+++ b/src/commands/actions/install_cmd.go
@@ -27,7 +27,7 @@ const (
 	templateGetImage          = "podman images --quiet %s"
 	failedToPullImageExitCode = 2
 	defaultImagePullRetries   = 3
-	defaultImagePullTimeout   = 30
+	defaultImagePullTimeout   = 600
 )
 
 var podmanBaseCmd = [...]string{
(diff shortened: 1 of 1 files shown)


omertuc:
For future reference the timeout was introduced in a7a0bc03f92e77a81de48e49f362ffe3d6d2d9a8 

Does that require a backport?

omertuc (approved):

CrystalChun:
> For future reference the timeout was introduced in [a7a0bc0](https://github.com/openshift/assisted-installer-agent/commit/a7a0bc03f92e77a81de48e49f362ffe3d6d2d9a8)
> 
> Does that require a backport?

Yep! This was introduced in MCE 2.6 (ACM 2.11) and OCP 4.16. 
I think that we should backport it to: 

- ACM 2.12 and ACM 2.11 
- OCP 4.16, OCP 4.17, and OCP 4.18

CrystalChun:
/retitle OCPBUGS-50655: Increase image pull timeout during install

PR #910, opened 2025-02-13
MGMT-19927: Modify MTU Check to Only MTUs  != 1500

linoyaslan:
This PR updates the MTU check to apply only to MTUs != 1500. Additionally, it adjusts the header subtraction to account only for IPv4/IPv6 headers.

/cc @zaneb @AlonaKaplan

### DIFF
--- a/src/connectivity_check/mtu_checker.go
+++ b/src/connectivity_check/mtu_checker.go
@@ -9,10 +9,9 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-// Since the exact headers in use are uncertain, the total overhead is estimated to be a maximum of ~100 bytes.
-// This includes 18 bytes for VLAN, a maximum of 60 bytes for IPv4/IPv6, 8 bytes for ICMP, and potential additional headers.
-// To be on the safe side, we assume an overhead of 150 bytes.
-const headers = 150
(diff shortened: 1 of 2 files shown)


src/connectivity_check/mtu_checker.go
```diff
@@ -38,6 +37,13 @@ func (m *mtuChecker) Check(attributes Attributes) ResultReporter {
 
 	var localIP string
 
+	mtu := attributes.OutgoingNIC.MTU
+
+	// Only consider hosts with MTU greater than 1500
+	if mtu <= defaultMTUThreshold {
```
zaneb, line 43: I'd suggest only skipping if mtu == 1500. That way all non-default values get validated.

zaneb:
LGTM. I assume you have tested this locally with a non-default MTU? I expect none of the CI jobs are exercising it.

linoyaslan:
> LGTM. I assume you have tested this locally with a non-default MTU? I expect none of the CI jobs are exercising it.
Yes, I tested it locally using stat-net to configure hosts with MTU < 1500 and MTU > 1500. I set the switch MTU smaller than the hosts' MTU, and the installation timed out—similar to what was described in the original ticket.

PR #911, opened 2025-02-14
MGMT-19420: Replace HighAvailabilityMode with ControlPlaneCount

andrej1991:
[MGMT-19420](https://issues.redhat.com//browse/MGMT-19420): Replace HighAvailabilityMode with ControlPlaneCount. It is the first phase. Now in the install parameters it still accepts the high-availability-mode just it translates the parameter to control-plane-count in the following maner:
- ControlPlaneCount is being set to 3 if --high-availability-mode is set to full
- ControlPlaneCount is being set ot 1 if --high-availability-mode is set to none
This PR is dependent on https://github.com/openshift/assisted-installer/pull/1027


src/commands/actions/install_cmd.go
```diff
@@ -105,8 +105,17 @@ func (a *install) getFullInstallerCommand() string {
 		"--agent-image", a.agentConfig.AgentVersion,
 	}
 
+	if a.installParams.ControlPlaneCount != nil {
+		installerCmdArgs = append(installerCmdArgs, "--control-plane-count", strconv.FormatInt(int64(*a.installParams.ControlPlaneCount), 10))
```
danmanor, line 109: Is there a scenario where `a.installParams.ControlPlaneCount` is `nil` ?

src/commands/actions/install_cmd.go
```diff
@@ -105,8 +105,17 @@ func (a *install) getFullInstallerCommand() string {
...
+	}
+
 	if a.installParams.HighAvailabilityMode != nil {
-		installerCmdArgs = append(installerCmdArgs, "--high-availability-mode", swag.StringValue(a.installParams.HighAvailabilityMode))
+		switch *a.installParams.HighAvailabilityMode {
```
danmanor, line 113: Is there a scenario where `a.installParams.HighAvailabilityMode` is nil ? maybe in the future ?

danmanor (approved):

src/commands/actions/install_cmd.go
```diff
@@ -272,6 +284,19 @@ func (a *install) validateDisks() error {
 	return nil
 }
 
+func (a *install) validateControlPlaneCount() error {
```
danmanor, line 287: As this value is provided by assisted-service, we don't have to validate it

danmanor: The rest of the arguments are not validated as well

danmanor (approved):

adriengentil (approved):

PR #920, opened 2025-02-20
MGMT-19588: Also use bond interface for hostname workaround

pastequo:
Tested with libvirt by creating a VM with 2 network interfaces. Then inside the VM, the bond was created manually with nmcli following [this](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/configuring_and_managing_networking/configuring-network-bonding_configuring-and-managing-networking#configuring-a-network-bond-by-using-nmcli_configuring-network-bonding) 

I only tried the `active-backup` mode, not sure if it could have an impact

### DIFF
--- a/src/inventory/inventory.go
+++ b/src/inventory/inventory.go
@@ -3,6 +3,7 @@ package inventory
 import (
 	"encoding/json"
 	"net"
+	"slices"
 	"sort"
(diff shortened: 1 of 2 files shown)


pastequo:
cc @jhernand

PR #925, opened 2025-03-12
MGMT-19966: fix route protocol returns string instead of int

eifrach:
- fix when compare the route protocol vs `unix.RTPROT_RA` it return a string value of the protocol.

### DIFF
--- a/src/util/network.go
+++ b/src/util/network.go
@@ -107,7 +107,7 @@ func isUsableIPv6Route(route netlink.Route) bool {
 		return false
 	}
 	// Ignore non-advertised routes
-	if route.Protocol != unix.RTPROT_RA {
+	if int(route.Protocol) != unix.RTPROT_RA {
 		return false
 	}
 


src/util/network.go
```diff
@@ -107,7 +107,7 @@ func isUsableIPv6Route(route netlink.Route) bool {
 		return false
 	}
 	// Ignore non-advertised routes
-	if route.Protocol != unix.RTPROT_RA {
+	if int(route.Protocol) != unix.RTPROT_RA {
```
linoyaslan, line 110: it's seems it's already int:

`//Protocol describe what was the originator of the route
type RouteProtocol int`

eifrach: it return's a String when I run a debug
`Mar 12 11:05:19 localhost.localdomain inventory[2007]: time="12-03-2025 11:05:19" level=info msg="Eran isUsableIPv6Route protcol kernel should be 9" file="network.go:102"`

due to the String function
```go
func (p [RouteProtocol]) String() string
```

linoyaslan: I believe the reason for that isn't because it's a string, but because unix.RTPROT_RA is of type int, and Route.Protocol is an alias of int, not int itself. AFAIK, implementing the Stringer interface doesn't mean the type will be treated as a string just because it has a String method.

linoyaslan: AFAIK, implementing the Stringer interface doesn't mean the type will be treated as a string just because it has a String method.
I know it's tiny thing, but I want to understand that, before lgtm.
https://go.dev/play/p/RWXdbsEzFWU
In the above I tried same +- structure of code. The current implementation should be fine, unless the route.Protocol change it's type in the flow, need to figure out why and what happen, don't you think?

eifrach: you are right let me rerun it with a Debug

PR #927, opened 2025-03-12
MGMT-19420: --high-availability-mode was removed from CLI arguments.

andrej1991:
[MGMT-19420](https://issues.redhat.com//browse/MGMT-19420): second phase, which means that --high-availability-mode was removed from CLI arguments.


andrej1991:
this code part was merged by PR 943

PR #990, opened 2025-04-24
MGMT-20296: Modify RegisterHostConflict case log in RegisterHostWithRetry

linoyaslan:
Following the changes introduced in [assisted-service PR #7604](https://github.com/openshift/assisted-service/pull/7604), when an installed host attempts to register, it triggers the RegisterHostWithRetry method in the agent. The returned error code has been updated to 409 Conflict, which now requires updating the corresponding log message in the agent to be more informative.

The expected log after both fixes:

`Host will stop trying to register; cluster cannot accept new hosts in its current state: [409] Host is trying to register after the cluster has already been installed. That most probably means that the host is booting from the installation ISO, and therefore not effectively joining the cluster. The request will be ignored. Fix the boot order and reboot the host.`

This also resolves the issue where the agent would enter an infinite registration loop when a 403 was returned. Previously, the error would fall into the default case in the switch statement and keep retrying. Now, with the 409 response, it matches the conflict case, causing the agent to wait forever instead of continuing to retry.


rccrdpccl:
/lgtm
/hold

Holding to give a chance to others to review before it auto-merges. Feel free to unhold when done

PR #991, opened 2025-04-24
OCPBUGS-51230: Bump github.com/go-jose/go-jose/ to v3.0.4  in release 4.17

jkaurredhat:
To align with assisted-service go-jose package, go-jose is upgraded to v3.0.4 which contains CVE fix. this go-jose version is supported with go1.20 used in this repository.


pastequo:
@jkaurredhat build is broken. You probably need to chnage the hive version as well

jkaurredhat:
@pastequo Thank you so much for reviewing this PR. 

What is the ideal practice in assisted-installer-agent:release-4.17  reference assisted-service : master or release-4.17 ? 

You are correct that it is referencing master right now and I am moving back to 4.17 but I checked history and before switching to master to release-4.17.  https://github.com/openshift/assisted-installer-agent/commit/29149b7793e1f15dcb0e39a54e79fddcac744ff8

image

which was referencing release-4.17 : https://github.com/openshift/assisted-service/pull/6750

Please suggest what would be the best possible solution 

1. to try upgrading hive in asissted-service in release-4.17 or 
2. to try updating assisted-service in this PR to master ? but that will reference 4.0.5.

pastequo:
What concerns me is that it's going to use a 2022 version of hive.
We need to assert the usage of it in this repo - even indirect - to be sure we are not going to introduce regression.

- If the default are the same etc (and it's just so golang typing) then you can update the go.mod to use the 2022 version of hive/apis
- If you suspect there is a bug fix or something, I guess the right move is to update in assisted-service first

PR #999, opened 2025-05-07
MGMT-20023: Route inventory deleted when route without next hop interface

rivkyrizel:
Before this fix, if a node had a special route like `route add -host 64.1.2.3 reject`, the agent failed to get the route’s link name and stopped collecting any routes. 
This resulted in an empty route list in the inventory and caused the host to fail validation. 
Now, if one route causes an error, we log it and skip it. so the rest of the routes are still collected correctly.


eliorerz (changes requested):
Look good, I think that we might want to add a UT that check the case that was before (except error) and the case now (except pass)

eliorerz (commented):
LGTM

eliorerz (approved):
LGTM

PR #1007, opened 2025-05-27
MGMT-20669: Fix hostname update for inventory get host

bkopilov:
When hostname changed from node, the inventory get host does not update the service. On hostname action , linux updates /proc/sys/kernel/hostname file on node but it does not update the kernel hostname on next_step_runner container.

Due to this issue manual hostname update or changes after starting next_step_runner never reflected unless we restart agent.

In order to fix it added to next_step_runner --uts=host flag. With this flag changes on /proc/sys/kernel/hostname reflected to the container path.

### DIFF
--- a/src/commands/actions/next_step_runner.go
+++ b/src/commands/actions/next_step_runner.go
@@ -53,7 +53,7 @@ func (a *nextStepRunnerAction) cleanupPrevious() {
 func (a *nextStepRunnerAction) Args() []string {
 	a.cleanupPrevious()
 
-	arguments := []string{"run", "--rm", "-ti", "--privileged", "--pid=host", "--net=host",
+	arguments := []string{"run", "--rm", "-ti", "--privileged", "--pid=host", "--uts=host", "--net=host",
 
 		// unlimited number of processes in the container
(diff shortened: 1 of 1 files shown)


danmanor:
@bkopilov  You probably need to rebase

tsorya:
We are talking about inventory? 
We run inventory as new container so it should get updated one
https://github.com/openshift/assisted-installer-agent/blob/master/src/commands/actions/inventory_cmd.go#L41

danmanor:
@bkopilov Did you test before and after this change and saw the issue goes away after the fix ?

bkopilov:
> @bkopilov what is blocking this PR from being merged?

Hi , Igal commented , i can try to reproduce it .  The problem is that nodes hostname not shared with container hostanme... so inventory running in a container not affected from changed hostname.

Adding the flag trigger changed in the running container the hostname (shared with the nodes hostname )

bkopilov:
i will double check it to make sure i am not missing

bkopilov:
I think we can move on with this fix, it worked locally.

PR #1021, opened 2025-06-12
MGMT-18993: ensure the agent reports the permMacAddress of network interfaces

omer-vishlitzky:
https://issues.redhat.com/browse/MGMT-18993
When installing a node configured with a bond interface comprising of two or more network interfaces, the mac address of the bond and all its slave interfaces is set to match the mac address of the first slave. This can lead to a mismatch with the bootMACAddress field of the corresponding BareMetalHost, preventing the agent from being correctly matched with it.
This PR ensures that each interface retains its actual permanent mac address, rather than the visible address set by the bonding configuration.

### DIFF
--- a/src/inventory/interfaces.go
+++ b/src/inventory/interfaces.go
@@ -8,6 +8,7 @@ import (
 	"github.com/openshift/assisted-installer-agent/src/util"
 	"github.com/openshift/assisted-service/models"
 	"github.com/sirupsen/logrus"
+	"github.com/vishvananda/netlink"
 )
 
 const ipv4LocalLinkCIDR = "169.254.0.0/16"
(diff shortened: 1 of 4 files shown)


src/connectivity_check/util_test.go
```diff
@@ -8,6 +8,13 @@ import (
 	"github.com/openshift/assisted-installer-agent/src/util"
 )
 
+func addHardwareAddrMock(iface util.Interface, macAddr string) {
```
carbonin, line 11: Why is this needed? Isn't this already being done by `NewFilledMockInterface`?

omer-vishlitzky: I have removed it from `FillInterfaceMock` since the interface tests do not need this mock anymore

carbonin: Okay, but does it break anything to leave it in there?
It seems odd to require the same mac string in two subsequent calls.

carbonin:
/hold
/approve
/lgtm

Looks good, but holding until @omer-vishlitzky can test this in a real environment.

carbonin (approved):

PR #1028, opened 2025-06-16
MGMT-20141: collecting and uploading ironic-agent logs

omer-vishlitzky:
https://issues.redhat.com/browse/MGMT-20141


src/config/logs_sender_config.go
```diff
@@ -61,8 +61,9 @@ func ProcessLogsSenderConfigArgs(defaultTextLogging, defaultJournalLogging bool)
 	}
 
 	loggingConfig.Tags = []string{"agent", "installer"}
+	loggingConfig.Services = []string{"ironic-agent"}
```
carbonin, line 64: Should this be conditional somehow? What happens if we're not running the `ironic-agent`? In SaaS for example.

PR #1316, opened 2026-02-06
MGMT-23300: Ensure /boot dir has enough space during reclaim

CrystalChun:
Reclaim can fail if there's not enough space to write the boot artifacts. Recently, OCP 4.19 and later has taken up more space in the `/boot` directory, which prevents the Agent from writing all of the artifacts needed for reclaim.

To resolve this, the Agent will check if there's enough space in the `/boot` directory and if there isn't, it will try to remove the ostree by running `rpm-ostree cleanup --os=rhcos -r`.

## Summary by CodeRabbit

* **Improvements**
  * Boot artifact downloads now use a temporary staging area and are moved atomically to avoid touching existing host files.
  * Bootloader config creation is more robust and written to a safe temporary location with clearer success messages.
  * Added disk space checks with optional reclamation, permission-preserving file moves/copies, and more verbose progress logging to improve reliability.

### DIFF
--- a/src/commands/actions/download_boot_artifacts_cmd.go
+++ b/src/commands/actions/download_boot_artifacts_cmd.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/openshift/assisted-installer-agent/src/config"
+	"github.com/openshift/assisted-installer-agent/src/util"
 	"github.com/openshift/assisted-service/models"
 	log "github.com/sirupsen/logrus"
 )
@@ -102,6 +103,11 @@ func run(infraEnvId, downloaderRequestStr, caCertPath string) error {
 		return fmt.Errorf("failed creating bootloader config file on host: %s", err.Error())
 	}
 	log.Infof("Successfully wrote bootloader config to %s", path.Join(tempBootArtifactsFolder, bootLoaderConfigFileName))
+
+	if err := ensureBootHasSpace(getMountedBootFolder(*req.HostFsMountDir)); err != nil {
+		log.Errorf("failed to ensure boot folder has enough space: %s", err.Error())
+		return fmt.Errorf("failed to ensure boot folder has enough space: %s", err.Error())
+	}
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


src/commands/actions/download_boot_artifacts_cmd.go
```diff
@@ -155,10 +174,26 @@ func download(httpClient *http.Client, filePath, url string, retry int) error {
 	return nil
 }
 
-func createBootLoaderConfig(rootfsUrl, artifactsPath, bootLoaderPath string) error {
-	kernelPath := path.Join(artifactsPath, kernelFile)
-	initrdPath := path.Join(artifactsPath, initrdFile)
-	bootLoaderConfigFile := path.Join(bootLoaderPath, bootLoaderConfigFileName)
+func downloadArtifacts(req models.DownloadBootArtifactsRequest, caCertPath string, folders *folders) error {
```
carbonin: Can this just take `tempDownloadFolder` as an argument? That would make it more clear that they're being downloaded into a temp dir, then moved later.

src/commands/actions/download_boot_artifacts_cmd.go
```diff
@@ -155,10 +174,26 @@ func download(httpClient *http.Client, filePath, url string, retry int) error {
...
+		return fmt.Errorf("failed downloading kernel to host: %s", err.Error())
+	}
+
+	if err := download(httpClient, path.Join(folders.tempDownloadFolder, initrdFile), *req.InitrdURL, defaultRetryAmount); err != nil {
+		return fmt.Errorf("failed downloading initrd to host: %s", err.Error())
+	}
+	return nil
+}
+
+func createBootLoaderConfig(rootfsUrl string) error {
```
carbonin: Can we just rename this to something that indicates it's being created in the temp dir rather than the real location?

src/commands/actions/download_boot_artifacts_cmd.go
```diff
@@ -178,14 +251,134 @@ func createFolderIfNotExist(folder string) error {
...
+
+	// Preserve file permissions
+	sourceInfo, err := os.Stat(src)
+	if err != nil {
+		return fmt.Errorf("failed to stat source file %s: %w", src, err)
+	}
+	if err := os.Chmod(dst, sourceInfo.Mode()); err != nil {
+		return fmt.Errorf("failed to set permissions on %s: %w", dst, err)
 	}
 	return nil
```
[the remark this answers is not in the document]
carbonin: Isn't this being done on line 401 now?

src/commands/actions/download_boot_artifacts_cmd.go
```diff
@@ -65,40 +79,45 @@ initrd %s`
 )
 
 func run(infraEnvId, downloaderRequestStr, caCertPath string) error {
+	if _, err := os.Stat(path.Join(tempBootArtifactsFolder, "boot_artifacts_moved")); err == nil {
```
carbonin: Instead of doing a flag could we check if we already have boot files in the paths we expect? I think that would accomplish the same thing.

carbonin (approved):
Thanks for doing all this. Looks great now!

CrystalChun:
/retitle [MGMT-23300](https://redhat.atlassian.net/browse/MGMT-23300): Ensure /boot dir has enough space during reclaim

PR #1374, opened 2026-03-24
OCPBUGS-63472: add --copy-network when manual network config is detected

rwsu:
When the agent receives the install step, check for NetworkManager keyfiles created by the user during the agent-tui session before constructing the coreos-installer command. If found, append --copy-network to the installer args so that network configuration created via the agent-tui persists into the installed OS.

The next-step-runner runs in a container with --pid=host but without /etc/NetworkManager/system-connections mounted. The host's NM connections directory is accessed via /proc/1/root/etc/NetworkManager/system-connections, which resolves to the host's filesystem when the container shares the host's PID namespace.

The agent-tui log file (/var/log/agent/agent-tui.log) is used to confirm the TUI ran on this node (i.e. we are in an ABI workflow). If absent, the check is skipped entirely to avoid affecting non-ABI workflows. The /var/log directory is mounted into the next-step-runner container so this path is accessible directly.

The mtime of the agent log directory (/var/log/agent) is used as the TUI start time reference. Auto-generated keyfiles (nm-initrd-generator, pre-network-manager-config.sh for NMState configs) are created before the agent-tui starts and are excluded by the mtime comparison. User-created keyfiles via nmtui can only appear at or after the TUI starts. The comparison uses >= rather than > to handle filesystems with second-level mtime precision.

When NMState static configs are provided via agent-config.yaml, assisted-service already adds --copy-network via StaticNetworkConfig so those keyfiles do not need to be detected here. --copy-network is not added if already present in the installer args set via the API.

Assisted-by: Claude Sonnet 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/commands/actions/action.go
+++ b/src/commands/actions/action.go
@@ -60,7 +60,7 @@ func New(agentConfig *config.AgentConfig, stepType models.StepType, args []strin
 		models.StepTypeContainerImageAvailability: {&imageAvailability{args: args, agentConfig: agentConfig}},
 		models.StepTypeStopInstallation:           {&stopInstallation{args: args}},
 		models.StepTypeLogsGather:                 {&logsGather{args: args, agentConfig: agentConfig}},
-		models.StepTypeInstall:                    {&install{args: args, filesystem: afero.NewOsFs(), agentConfig: agentConfig}},
+		models.StepTypeInstall:                    {&install{args: args, filesystem: afero.NewOsFs(), agentConfig: agentConfig, birthTimeFn: defaultBirthTimeFn}},
 		models.StepTypeUpgradeAgent:               {&upgradeAgent{args: args}},
 		models.StepTypeDownloadBootArtifacts:      {&downloadBootArtifacts{args: args, agentConfig: agentConfig}},
 		models.StepTypeRebootForReclaim:           {&rebootForReclaim{args: args}},
--- a/src/commands/actions/install_cmd.go
+++ b/src/commands/actions/install_cmd.go
@@ -4,7 +4,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"path/filepath"
 	"regexp"
+	"time"
 	"strconv"
 	"strings"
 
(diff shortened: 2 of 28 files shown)


src/commands/actions/install_cmd.go
```diff
@@ -4,7 +4,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"path/filepath"
 	"regexp"
+	"time"
```
zaneb: The linter is [not happy](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_assisted-installer-agent/1374/pull-ci-openshift-assisted-installer-agent-master-edge-lint/2036538292257689600) with the sorting here.

zaneb (commented):
/approve

src/commands/actions/install_cmd.go
```diff
@@ -272,6 +287,120 @@ func (a *install) validateDisks() error {
...
+		// keyfiles (nm-initrd-generator, NMState via pre-network-manager-config.sh)
+		// are written before the TUI launches and will have an earlier mtime. We use
+		// >= rather than > to handle filesystems with second-level mtime precision
+		// where the user's keyfile and the TUI start may share the same timestamp.
+		info, err := a.filesystem.Stat(filePath)
+		if err != nil {
+			continue
+		}
+		log.Infof("Found NM keyfile: %s (mtime %v)", filePath, info.ModTime())
+		if info.ModTime().Before(tuiStart) {
```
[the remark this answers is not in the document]
andfasano: Ok

src/commands/actions/action.go
```diff
@@ -60,7 +60,7 @@ func New(agentConfig *config.AgentConfig, stepType models.StepType, args []strin
 		models.StepTypeContainerImageAvailability: {&imageAvailability{args: args, agentConfig: agentConfig}},
 		models.StepTypeStopInstallation:           {&stopInstallation{args: args}},
 		models.StepTypeLogsGather:                 {&logsGather{args: args, agentConfig: agentConfig}},
-		models.StepTypeInstall:                    {&install{args: args, filesystem: afero.NewOsFs(), agentConfig: agentConfig}},
+		models.StepTypeInstall:                    {&install{args: args, filesystem: afero.NewOsFs(), agentConfig: agentConfig, birthTimeFn: defaultBirthTimeFn}},
```
andfasano, line 63: Q: Any specific reason for this kind of configurability (for the birthTime func)? I'm not sure why it's required (and if not, another occasion to reduce the lines of code)

rwsu: The configurability is needed for unit testing. defaultBirthTimeFn calls times.Stat() from djherbis/times, which invokes the real statx syscall under the hood. That doesn't work on afero.MemMapFs, so there's no way to control the returned birth time in tests without making it injectable.

gamli75 (approved):

src/commands/actions/action.go
andfasano: Ack thanks

andfasano:
It looks like the previous command was ignored by the CI

/cherry-pick release-4.22

PR #1401, opened 2026-04-14
NO-ISSUE: Refactor flag Parsing to avoid multiple calls

pastequo:
This PR is about preparing a second PR to add more parameter for the `inventory` command.

- Inventory args are isolated to avoid impacting other subprocesses
- flag.Parse is no longer called several times. AFAIU it is not supported to do it
```golang
// Parse parses the command-line flags from [os.Args][1:]. Must be called
// after all flags are defined and before flags are accessed by the program.
```

I believe it would be best to refactor using a framework like [cobra](https://github.com/spf13/cobra), but that would have probably way more impact

## Summary by CodeRabbit

* **Refactor**
  * Unified CLI argument and logging registration so dry‑run and logging flags behave consistently across commands.
  * Separated inventory-specific configuration from subprocess tooling for clearer configuration boundaries.

* **New Features**
  * Inventory command now has a dedicated config with a GPU config file option and dry‑run overrides (forced MAC/host IP) for simulated runs.

### DIFF
--- a/src/config/agent_config.go
+++ b/src/config/agent_config.go
@@ -22,13 +22,18 @@ func printHelpAndExit() {
 
 func ProcessArgs() *AgentConfig {
 	ret := &AgentConfig{}
+
+	err := RegisterDryRunArgs(&ret.DryRunConfig)
+	if err != nil {
+		log.Fatalf("Failed to register dry run arguments: %v", err)
+	}
+
+	RegisterLoggingArgs(&ret.LoggingConfig)
+
 	flag.StringVar(&ret.TargetURL, "url", "", "The target URL, including a scheme and optionally a port (overrides the host and port arguments")
 	flag.StringVar(&ret.InfraEnvID, "infra-env-id", "", "The value of infra-env-id")
 	flag.StringVar(&ret.AgentVersion, "agent-version", "", "Full image reference of the agent, for example 'quay.io/edge-infrastructure/assisted-installer-agent:v2.5.2'")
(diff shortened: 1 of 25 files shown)


src/config/agent_config.go
rccrdpccl: Any reason why we're not calling `RegisterLoggingArgs` here?

rccrdpccl:
LGTM overall, just a small question and please acknowledge or dismiss the rabbit's comments :pray:

src/config/agent_config.go
pastequo: I hesitated at first but yeah, I will do it too

src/inventory/bmc.go
```diff
@@ -14,16 +14,16 @@ import (
...
-	subprocessConfig *config.SubprocessConfig
+	dependencies    util.IDependencies
+	inventoryConfig *config.InventoryConfig
 }
 
-func newBMC(subprocessConfig *config.SubprocessConfig, dependencies util.IDependencies) *bmc {
-	return &bmc{dependicies: dependencies, subprocessConfig: subprocessConfig}
+func newBMC(inventoryConfig *config.InventoryConfig, dependencies util.IDependencies) *bmc {
+	return &bmc{dependencies: dependencies, inventoryConfig: inventoryConfig}
 }
```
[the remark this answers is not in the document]
pastequo: I didn't change that part, just renamed few variables (and change to InventoryConfig). I think your check is valid but belongs to another PR

PR #1518, opened 2026-06-24
MGMT-20398: Don't start main agent until ironic reports it's ready

pastequo:
The check is based on a json file.
If the file is missing, the feature is considered as not implemented or that we aren't in a ABI installation ; then the check is skipped.

Assisted-by: Opencode

Following [this slack discussion](https://redhat-internal.slack.com/archives/CFP6ST0A3/p1772810746105179?thread_ts=1767898775.874839&cid=CFP6ST0A3)

Will require another PR in the ironic agent repo to actually write the file

cc @CrystalChun @dtantsur 

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring the ironic status file via a new command-line flag or environment variable, with a default path if neither is set.
  * The agent now validates the status file before starting and waits until the service is ready when needed.
* **Bug Fixes**
  * Improved handling of missing, empty, initializing, finalizing, and unexpected status values, including clearer error outcomes.
  * The agent terminates with a fatal error if status-file validation fails.
* **Tests**
  * Added coverage for status-file validation and readiness waiting behavior.

### DIFF
--- a/src/util/status_file.go
+++ b/src/util/status_file.go
@@ -1,0 +1,123 @@
+package util
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+// StatusFileContent represents the structure of the ironic agent status file
+type StatusFileContent struct {
+	Status string `json:"status"`
+}
+
+const (
+	StatusInitializing    = "initializing"
+	StatusWaitingForAgent = "waiting for assisted agent"
+	StatusFinalizing      = "finalizing"
+)
+
+var (
+	waitingLoop = 2 * time.Second
+)
+
+// ValidateStatusFile checks the ironic agent status file and waits for readiness if needed.
+// It implements the following state machine:
+// - File not found: Feature not implemented (skip wait)
+// - Empty file: Feature available but initializing (wait indefinitely)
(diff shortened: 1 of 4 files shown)


src/util/status_file.go
```diff
@@ -0,0 +1,123 @@
...
+
+	err = waitForIronicReadiness(filePath)
+	if err != nil {
+		return fmt.Errorf("failed to check ironic readiness: %w", err)
+	}
+
+	return nil
+}
+
+func waitForIronicReadiness(filePath string) error {
```
shay23bra, line 72: Maybe we should add a timeout? In case ironic never reaches readiness the agent might hang forever.

shay23bra (changes requested):

src/util/status_file.go
pastequo: That is a real question. If Ironic is never ready, should we still proceed and start the installation ? I think in this case we are very likely to create a stale status (handled by the other PR)

More context: practically it's ironic that is starting the agent, this code was added to prevent users to manually start the agent before ironic is ready. In the happy path, this additional check is 100% useless. 
So Im afraid that if we put a timeout, we are defeating the initial intent

shay23bra: Makes sense!

